PR #7, opened 2016-10-01
Reduce founder reward to 10% from block 2000

abrkn:
From website

> There will be 21 million Zcoins. Zcoin follows the same halving cycle as Bitcoin (every 4 years). In the first 4 years, 10% of Zcoins will be distributed to the Founders Reward. In other words, during the first 4 years, 40 Zcoins will go to the miners and 10 Zcoins will go towards the Founders reward. After the first 4 years, the block reward goes completely towards the miners.

10 coins of 50 coins is 20%, not 10%

I suggest we hard fork soon, for example at block 1000, to reduce payouts

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,11 @@
 ZCoin integration/staging tree
 ================================
 
-http://zcoin.tech/
+https://zcoin.tech/
 
-Copyright (c) 2009 - 2016 Bitcoin Developers
-Copyright (c) 2015 - 2016 ZCoin Developers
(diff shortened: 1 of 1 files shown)


abrkn:
There are some typos there I'm sure, but you get the point

nopslide:
I will raise this issue to investors and my team, tq

zcoindude:
Hi there, 20% for the first 4 years totals to 10% of the total Zcoin supply. We will update the FAQ shortly to clarify this point.

abrkn:
:+1:

PR #57, opened 2017-02-07
Final Lyra2Z

aizensou:


src/main.cpp
```diff
@@ -757,7 +757,7 @@ bool CTransaction::CheckTransaction(CValidationState &state, uint256 hashTx, boo
 
                     // CHECKING PROCESS
                     BOOST_FOREACH(const CZerocoinEntry& pubCoinItem, listPubCoin) {
-                        if (pubCoinItem.value == pubCoin)
+                        if (pubCoinItem.value == pubCoin && pubCoinItem.denomination == denomination)
```
daira: Why is this being modified in a commit that is supposed to be about the proof of work?

PR #72, opened 2017-02-18
Merge back to Dev branch

nopslide:


daira:
You're setting MIN_PEER_PROTO_VERSION to be the same as the new PROTOCOL_VERSION. Isn't that going to cause a precipitous drop in mining power as everyone has to upgrade at once, so the network connectivity has to start from scratch?

bitcartel:
What does the one line change in wallet.cpp fix/tweak?

https://github.com/zcoinofficial/zcoin/pull/72/files#diff-d7618bdc04db23aa74d6a5a4198c58fdL1487

Previously the denomination was always set to 1 i.e. libzerocoin::ZQ_LOVELACE but now it can span the entire range of denominations upto libzerocoin::ZQ_WILLIAMSON.

daira:
There are *171 lines* of security-critical code duplicated for each of 5 coin denominations. This is not normal, and really demands an explanation of why it was written this way.

nopslide:
@daira Regarding MIN_PEER_PROTO_VERSION, this is the correct number, please have a look at master before I merged. It already is that version. So it won't have the problem like you said.

For the 171 lines of code, if you want to make it generic, I really welcome it and have no problem with that.  I wrote it in that way because I didn't want to overthink to make it generic as long as it can do the job, it's fine for me.

bitcartel:
What happens if a new user joins the network or an existing user decides to -reindex?  Won't the fraudulent transactions now be treated as invalid, thereby leading to a fork?

PR #73, opened 2017-02-18
Merge Dev

nopslide:


bitcartel:
There's a risk of an integer overflow with pubcoinId as nSequence is an unsigned int.

```
 int pubcoinId = txin.nSequence;
...
 if (pubcoinId < 1 && pubcoinId == INT_MAX) { // IT BEGINS WITH 1
```

```
class CTxIn
...
    unsigned int nSequence;
```

PR #74, opened 2017-02-22
PVS-Studio: fixed vulnerability CWE-401 (Memory Leak)

SvyatoslavRazmyslov:
We have found and fixed a vulnerability (CWE-401) using PVS-Studio tool. Analyzer warning: [V773](https://www.viva64.com/en/w/V773/).
[PVS-Studio](https://www.viva64.com/en/pvs-studio/) is a static code analyzer for C, C++ and C#.


sn-ntu:
The fix is good. However, the correct file path is "src/crypto/Lyra2Z/Lyra2.c", not "zcoin/src/Lyra2Z/Lyra2.c". That's why the patch could not be merged into the master.

Could you please make another PR which correct file path? If I do it manually, you will not receive credit which I am afraid of.

Thanks.

PR #82, opened 2017-03-25
[FIX] miniupnpc compilation issue

TimurKiyivinski:
This fixes the compilation issue with various miniupnpc version across
various systems. The patch is adapted from what was used in the Monero
project here: https://github.com/monero-project/monero/pull/422/files


TimurKiyivinski:
A fix has been merged at last 👍

PR #91, opened 2017-05-27
Check API version of miniupnc

real-or-random:
I've seen just that #82 will fix this as well... Anyway, this PR or #82 does the job.


real-or-random:
And a second compilation fix ...

PR #101, opened 2017-06-16
change the algorithm to index the transactions

leoreinaux:
The maxid approach could cause problems if a old wallet has a wrong index
- performance improvement - just read listpubcoin one time


nopslide:
tested without problem ?

leoreinaux:
yes, did some tests and at #testing channel too

PR #105, opened 2017-07-06
Build for ARM64

rainbowstain:
Building Zcoin daemon for ARM64 on Ubuntu 16.04 with GCC 5.4 yield a compiler error. [`Please implement AtomicPointer for this platform`](https://github.com/zcoinofficial/zcoin/blob/master/src/leveldb/port/atomic_pointer.h#L212). This indicates that neither GCC 5.4 for ARM64 has cstdatomic support, nor `atomic_pointer.h` has an appropriate arch section for ARM64.

To remedy the error, two places are patched.

1. **`Lyra2Z/sph_types.h`**
  - ARM64's instruction set only handles little-endian (could handle big-endien data though).  
2. **`leveldb/port/atomic_pointer.h`**
  - GCC 5 has `__AARCH64EL__` definition for ARM64 which is missing in the header and causes the compiler to complain. `ARCH_CPU_A64_FAMILY` is added to support ARM64.
  - ARM64 memory barrier function isn't universally available so `asm volatile` is deployed.

Please review and modify as it fits.

### DIFF
--- a/src/Lyra2Z/sph_types.h
+++ b/src/Lyra2Z/sph_types.h
@@ -1053,7 +1053,7 @@ typedef long long sph_s64;
 /*
  * ARM, little-endian.
  */
-#elif defined __arm__ && __ARMEL__
+#elif defined __arm__ && __ARMEL__ || __AARCH64EL__
 
 #define SPH_DETECT_LITTLE_ENDIAN     1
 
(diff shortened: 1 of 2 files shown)


rainbowstain:
As `v0.13.2.4` compiles & runs fine, I'll close this. Thanks for the upgrade!

PR #112, opened 2017-09-11
test dummy myfix

glynn-farrow:
Simple dummy fix to see if I can create a 'fix' and relevant pull-request on core_upgrade

Since I have FUG (fear of using git) I test first
Feel free to reject the 'fix'


snguyen-vn:
Closed

PR #113, opened 2017-09-12
fix a couple of 'unused variable' compile warnings

glynn-farrow:
definition: made nStartRewardTime const
policy/fees: commented out line with unused var curPri
commented and did not delete as looks like others are working in there


devwarrior777:
Will put in new fix for this build warning against 'master' from a different clone - this can be closed

PR #117, opened 2017-09-14
Separated zerocoin code from main.cpp

psolstice:


nopslide:
Could you changed base on current code ?

PR #142, opened 2017-11-22
Fixed problem when client gets stuck on some blocks

sn-ntu:


sn-ntu:
Fixed problem when client gets stuck on some blocks

PR #144, opened 2017-11-23
fix typo in zcoin-cli znode help command

rloomba:
Just a trivial typo fix.


o0oBluePhoenixo0o:
Also in line 463 as well. 

>  "  payee          - Print Dash address associated with a znode (can be additionally filtered,\n"

Thanks @rloomba , I was going to pull here but you did it before me 👍

PR #154, opened 2017-12-19
Update README.md

ChairmanMeow:
TYPO - Don't you mean Zcoin instead of Bitcoin?


riordant:
Thanks - please do find more of these!

PR #177, opened 2018-02-03
Suggested changes to apply wallet fees from 'fundrawtransaction'

devwarrior777:
[Wallet] Re-introduced bitcoin 0.13 code to apply fees in wallet.cpp
Please use as a basis to fix:
Issue https://github.com/zcoinofficial/zcoin/issues/172 Fees requested in `fundrawtransaction` are not applied


devwarrior777:
Coin control feature is from QT wallet. QT wallet is not running. Zcoind is running.
The atomicswap code calls `fundrawtransaction` as RPC and then `FundTransaction` in wallet.cpp which always fills in CoinControl structure. and pass to `CreateTransaction` in wallet.cpp.

PR #201, opened 2018-04-29
Change confusing wording of error message

catchingknives:
Existing error message "[...] it has to have at least two mint coins with at least 6 confirmation in order to spend a coin" is confusing to many users. 

"What does this mean?", "Do I have to create two mints in order to do one spend?", "What does 'it' refer to?" etc. were common questions. 

Hoping to provide a clearer error message here.


devwarrior777:
I agree this is a better message.

PR #205, opened 2018-05-09
Separation of Znode and Wallet libraries

riordant:
Currently the code for Znode's depends on Wallet functionality being enabled; As a result, passing the `disable-wallet` flag to `configure` will cause the build to fail outright (see https://github.com/zcoinofficial/zcoin/issues/204).
Myself and @devwarrior777 propose a two step process forward:

(1) initially disable the `disable-wallet` option; this will ensure building regardless of flags passed to configure
(2) work towards separation of Znode functionality into a separate library

Part (1) is contained in this pull request.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,11 +92,7 @@ AC_PATH_TOOL(OBJCOPY, objcopy)
 AC_ARG_VAR(PYTHONPATH, Augments the default search path for python module files)
 
 # Enable wallet
-AC_ARG_ENABLE([wallet],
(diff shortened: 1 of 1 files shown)


devwarrior777:
`(2) work towards separation of Znode functionality into a separate library`

.. if feasible ;-)

reubenyap:
@nopslide @psolstice Please comment on your views. Would also be a good idea to take a look at latest Dash code as to how this is implemented.

reubenyap:
@psolstice Should we look into this now or later when we refactor?

PR #209, opened 2018-05-24
Allow znode to run on any port.

brndnmtthws:
Currently there's a requirement that znodes run on the same port (8168).
However, there are 2 major issues with this:

 - the network is susceptible to an attack where an ISP or other
   provider could simply block port 8168, breaking all znodes
 - it's impossible to run multiple znodes with the same IP, which is
   largely an artificial limitation

### DIFF
--- a/src/znodeconfig.cpp
+++ b/src/znodeconfig.cpp
@@ -68,23 +68,6 @@ bool CZnodeConfig::read(std::string& strErr) {
         LogPrintf("mainnetDefaultPort=%s\n", mainnetDefaultPort);
         LogPrintf("Params().NetworkIDString()=%s\n", Params().NetworkIDString());
         LogPrintf("CBaseChainParams::MAIN=%s\n", CBaseChainParams::MAIN);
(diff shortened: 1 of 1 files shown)


soerenmartius:
+1

Kick1986:
Good idea!
It will be great if the team update us about this, because you are right about the ISP can block the port!

ultimaweapon:
Thanks for contribution! We are sorry to take so long time to merge this.

ultimaweapon:
@brndnmtthws we decided to revert this changes. After talking with @yura-pakhuchiy, the purpose of Znode is to make Zcoin network reliable. We cannot compare Znode with normal node since it have a different purpose. That mean it does not make sense to have more than one Znode on the same physical computer since if that computer is going down it will also take down all Znode that currently running on it.

reubenyap:
It is noted that Dash's code also prevents this and only runs on one port. I think this is the reason why they also disable ipv6 by default.

PR #212, opened 2018-07-02
RPC additions and updates

riordant:
## Added spendmanyzerocoin RPC command

this RPC command allows the user to specify multiple addresses and denominations to spend. If the user wishes to make 1 or more spends to themselves, the address field is left blank. 
The number of spends to make to the address can also be specified.
For example, to spend 25 and 50 to two different third parties respectively:
```
spendmanyzerocoin "[{\"address\":\"TVyUgzLh1MCnjtwv3b4oCFB8STbpKPVSxu\",\"denomination\":25, \"amount\": 1},{\"address\":\"TY842FSFJH4TbFMXAdrgmb9XgDjfvJKqvc\",\"denomination\":50, \"amount\": 1}]"
```

or to spend 2 50's to a 3rd party, and 10 to yourself:
```
spendmanyzerocoin "[{\"address\":\"TWDwwUawEB5noiBuf8auamWPxouVeZn9hh\",\"denomination\":50, \"amount\": 2},{\"address\":\"\",\"denomination\":10, \"amount\": 1}]"
```
The function returns a list of the transaction IDs generated for the spends.

## Add mintmanyzerocoin RPC command

Along the same lines as `spendmanyzerocoin`, add a function that does several mints in a single call. the only parameter is a JSON object of the form `{<denomination>(1,10,25,50,100):amount,...}`, where `amount` is the number of mints to do for the `denomination` specified.
For example, to do 2 mints of denomination 10, and 5 mints of denomination 25:
```
mintmanyzerocoin "{\"10\":2,\"25\":5}"
```
The function returns the list of txids for the mints generated.

## Remove temporary sync block for zerocoin spend (for testnet usage)

The addition of the following into `wallet/wallet.cpp` makes the RPC spending functionality not work on testnet and regtest mode:
```
    // Do not allow mint to take place until fully synced
    // Temporary measure: we can remove this limitation when well after spend v1.5 HF block
    if (fImporting || fReindex || !znodeSync.IsBlockchainSynced())
        return _("Not fully synced yet");
``` 
As this was a temporary measure I propose we now remove this.

## Return txid instead of pubcoin after mint

currently the `mintzerocoin` RPC command returns the pubcoin that was generated from the mint. In line with other transaction-generating calls, I propose we instead return the transaction ID that the mint was contained within. The data related to the mint transaction can be returned following the `mintzerocoin` call in the `listmintzerocoins` function.

## Add znode, mint, and spend payment categories in ListTransactions
Add the ability to return whether or not this transaction is an incoming znode payment, a mint transaction, or a spend transaction in the `ListTransactions` RPC call in `wallet/rpcwallet.cpp`. This also means extending the `MaybePushAddress` function to take a pointer to a `CBitcoinAddress` instance.

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -115,6 +115,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
         //[zcoin]
     { "setmininput", 0 },
     { "mintzerocoin", 0 },
+    { "mintmanyzerocoin", 0 },
     { "spendzerocoin", 0 },
     { "spendmanyzerocoin", 0 },
     { "setgenerate", 0 },
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -116,6 +116,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "setmininput", 0 },
     { "mintzerocoin", 0 },
     { "spendzerocoin", 0 },
+    { "spendmanyzerocoin", 0 },
     { "setgenerate", 0 },
     { "setgenerate", 1 },
     { "setmintzerocoinstatus", 2 },
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -21,6 +21,8 @@
 #include "walletdb.h"
 #include "zerocoin.h"
 
+#include <znode-payments.h>
+
 #include <stdint.h>
 
 #include <boost/assign/list_of.hpp>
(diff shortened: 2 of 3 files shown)


joernroeder:
i would flip keys and values for `mintmanyzerocoin` to `<denomination>: amount`. imo key should be the denomination and value the amount how much of that particular denomination you're minting:

> For example, to do 2 mints of denomination 10, and 5 mints of denomination 25:

```json
mintmanyzerocoin "{\"10\": 2, \"25\": 5}"
```

riordant:
Sure. makes no difference to me. Ideally I would like the JSON to be
directly convertible from the API call so whatever works

catchingknives:
Excuse my noobishness but why is it necessary to have double quotes in double quotes creating the need to escape those? Doesn't {}:, delimit the data just fine?

riordant:
@catchingknives I think it needs to parse the whole argument as a string, so you need to properly format special characters when making a call from command line

riordant:
**Updates**: 
- Update `spendmanyzerocoin` to include an amount to spend
- swap positions of denomination and amount as per @joernroeder's comment

PR #218, opened 2018-07-09
Zerocoin Mint updates

riordant:
The `IsZerocoinMint` function in `transaction.cpp` takes as a parameter a reference to a transaction. As the object that implements this function is itself a transaction, passing the transaction reference is unnecessary, and can be removed.

What this function is doing is checking whether or not a vout entry for this transaction contains a `ZEROCOIN_MINT`. I propose passing a reference to a `CTxOut` object, and setting this object with the vout entry containing the `ZEROCOIN_MINT`, should the transaction contain a mint. The passing of this reference is optional to the caller; ie. if not passed, it will simply check whether this transaction is a mint transaction. This allows us to more easily do things from the calling code, as is discussed next.

Finally, a function (`IsMintFromTxOutUsed`) was added, which allows checking whether a `CTxOut` object contains a mint transaction that has been used.


reubenyap:
@psolstice any feedback on this? Please look into this once you have finished refactor-mtp.

riordant:
This PR has been superseded by updated code in `client-api/`. As a result, I propose we close it and then review the relevant changes when merging `client-api/`.

PR #238, opened 2018-08-13
Minor mtp refactoring

a-bezrukov:


src/miner.cpp
```diff
@@ -1142,25 +1142,17 @@ void static ZcoinMiner(const CChainParams &chainparams) {
             LogPrintf("pblock->nNonce: %s\n", &pblock->nNonce);
             LogPrintf("powLimit: %s\n", Params().GetConsensus().powLimit.ToString());
 
-            bool fMTPIsRequired = pblock->nTime >= Params().nMTPSwitchTime;
-            CMTPHashData *mtpHashData = nullptr;
-            if (fMTPIsRequired) {
-                pblock->mtpHashData = make_shared<CMTPHashData>();
-                mtpHashData = pblock->mtpHashData.get();
-            }
+            bool const fMTPIsRequired = pblock->nTime >= Params().nMTPSwitchTime;
```
psolstice: It's better to use pblock->IsMTP() here

configure.ac
```diff
@@ -110,10 +110,17 @@ AC_ARG_ENABLE([upnp-default],
...
+  *linux*)
+     zusetests=yes
+  ;;
+esac
+
 AC_ARG_ENABLE(tests,
     AS_HELP_STRING([--disable-tests],[do not compile tests (default is to compile)]),
     [use_tests=$enableval],
-    [use_tests=no])
+    [use_tests=$zusetests])
```
psolstice: Please don't enable building tests by default

PR #247, opened 2018-09-04
Exodus Protocol

nopslide:
Enable "Exodus Protocol" on top of Zcoin to allow everyone to tokenize digital asset

PR #249, opened 2018-09-06
Fixed build error when system have zstd installed

### LINKED ISSUE #248: Try to enable ZSTD for Tor
Right now it seem like it is disabled due to it build successfully. On the system that have zstd installed it will fail in the linking state due to unresolved symbols for zstd (which can be workaround by passing `--disable-zstd` to `src/tor`). We may try to update Tor to the latest version to see if they already fixed this issue. If it still does not work, we might try to fix by ourself.

Related issue: https://github.com/zcoinofficial/zcoin/issues/193#issuecomment-416519201

ultimaweapon:


ultimaweapon:
Still have no idea how to fix error on Linux CI. I will switch to #247 until I figure it out.

ultimaweapon:
@reubenyap I think we don't need ZSTD since we already disabled LZMA. ZSTD is just another way to compress data alternate to LZMA. Right now we also did not use both of its and everything is still working. Disable it instead of try to make it work is more easier. My idea is we should disable it for now to fix build errors on some platform then we can try to enable it again later when we really need it. What do you think?

reubenyap:
Okay it is to be noted that Tor should be upgraded to latest release version rather than a beta version. There are also some fixes to zstd in latest release build. https://blog.torproject.org/new-release-tor-0348-also-other-stable-updates-02917-03212-and-03310

ultimaweapon:
I agreed with upgrade Tor :) For ZSTD fix in the latest Tor I think it does not fix our issue. Our issue come from we try to link Tor as a static library. The problem with this way is we need to manually link all of dependencies that Tor required into our binary.

ultimaweapon:
Look like previous build fail come from refactoring configuration scripts instead of ZSTD fix. I have no idea why but reverse it to the latest version then just added option to disable ZSTD and it work. But just realized that we still should not try to enable ZSTD unless we tested it. So this PR is ready to merge.

PR #250, opened 2018-09-10
Repairing bitcore-node-zcoin auto tests

a-bezrukov:


src/chainparams.cpp
```diff
@@ -408,16 +408,8 @@ class CRegTestParams : public CChainParams {
         extraNonce[1] = 0x00;
         extraNonce[2] = 0x00;
         extraNonce[3] = 0x00;
-        genesis = CreateGenesisBlock(ZC_GENESIS_BLOCK_TIME, 414098458, 0x207fffff, 1, 0 * COIN, extraNonce);
+        genesis = CreateGenesisBlock(ZC_GENESIS_BLOCK_TIME, 3, 0x207fffff, 1, 0 * COIN, extraNonce);
```
psolstice: Why 3? Nonce is chosen (via brute force) to get block PoW hash lower than the target value (encoded as 0x207fffff here). It's highly unlikely 3 will give you the result. In some cases (e.g. reading from disk) it's actually checked and will lead to assert failure

src/chainparams.cpp
```diff
@@ -451,9 +443,12 @@ class CRegTestParams : public CChainParams {
         nModulusV2StartBlock = 130;
         nModulusV1MempoolStopBlock = 135;
         nModulusV1StopBlock = 140;
-        nMTPSwitchTime = INT_MAX;
+        nMTPSwitchTime = 1;
         nDifficultyAdjustStartBlock = 0;
         nFixedDifficulty = 0;
+
+        if(genesis.IsMTP())
```
psolstice: genesis block is never MTP

a-bezrukov: Yes, realized that but left the code here by mistake. Removing.

PR #251, opened 2018-09-10
Add Dockerfile for zcoind

ultimaweapon:
After have a discussion with Joern, there is still some issues that need to verify:

1. Check which IP and port that `zcoind` broadcasting to the network. We need a way to make it broadcast host IP instead of container IP. We also need some way to make it able to broadcast a custom port to support Docker port mapping.

Run the following command in a root repository to build docker image:

```sh
docker build -t zcoin-daemon .
```


ultimaweapon:
After looking at available command line options, we can use option `-externalip=<ip>` to specify host IP. So we can merge this PR now. The only remaining issues is port number broadcasting which we need one more option to specify external port like external IP (currently not available). Even we are not able to map external port we still get the benefit from this PR since it open more platforms for zcoind to run. Also make it able for who are already familiar with Docker to easily getting started with Zcoin.

soerenmartius:
@ultimaweapon shouldn't docker build by default use max cpu capicity unless we limit it?

https://docs.docker.com/engine/reference/commandline/build/#options

ultimaweapon:
That is a good information :) Actually the problem come from the step that execute `make` for building Zcoin which currently is not parallelized. What I mean is Docker allow us to do jobs in parallel but we did not take that advantage. With your information now we can make it default to parallel to the number of available CPU cores.

PR #267, opened 2018-09-29
Allow for multiple ZeroSpend inputs in a single transaction

riordant:
- Adds transaction creation logic
- Extends transaction validation logic first created by @psolstice 
- Adds RPC method
- Adds relevant tests

A few more tests are to be added, however I think it's best to get the review process underway now.


src/zerocoin.cpp
```diff
@@ -53,26 +53,28 @@ static bool CheckZerocoinSpendSerial(CValidationState &state, CZerocoinTxInfo *z
 }
 
 bool CheckSpendZcoinTransaction(const CTransaction &tx,
-                                libzerocoin::CoinDenomination targetDenomination,
+                                vector<libzerocoin::CoinDenomination> targetDenominations,
```
ultimaweapon: I think it will be better if we change to `const vector<libzerocoin::CoinDenomination>& targetDenominations` instead to prevent copying the whole vector.

src/wallet/wallet.cpp
```diff
@@ -3297,6 +3297,47 @@ bool CWallet::CreateZerocoinSpendModel(string &stringError, string thirdPartyAdd
 
 }
 
+bool CWallet::CreateZerocoinSpendModel(CWalletTx& wtx, string &stringError, string thirdPartyAddress, vector<string> denomAmounts, bool forceUsed) {
```
ultimaweapon: I think it should be `const string& thirdPartyAddress` and `const vector<string>& denomAmounts` instead to prevent create another object.

src/wallet/wallet.cpp
```diff
@@ -3297,6 +3297,47 @@ bool CWallet::CreateZerocoinSpendModel(string &stringError, string thirdPartyAdd
...
+    if (!fFileBacked)
+        return false;
+     
+    vector<pair<int64_t, libzerocoin::CoinDenomination>> denominations;
+    for(vector<string>::iterator it = denomAmounts.begin(); it != denomAmounts.end(); it++){
+        string denomAmount = (*it).c_str();
```
ultimaweapon: `const string& denomAmount = *it;` should be better to prevent instantiate another object.

src/wallet/wallet.cpp
```diff
@@ -3880,6 +3921,296 @@ bool CWallet::CreateZerocoinSpendTransaction(std::string &thirdPartyaddress, int
...
+ * @param wtxNew
+ * @param reservekey
+ * @param coinSerial
+ * @param txHash
+ * @param zcSelectedValue
+ * @param zcSelectedIsUsed
+ * @param strFailReason
+ * @return
+ */
+bool CWallet::CreateMultipleZerocoinSpendTransaction(std::string &thirdPartyaddress, std::vector<std::pair<int64_t, libzerocoin::CoinDenomination>> denominations,
```
ultimaweapon: This method is quite big. This make it hard to understand the code. Can we split it out to smaller methods?

src/wallet/wallet.cpp
```diff
@@ -4045,6 +4376,77 @@ string CWallet::SpendZerocoin(std::string &thirdPartyaddress, int64_t nValue, li
...
+ * @param nValue
+ * @param denomination
+ * @param wtxNew
+ * @param coinSerial
+ * @param txHash
+ * @param zcSelectedValue
+ * @param zcSelectedIsUsed
+ * @return
+ */
+string CWallet::SpendMultipleZerocoin(std::string &thirdPartyaddress, std::vector<std::pair<int64_t, libzerocoin::CoinDenomination>> denominations, CWalletTx &wtxNew,
```
ultimaweapon: Is it possible to change to `const std::vector<std::pair<int64_t, libzerocoin::CoinDenomination>>& denominations` instead? This will prevent a new vector instantiation, thus increase execution performance.

src/wallet/wallet.cpp
```diff
@@ -4045,6 +4376,77 @@ string CWallet::SpendZerocoin(std::string &thirdPartyaddress, int64_t nValue, li
...
+    // if (nValue <= 0)
+    //     return _("Invalid amount");
+     // Do not allow spend to take place until fully synced
+    // Temporary measure: we can remove this limitation when well after spend v1.5 HF block
+    // if (fImporting || fReindex || !znodeSync.IsBlockchainSynced())
+    //     return _("Not fully synced yet");
+     CReserveKey reservekey(this);
+     if (IsLocked()) {
+        string strError = _("Error: Wallet locked, unable to create transaction!");
+        LogPrintf("SpendZerocoin() : %s", strError);
```
ultimaweapon: I think it need to be `strError.c_str()` instead. I don't known what type that is returned from `_()` but if it is `char *` then we can switch `strError` to `const char *` instead to increase performance.

src/wallet/wallet.cpp
riordant: Yes, a separate instance is also not necessary here.

src/wallet/rpcwallet.cpp
```diff
@@ -3000,38 +3006,83 @@ UniValue spendmanyzerocoin(const UniValue& params, bool fHelp) {
...
-                                                         zcSelectedIsUsed);
+    if (pwalletMain->IsLocked()) {
+        string strError = _("Error: Wallet locked, unable to create transaction!");
+        LogPrintf("SpendZerocoin() : %s", strError);
+        return strError;
+    }
 
-            if (strError != "")
-                throw JSONRPCError(RPC_WALLET_ERROR, strError);
+    string strError = "";
```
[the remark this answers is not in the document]
riordant: The whole layout of `strError` here is confusing. I'm going to add it as an empty object initially and assign it as needed.

riordant:
@ultimaweapon : All changes submitted bar the function split.

ultimaweapon:
@riordant Travis still fail :)

riordant:
@ultimaweapon Passing now.. what happened was our .gitignore (for some reason) includes the Makefile in `depends` aswell as the patches there.. I'm not sure why. To me, .gitignore should only specify files that should not be included in the remote repo at all. What do you think?

ultimaweapon:
You just found another bugs in our repository :) What I guess is `.gitignore` in the root of repository recursively excluded `Makefile` in all children directories because usually `Makefile` will be generated by `configure` script so we don't want to include it. After looking at it found out it is true. We can put `!Makefile` in `depends/.gitignore` to negate the exclusion in this directory and all of its child.

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
ultimaweapon: Sorry I don't quite understand what you mean.

riordant:
@ ultimaweapon:
- Those files are autogenerated yes, however to stop them being pushed to remote, they should be included in `.gitignore`.
- I did, they will be re-added. Again, this is an issue with our `.gitignore` and the way it handles files.
- It is an IDE file for Qt-Creator. I will add it to .gitignore and remove it.

src/zerocoin_params.h
```diff
@@ -44,6 +44,8 @@ static const int64_t DUST_HARD_LIMIT = 1000;   // 0.00001 XZC mininput
 
 #define ZC_MODULUS_V2_BASE_ID			1000
 
+#define ZC_MULTIPLE_SPEND_INPUT_STARTING_BLOCK  105000
```
riordant: This will have to be changed.

src/zerocoin.cpp
```diff
@@ -1064,6 +1098,17 @@ set<CBlockIndex *> CZerocoinState::RecalculateAccumulators(CChain *chain) {
     return changes;
 }
 
+bool CZerocoinState::AddSpendToMempool(const vector<CBigNum> &coinSerials, uint256 txHash) {
```
ultimaweapon: `const uint256& txHash` should be better :)

src/zerocoin.cpp
```diff
@@ -1064,6 +1098,17 @@ set<CBlockIndex *> CZerocoinState::RecalculateAccumulators(CChain *chain) {
...
+    BOOST_FOREACH(CBigNum coinSerial, coinSerials){
+        if (IsUsedCoinSerial(coinSerial) || mempoolCoinSerials.count(coinSerial))
+            return false;
+
+        mempoolCoinSerials[coinSerial] = txHash;
```
ultimaweapon: Remove `mempoolCoinSerials.count(coinSerial)` from the above line and replace this line with the following code should be better:

```cpp
auto r = mempoolCoinSerials.insert(std::make_pair(coinSerial, txHash));
if (!r.second) {
    return false;
}
```

The benefits with this way is:

1. Hash for `coinSerial` need to calculate only one time instead of two.
2. Lookup need to do only one time instead of two.
3. A new value of `uint256` will construct with copy constructor instead of construct with default constructor then invoke assignment operator. Most of the time copy constructor is efficient than assignment operator.

ultimaweapon:
The code itself is okay for me the remaining comments is just a small suggestion. But still need someone who are expert in this field to review business logic.

psolstice:
Merged with refactor-mtp-L64. Should probably close this PR without merge as it'll get merged into master with MTP branch

reubenyap:
Closing this PR as merged into refactor-mtp-l64

Issue #278, opened 2018-10-08
Add instructions how to run zcoind with docker


ultimaweapon:
We need to update both GitHub and https://hub.docker.com/r/zcoinofficial/zcoind.

ultimaweapon:
We should also update our website too.

reubenyap:
Resolved in #279

PR #279, opened 2018-10-09
Add instructions how to run with Docker

ultimaweapon:
Part of #278.


reubenyap:
@FreakyMaryk Please verify that the instructions work and that you can setup docker too by following it.

riordant:
Updated grammar.

PR #282, opened 2018-10-09
Exodus UI cosmetic improvements

a-bezrukov:


src/qt/exodus_qtutils.cpp
```diff
@@ -109,4 +110,34 @@ std::string ReplaceStr(const std::string& findText, const std::string& replaceTe
...
+        isUiNeeded_ = GetBoolArg("-showExodusUi", false);
+    }
+
+private:
+    bool isUiNeeded_;
+};
+
+bool uiNeeded()
+{
+    return ShowUiChecker::getInstance().isUiNeeded();
```
ultimaweapon: I think `ShowUiChecker` is not necessary. We can just use:

```cpp
static bool v = GetBoolArg("-showExodusUi", false);
return v;
```

psolstice: I mean don't use GetBoolArg() from static variables at all. It might work in some cases and not in others (because it relies on global variable initialization order which is not well defined)

a-bezrukov: Peter, this is not global static var but a local static var. It's gonna be initialized on the method first call, which happens from the UI initialization. At UI initialization  the args map has been already initialized

a-bezrukov: Here is the code snippet which shows the initialization order (bitcoin.cpp:516):
`int main(int argc, char *argv[])
{
    SetupEnvironment();

    /// 1. Parse command-line options. These take precedence over anything else.
    // Command-line options take precedence:
    ParseParameters(argc, argv);

    // Do not refer to data directory yet, this can be overridden by Intro::pickDataDirectory

    /// 2. Basic Qt initialization (not dependent on parameters or configuration)
#if QT_VERSION < 0x050000
    // Internal string conversion is all UTF-8
    QTextCodec::setCodecForTr(QTextCodec::codecForName("UTF-8"));
    QTextCodec::setCodecForCStrings(QTextCodec::codecForTr());
#endif

    Q_INIT_RESOURCE(bitcoin);
    Q_INIT_RESOURCE(bitcoin_locale);

    BitcoinApplication app(argc, argv);
#if QT_VERSION > 0x050100
    // Generate high-dpi pixmaps
    QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);`

ultimaweapon:
After talking with @a-bezrukov, he will refactor later. So I will merge it.

PR #291, opened 2018-10-14
Add warning before dumpwallet is executed

riordant:
PR for issue described here: https://github.com/zcoinofficial/zcoin/issues/290
Currently the text is @reubenyap 's warning.

### DIFF
--- a/src/wallet/authhelper.cpp
+++ b/src/wallet/authhelper.cpp
@@ -1,0 +1,101 @@
+#include <random>
+#include <string>
+#include <map>
+
+#include "base58.h"
+#include "authhelper.h"
+
+struct AuthorizationHelper::Impl
+{
+    using rnd_type = size_t;
+    using code_type = std::string;
+    Impl()
+    : seed()
+    , generator(seed())
+    , rnd_dist(std::numeric_limits<rnd_type>::min(), std::numeric_limits<rnd_type>::max())
+    {}
+
+    bool authorize(code_type const & functionName, code_type code)
+    {
+        auto const fn_key_iter = key_store.find(functionName);
+        if(fn_key_iter == key_store.end())
+        {
+            generateAuthorizationCode(functionName);
+            return false;
+        }
+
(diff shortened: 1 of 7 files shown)


reubenyap:
@a-bezrukov Please double check if changing the dumpwallet command by adding a string at the end as per Yura's request or to add a dialog box would break any compatibility with Electrum / Insight before we merge this.

src/wallet/authhelper.cpp
```diff
@@ -0,0 +1,101 @@
...
+    code_type generateAuthorizationCode(code_type const & functionName)
+    {
+        rnd_type key = rnd_dist(generator);
+        key_store[functionName] = key;
+        return toCode(key);
+    }
+
+private:
+    std::random_device seed;
+    std::mt19937 generator;
```
psolstice: There is (was?) a bug in mingw implementation of random_device
https://sourceforge.net/p/mingw-w64/bugs/338/
I need to double check if on Windows it returns always the same string

a-bezrukov: Wow! Checking also.

psolstice: Yes, it's always the same
There are two options
1. Use boost::random_device. This requires adding new dependency (libboost-random-dev)
2. Switch to hashing wallet data (keys)

a-bezrukov: Changed it to be according to the 1st option.
LOCK2(cs_main, pwalletMain->cs_wallet);
generator.seed(pwalletMain->GetHDChain().masterKeyID.GetUint64(0) ^ time(NULL));

psolstice: LGTM but please manually test if it gives different results on different Windows installations (even if the `time(NULL)` is exactly the same) before merging to master

psolstice:
don't forget to manually test on Windows

a-bezrukov:
Tested on Win7 and Win8, all cases provide random OTAC.

PR #292, opened 2018-10-15
Correct Spelling of Tor

catchingknives:
From the [Tor FAQ](https://www.torproject.org/docs/faq.html.en):
>Note: even though it originally came from an acronym, Tor is not spelled "TOR". Only the first letter is capitalized. In fact, we can usually spot people who haven't read any of our website (and have instead learned everything they know about Tor from news articles) by the fact that they spell it wrong.

That's not us, right? :wink:


riordant:
This should be fine, I'm just not sure if this overlaps with Bitcoin code. @psolstice ? Non-cosmetic changes where bitcoin core code is affected are to be left as-is for now

catchingknives:
@riordant I know as much about Bitcoin code as the next goat but didn't we put that there? Should be fine

PR #298, opened 2018-10-20
MTP release

psolstice:
MTP starting date is December 10th 12:00 UTC
Initial MTP difficulty is needed before this PR can be merged
Version is set to 0.13.7.1


src/test/transaction_tests.cpp
```diff
@@ -776,78 +779,4 @@ BOOST_AUTO_TEST_CASE(test_IsStandard)
     BOOST_CHECK(!IsStandardTx(t, reason));
 }
 
-BOOST_AUTO_TEST_CASE(Test_EnforceZnodePayment)
```
a-bezrukov: Was this test removed by mistake?

a-bezrukov: Was this test removed by mistake?

a-bezrukov: Was this test removed by mistake?

a-bezrukov: Was this test removed by mistake?

mfrankovi: This test was moved to another file znode_tests. The reason for that is that znode checks are not done on checkTransaction but in checkBlock so another test suite was needed to build actual blockchain and do tests on blocks and not transactions.

PR #300, opened 2018-10-24
Fix for the unit tests -- rejected

a-bezrukov:
The switch to MTP will halve rewards per block because the distance between blocks will reduce to 5 minutes. We change bitcoin's subsidy halving block number (originally 210000) to keep total reward amount over given period of time exactly the same. This is a fix for bitcoin test, nothing more.

### DIFF
--- a/src/amount.h
+++ b/src/amount.h
@@ -28,7 +28,7 @@ extern const std::string CURRENCY_UNIT;
  * critical; in unusual circumstances like a(nother) overflow bug that allowed
  * for the creation of coins out of thin air modification could lead to a fork.
  * */
-static const CAmount MAX_MONEY = 21000000 * COIN;
+static const CAmount MAX_MONEY = 30500000 * COIN;
 inline bool MoneyRange(const CAmount& nValue) { return (nValue >= 0 && nValue <= MAX_MONEY); }
 
 /**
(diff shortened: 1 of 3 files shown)


src/amount.h
```diff
@@ -28,7 +28,7 @@ extern const std::string CURRENCY_UNIT;
  * critical; in unusual circumstances like a(nother) overflow bug that allowed
  * for the creation of coins out of thin air modification could lead to a fork.
  * */
-static const CAmount MAX_MONEY = 21000000 * COIN;
+static const CAmount MAX_MONEY = 30500000 * COIN;
```
yura-pakhuchiy: Zcoin going to have 30.5M coins?

soerenmartius: what? this is new to me!

yura-pakhuchiy: @reubenyap @a-bezrukov @psolstice

psolstice: switch to MTP will halve rewards per block because the distance between blocks will reduce to 5 minutes. We changed the bitcoin's subsidy halving block number (originally 210000) to keep total reward amount over given period of time exactly the same. This is a fix for bitcoin test, nothing more. Incorrectly named.

riordant:
@a-bezrukov If this is a valid change (and not accidental), please provide a description of the reasons for the change

catchingknives:
To the interested public: We are **not** changing our supply

a-bezrukov:
This PR was merging to a branch, not to master. I created it just to use the convenience that github provides for PRs. Sorry for the alarm.

PR #314, opened 2018-11-26
Fix for multiple indexing issues which affect Insight

a-bezrukov:


src/addresstype.h
```diff
@@ -0,0 +1,29 @@
+#ifndef ADDRESSTYPE_H
+#define ADDRESSTYPE_H
+
+enum struct AddressType
+{
+      unknown = 0
+    , payToPubKeyHash = 1
+    , payToScryptHash = 2
```
psolstice: Should it be `script`?

src/main.cpp
```diff
@@ -3033,24 +2904,27 @@ bool ConnectBlock(const CBlock &block, CValidationState &state, CBlockIndex *pin
...
             return AbortNode(state, "Failed to write transaction index");
 
 
     if (fTimestampIndex)
         if (!pblocktree->WriteTimestampIndex(CTimestampIndexKey(pindex->nTime, pindex->GetBlockHash())))
             return AbortNode(state, "Failed to write timestamp index");
 
+    if (!pblocktree->AddTotalSupply(block.vtx[0].GetValueOut() - nFees))
+        return AbortNode(state, "Failed to write total supply");
+
```
psolstice: Should we call it only when some flags are true?

src/main.cpp
```diff
@@ -2517,57 +2490,36 @@ bool DisconnectBlock(const CBlock &block, CValidationState &state, const CBlockI
...
+                return error("Failed to write address unspent index");
+            }
         }
-        if (!pblocktree->UpdateAddressUnspentIndex(addressUnspentIndex)) {
-            AbortNode(state, "Failed to write address unspent index");
-            return error("Failed to write address unspent index");
+
+        if (!pblocktree->AddTotalSupply(-(block.vtx[0].GetValueOut() - nFees))) {
+            AbortNode(state, "Failed to write total supply");
+            return error("Failed to write total supply");
```
psolstice: Move in under `fAddressIndex` if?

src/rpc/misc.cpp
```diff
@@ -958,6 +1003,157 @@ UniValue getspentinfo(const UniValue& params, bool fHelp)
...
+                        "\nExamples:\n"
+                + HelpExampleCli("gettotalsupply", "")
+                + HelpExampleRpc("gettotalsupply", "")
+        );
+
+    CAmount total = 0;
+
+    if(!pblocktree->ReadTotalSupply(total))
+        throw JSONRPCError(RPC_DATABASE_ERROR, "Cannot read the total supply from the database");
+
```
psolstice: It will return bogus value if we upgraded from previous version and didn't do reindex. If `fAddressIndex` (or something similar) is required to get the total supply at least we can warn user here

src/addresstype.h
a-bezrukov: Renamed

src/main.cpp
a-bezrukov: Done

src/rpc/misc.cpp
a-bezrukov: Done

psolstice:
LGTM

PR #315, opened 2018-11-27
Fix rounduint64() for Windows

ultimaweapon:
This fix the following issue when running on Windows:

```
exodus/test/rounduint64_tests.cpp(77): error: in "exodus_rounduint64_tests/rounduint64_limits": check 18446744073709551615U == rounduint64(18446744073709551615.0L) has failed [18446744073709551615 != 0]
```

This caused by Windows does not support 80-bit extended precision floating point via `long double` (it just an alias to `double` ([source](https://msdn.microsoft.com/en-us/library/9cx8xs15.aspx)).

### DIFF
--- a/src/exodus/convert.cpp
+++ b/src/exodus/convert.cpp
@@ -18,9 +18,9 @@ static bool isBigEndian()
   return 1 == bint.c[0];
 }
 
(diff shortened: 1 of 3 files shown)


psolstice:
LGTM, runs fine on WSL now

PR #330, opened 2018-12-14
Expose missing ports for Docker

ultimaweapon:


riordant:
All of the ports are valid, this can be merged.

PR #336, opened 2018-12-20
Prohibit pre-0.13.7.* version from connecting

psolstice:
Test if it helps with znodes problem


reubenyap:
What is the current status on this?

psolstice:
It should have got merged a month ago, unless someone objects in next 24 hours I'll merge it

PR #337, opened 2018-12-25
update checkpoint to block height after MTP

nopslide:


nopslide:
Forget to update time to 1545712287, please help me on that one pls

PR #342, opened 2019-01-09
Changes to mintmanyzerocoin 

riordant:
- Change arguments
    To make it easier for the user, Instead of JSON input, inputs are expected in the following format:
    `mintmanyzerocoin <denomination> <amount> <denomination> <amount> ...`
    for all denominations necessary.
- update help message
- display help message on entering command
- fail should available balance be less than mint amount requested
- fail should mint inputs be greater than ZC_MINT_LIMIT (currently 100)
- Update error msgs

- update the `mintzerocoin` RPC function to return the help message should no arguments be entered, as is the standard with other RPC functions

- Correctly signal the GUI on mint status changes.

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -115,7 +115,6 @@ static const CRPCConvertParam vRPCConvertParams[] =
         //[zcoin]
     { "setmininput", 0 },
     { "mintzerocoin", 0 },
-    { "mintmanyzerocoin", 0 },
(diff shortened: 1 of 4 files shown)


FreakyMaryk:
@riordant , Verified, but please let me know when you are finish with manymint new RPC input string

riordant:
@FreakyMaryk updated and tested.

PR #344, opened 2019-01-14
Changed the fee estimation moving average params

a-bezrukov:


reubenyap:
Can we keep the comments in but modified with the correct values? Removing them might make it harder to remember later. Also to remind us what was the Bitcoin default.

a-bezrukov:
Done

FreakyMaryk:
Looks good. Tested.

PR #350, opened 2019-02-07
Reg test block generate fix

FreakyMaryk:
Now we fixed all test setup issues
- block generating
- node sync

Particular fix for wallet.py (optimized to zcoin)


qa/rpc-tests/test_framework/util.py
```diff
@@ -312,12 +312,12 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
     if binary is None:
         binary = os.getenv("ZCOIND", "zcoind")
         print(binary)
-    args = [ binary, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-reindex" ,"-mocktime="+str(get_mocktime()) ]
+    args = [ binary, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-reindex", "-dandelion=0", "-mocktime="+str(get_mocktime()) ]
```
thebevrishot: need to support enable dandelion because have test file `p2p-dandelion.py`

qa/rpc-tests/receivedby.py
```diff
@@ -41,9 +41,10 @@ def run_test(self):
         '''
         listreceivedbyaddress Test
         '''
+
         # Send from node 0 to 1
         addr = self.nodes[1].getnewaddress()
-        txid = self.nodes[0].sendtoaddress(addr, 0.1)
+        txid = self.nodes[0].sendtoaddress(addr, 10)
```
ultimaweapon: Why do we change from `0.1` to `10`?

src/main.cpp
```diff
@@ -1947,7 +1947,7 @@ bool ReadBlockHeaderFromDisk(CBlock &block, const CDiskBlockPos &pos) {
 
 CAmount GetBlockSubsidy(int nHeight, const Consensus::Params &consensusParams, int nTime) {
     // Just want to make sure no one gets a dime before 28 Sep 2016 12:00 AM UTC
-    if (nTime < nStartRewardTime && !consensusParams.IsTestnet())
+    if (nTime < nStartRewardTime && consensusParams.IsMain())
```
ultimaweapon: I don't think this is the right fix. Regtest should similar to mainnet as much as possible. We should try to fix in another way.

qa/rpc-tests/wallet.py
```diff
@@ -29,42 +30,43 @@ def setup_network(self, split=False):
...
         # Check that only first and second nodes have UTXOs
         assert_equal(len(self.nodes[0].listunspent()), 1)
         assert_equal(len(self.nodes[1].listunspent()), 1)
         assert_equal(len(self.nodes[2].listunspent()), 0)
 
-        # Send 21 BTC from 0 to 2 using sendtoaddress call.
-        self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 11)
+        # Send 21 Zcoin from 0 to 2 using sendtoaddress call.
         self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 10)
+        time.sleep(5)
```
ultimaweapon: Why do we need to sleep?

qa/rpc-tests/wallet.py
```diff
@@ -105,13 +107,14 @@ def run_test (self):
         self.nodes[1].sendrawtransaction(txns_to_send[0]["hex"], True)
         self.nodes[1].sendrawtransaction(txns_to_send[1]["hex"], True)
 
-        # Have node1 mine a block to confirm transactions:
-        self.nodes[1].generate(1)
+        # Have node1 mine a 7 block to confirm transactions:
+        self.nodes[1].generate(7)
+        time.sleep(2)
```
ultimaweapon: This sleep too.

qa/rpc-tests/wallet.py
```diff
@@ -123,251 +126,251 @@ def run_test (self):
...
+        # singletxid = self.nodes[0].sendrawtransaction(signedtx["hex"])
+        # txids = [singletxid, singletxid]
+        # self.nodes[0].generate(1)
+
+        # # Make a long chain of unconfirmed payments without hitting mempool limit
+        # txid_list = []
+        # for i in range(chainlimit*2):
+        #     txid_list.append(self.nodes[0].sendtoaddress(chain_addrs[0], Decimal('0.0001')))
+        # assert_equal(self.nodes[0].getmempoolinfo()['size'], chainlimit*2)
+        # assert_equal(len(txid_list), chainlimit*2)
```
ultimaweapon: Why do we comment out all above tests?

src/main.cpp
thebevrishot: This will be remove after we merge from bitcoin. we can leave it and comment testcase that mocktime as the time before constant start time.

qa/rpc-tests/wallet.py
FreakyMaryk: Cause starting from self.check_fee_amount test should be rewritten and adapted to zcoin.

qa/rpc-tests/wallet.py
FreakyMaryk: This test very unstable. Sometimes need to wait till transactions will be received by node. Need to investigate how to improve this test. I think now is a bad time to uncomment it in common regtests file.

qa/rpc-tests/wallet.py
FreakyMaryk: This test very unstable. Sometimes need to wait till transactions will be received by node. Need to investigate how to improve this test. I think now is a bad time to uncomment it in common regtests file.

qa/rpc-tests/test_framework/util.py
FreakyMaryk: I think, that now it is ok, cause the test still commented

qa/rpc-tests/wallet.py
thebevrishot: it should be solved after disable dandelion. isn't it?

qa/rpc-tests/test_framework/util.py
thebevrishot: It's not a big deal. then I'll create a card at backlog to support dandelion test.

src/main.cpp
ultimaweapon: If Bitcoin already removed this check then we should remove it now instead of comment out test cases. @psolstice do you see any problems if we remove this check?

PR #351, opened 2019-02-08
Fix Tor issue when cross compile for Linux

### LINKED ISSUE #188: undefined reference to `SSL_CIPHER_find'
Hello,

I'm struggling with building the newest zcoin 0.13.5.6. I'm getting an error:

 CXXLD    zcoind
tor/src/common/libor-crypto.a(tortls.o): In function `tor_tls_classify_client_ciphers':
tortls.c:(.text+0x4f7): undefined reference to `SSL_CIPHER_find'

Is there any fix for this?

ultimaweapon:
The reason it fail is because `$PKG_CONFIG --variable=libdir openssl` will try to grab system library instead of library in `depends` directory. I don't know why it also fixed undefined references to `evutil_secure_rng_set_urandom_device_file` too. Maybe the above `pkg-config` mess with configure prefix somehow.

Fixes #188

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1159,7 +1159,7 @@ if test x$need_bundled_univalue = xyes; then
 fi
 
 AX_SUBDIRS_CONFIGURE([src/secp256k1], [[--disable-shared], [--with-pic], [--with-bignum=no], [--enable-module-recovery], [--enable-experimental], [--enable-module-ecdh]])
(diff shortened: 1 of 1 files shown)


ultimaweapon:
@riordant you can try to see if it work on your machine or not. I already tested with my machine and it worked.

riordant:
Tested ACK 
Tested building `x86_64-linux-gnu` in the `depends` directory.
I think this can be merged - though shall we wait for CI to ensure it works for all cross-compilations?

ultimaweapon:
The upcoming CI will not build for other platforms when running on branches and PRs. It will build only when it do CD (which is not supported yet). So just a manual test on this should be enough.

PR #354, opened 2019-02-22
Add listtransactionsfrom RPC to support casino

ultimaweapon:
This is the original patch for Bitcoin: https://github.com/hackverket/bitcoin-core/commit/a13cd4d3f1cbfbb8f9af12b01e65ac0066c6e8ad


ultimaweapon:
The problems is this one is a special RPC for a specific purpose. I don't know what requirement they need to just apply their patch to our Zcoin. If we need to do it in another way we need to ask @reubenyap to see if it possible to do in that way.

riordant:
Ok. Please just take a look at my comments on the text in the function in that case.

src/wallet/rpcwallet.cpp
```diff
@@ -1594,6 +1594,115 @@ UniValue listtransactions(const UniValue& params, bool fHelp)
...
+// This for supporting casino. Ask Reuben if you need more information.
+UniValue listtransactionsfrom(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() > 4)
+        throw runtime_error(
+            "listtransactions ( \"account\" count from includeWatchonly)\n"
+            "\nReturns up to 'count' most recent transactions skipping the first 'from' transactions for account 'account'.\n"
```
riordant: `most recent` change to `first`

src/wallet/rpcwallet.cpp
```diff
@@ -1594,6 +1594,115 @@ UniValue listtransactions(const UniValue& params, bool fHelp)
...
+            "    \"otheraccount\": \"accountname\",  (string) For the 'move' category of transactions, the account the funds came \n"
+            "                                          from (for receiving funds, positive amounts), or went to (for sending funds,\n"
+            "                                          negative amounts).\n"
+            "    \"bip125-replaceable\": \"yes|no|unknown\"  (string) Whether this transaction could be replaced due to BIP125 (replace-by-fee);\n"
+            "                                                     may be unknown for unconfirmed transactions not in the mempool\n"
+            "  }\n"
+            "]\n"
+
+            "\nExamples:\n"
+            "\nList the most recent 10 transactions in the systems\n"
```
riordant: `most recent` change to `first`

reubenyap:
Ideally this is left as a flag but let's have the feedback first.

ultimaweapon:
@reubenyap they still not response?

reubenyap:
Still waiting. Have been following up closely.

ultimaweapon:
@reubenyap is there any updates on this?

reubenyap:
As there has been no reply from them for a while, I don't mind having this functionality but let's do it via a flag rather than introducing a new rpc for it.

nopslide:
this feature has been requested from Roger, not sure they are still using it atm. I think it is good to change to flag, but at the same time, please ask them first before remove the old rpc, otherwise, it will effect casino site that already integrated.

reubenyap:
After many months they haven't integrated it yet so they haven't been working on it. 


---- On Thu, 30 May 2019 20:07:43 +0800 notifications@github.com wrote ----



this feature has been requested from Roger, not sure they are still using it atm. I think it is good to change to flag, but at the same time, please ask them first before remove the old rpc, otherwise, it will effect casino site that already integrated.

ultimaweapon:
Thanks for update. Will come back to this after all important tasks is done.

reubenyap:
I agree with @a-bezrukov and @riordant on just keeping this as a flag to do it in the reverse order.

ultimaweapon:
After take a look there is some mismatch behavior. The example is if we provide `from` more than transactions we have `listtransactionfrom` will return all of transactions but for `listransaction` will return the empty list instead.

So if we really want to have this functionalities in our code base I think it is better to keep this as a separated function and hide it from `help` command.

reubenyap:
Since the casino guys didn't reply for several months we'll drop this until we need it.

PR #355, opened 2019-02-25
Re-org of TOR Qt checkbox as conf setting

riordant:
We currently have an option within Qt to enable TOR at the start page, via the `Anonymous communication with Tor` checkbox. When this checkbox is clicked, a `torsetting.dat` file is created within the datadir, which contains a single character `1`. This file is then checked for on startup and starts TOR with default settings should it be found.

The way that this is done is not consistent with the rest of the project: Settings are either considered as entries to the `conf` file, or are stored in QSettings for Qt, or both, with `conf` taking precedence in the latter case. 

This PR proposes making the TOR checkbox feature a conf setting, as `-torsetup`, with a corresponding Qt setting. As stated, the conf setting will override the Qt setting. The setting of the checkbox will also be based on this hierarchy. I've also removed relevant functions from `util.cpp` which are no longer needed.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -478,6 +478,9 @@ std::string HelpMessage(HelpMessageMode mode) {
     strUsage += HelpMessageOpt("-timeout=<n>",
                                strprintf(_("Specify connection timeout in milliseconds (minimum: 1, default: %d)"),
                                          DEFAULT_CONNECT_TIMEOUT));
+    strUsage += HelpMessageOpt("-torsetup",
+                               strprintf(_("Anonymous communication with TOR - Quickstart (default: %d)"),
+                                         DEFAULT_TOR_SETUP));
     strUsage += HelpMessageOpt("-torcontrol=<ip>:<port>",
                                strprintf(_("Tor control port to use if onion listening enabled (default: %s)"),
                                          DEFAULT_TOR_CONTROL));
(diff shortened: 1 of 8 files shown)


src/init.cpp
```diff
@@ -478,6 +478,9 @@ std::string HelpMessage(HelpMessageMode mode) {
     strUsage += HelpMessageOpt("-timeout=<n>",
                                strprintf(_("Specify connection timeout in milliseconds (minimum: 1, default: %d)"),
                                          DEFAULT_CONNECT_TIMEOUT));
+    strUsage += HelpMessageOpt("-torsetup=<n>",
```
ultimaweapon: Seem like we treat this as a boolean. So I think change `-torsetup=<n>` to `-torsetup` looking better.

riordant: Yes you're right.

riordant:
Updated and squashed as one commit

reubenyap:
I agree with this change. Can the coders review pls

reubenyap:
I agree with this change. Can the coders review pls

reubenyap:
I agree with this change. Can the coders review pls

ultimaweapon:
I'm okay with this but we should have at least 2 people to approve before merge.

FreakyMaryk:
For me looks good. I checked all options, with conf file and with app startup. But, if you have both conf and startup option, than startup option is stronger. Is it ok? @riordant

PR #358, opened 2019-03-04
Fix for txHash being assigned too early

riordant:
The value of `hashTx` in `CZerocoinSpendEntry` objects that are used for multi-spends is incorrect, as `txHash` is assigned the `wtxNew` hash value before it's initialised as a transaction object.
This PR assigns the value after creation.
(Not a critical fix; `hashTx` is for storage purposes only and not used in any processing at this time).

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -4320,9 +4320,6 @@ bool CWallet::CreateMultipleZerocoinSpendTransaction(std::string &thirdPartyaddr
                 }
             }
 
-            txHash = wtxNew.GetHash();
-            LogPrintf("wtxNew.txHash:%s\n", txHash.ToString());
-
             // Embed the constructed transaction data in wtxNew.
             *static_cast<CTransaction *>(&wtxNew) = CTransaction(txNew);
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -4320,9 +4320,6 @@ bool CWallet::CreateMultipleZerocoinSpendTransaction(std::string &thirdPartyaddr
                 }
             }
 
-            txHash = wtxNew.GetHash();
-            LogPrintf("wtxNew.txHash:%s\n", txHash.ToString());
-
             // Embed the constructed transaction data in wtxNew.
             *static_cast<CTransaction *>(&wtxNew) = CTransaction(txNew);
```
psolstice: This is really bad practice even if it works. Additional problem is in bitcoin 0.14+ `CWalletTx` and `CTransaction` classes are unrelated.

ultimaweapon: The problems is this is a code from original `CreateTransaction` function which also do the same like this. If we want to fix this we may need to refactor `CWalletTx`.

PR #360, opened 2019-03-11
Disable building tests by default

riordant:
Currently tests are built by default where the host is Linux-based. This PR disables building tests by default. 
We also have a configure flag to disable tests, `--disable-tests`. Given the new functionality, I propose we change this to `--enable-tests`. This is also updated here.
Fixes https://github.com/zcoinofficial/zcoin/issues/261


configure.ac
```diff
@@ -106,17 +106,10 @@ AC_ARG_ENABLE([upnp-default],
...
-     zusetests=yes
-  ;;
-esac
-
 AC_ARG_ENABLE(tests,
-    AS_HELP_STRING([--disable-tests],[do not compile tests (default is to compile)]),
-    [use_tests=$enableval],
-    [use_tests=$zusetests])
+    AS_HELP_STRING([--enable-tests],[compile tests (default is not to compile)]),
+    [use_tests=$withval],
```
ultimaweapon: Seem like `$withval` should be using with `--with-xxx` instead. For `--enable-xxx` it look like we need to use `$enableval` instead.

a-bezrukov:
Disabled the tests, made `make check` error when tests were disabled and did a minor refactoring.

src/Makefile.am
```diff
@@ -613,6 +613,9 @@ endif
 
 if ENABLE_TESTS
 include Makefile.test.include
+else
+check-local:
```
ultimaweapon: Not sure if this going to run first when we run `make check`. I take a look at `check-local` in `src/Makefile.test.include` and compare with test output found out it run after all unit tests have been ran.

a-bezrukov: Makefile.test.include is not included when we configured without tests. This check-local purpose is to fail the tests when we configured without tests.

riordant: @ultimaweapon Are you ok with Andrey's reply?

PR #365, opened 2019-04-18
Sigma transition fixes

ultimaweapon:
The issue is right now it will reject block that contains Sigma mint, which is not right since `scriptPubKey` can contains anything. The right way is to ignore it instead. This PR also refactor some code too. Also, we don't want to mint Sigma with Zerocoin GUI so I fix that too.


src/zerocoin.cpp
```diff
@@ -528,17 +528,33 @@ bool CheckZerocoinTransaction(const CTransaction &tx,
...
+    {
+        LOCK(cs_main);
+        realHeight = chainActive.Height();
+    }
+
     // Check Mint Zerocoin Transaction
-    BOOST_FOREACH(const CTxOut &txout, tx.vout) {
+    for (const CTxOut &txout : tx.vout) {
         if (!txout.scriptPubKey.empty() && txout.scriptPubKey.IsZerocoinMint()) {
+            if (realHeight > params.nSigmaStartBlock + params.nZerocoinV2MintGracefulPeriod) {
```
martun: I intentionally did not put this check here. The problem is this function is called both from ConnectBlock and AcceptToMemoryPoolWorker. And there are already checks for this. We needed 2 different checks, one for mempool addition, another for blocks.

src/main.cpp
```diff
@@ -1240,37 +1240,35 @@ bool AcceptToMemoryPoolWorker(
...
-        bool V2ZerocoinSpendsAllowedInMempool = (chainActive.Height() <= Params().GetConsensus().nZerocoinV2SpendStopBlock);
+        bool allow = (chainActive.Height() <= consensus.nSigmaStartBlock + consensus.nZerocoinV2SpendMempoolGracefulPeriod);
 
-        if (!V2ZerocoinSpendsAllowedInMempool)
+        if (!allow) {
             return state.DoS(100, error("Old zerocoin spends no more allowed in mempool"),
                              REJECT_INVALID, "bad-txns-zerocoin");
+        }
     }
-    if (tx.IsZerocoinSpendV3() || tx.IsZerocoinMintV3()) {
```
martun: By removing these checks you allowed both sigma Spend and mint in previous blocks. Maybe we just agree to set the nSigmaStartBlock to a block in the past, and increate zerocoin mint gracefull period for testnet? And not touch the code for mainnet?

thebevrishot: I think logic here do not allow to accept the block that we have a problem with. the logic that effect to the block is in `ConnectBlock` and I agree to accept the block even the block number is lower than `nSigmaStartBlock` because it may have some script like this in mainnet which effect us to stuck at some block.

PR #366, opened 2019-04-19
Sigma fix check transaction

thebevrishot:
https://trello.com/c/hdHd09g4/384-fix-user-can-use-same-serial-in-single-transaction


src/test/sigma_partialspend_mempool_tests.cpp
```diff
@@ -278,4 +278,66 @@ BOOST_AUTO_TEST_CASE(partialspend_remint) {
...
+    std::string stringError;
+    std::vector<std::pair<std::string, int>> denominationPairs = {{"1", 2}};
+    BOOST_CHECK_MESSAGE(pwalletMain->CreateZerocoinMintModel(
+        stringError, denominationPairs, SIGMA), stringError + " - Create Mint failed");
+
+    BOOST_CHECK_MESSAGE(mempool.size() == 1, "Mint was not added to mempool " + stringError);
+    CreateAndProcessBlock({}, scriptPubKey);
+    BOOST_CHECK_MESSAGE(mempool.size() == 0, "Mempool was not cleared");
+
+    // Add 5 more blocks and verify that Mint can not be spent until 6 blocks verification
```
ultimaweapon: Seem like comments is not correct since with the following line the total confirmation will be 6.

src/test/sigma_partialspend_mempool_tests.cpp
```diff
@@ -278,4 +278,66 @@ BOOST_AUTO_TEST_CASE(partialspend_remint) {
...
+
+    // Expect 2 spends
+    BOOST_CHECK_EQUAL(tx.vin.size(), 2);
+
+    // construct double spend transaction
+    auto extendedScript = CScript();
+    extendedScript.insert(extendedScript.end(), tx.vin[0].scriptSig.begin(), tx.vin[0].scriptSig.end());
+
+    const std::string extended = "test";
+    extendedScript.insert(extendedScript.end(), extended.begin(), extended.end());
```
ultimaweapon: We should put a comment to tell why we need to append `test` string at the end.

src/zerocoin_v3.cpp
```diff
@@ -167,6 +168,12 @@ bool CheckSpendZcoinTransactionV3(
 					return false;
 			}
 
+			// check duplicated serials in same transaction.
+			if (!txSerials.insert(serial).second) {
+				return state.DoS(0,
```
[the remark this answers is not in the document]
thebevrishot: This is from old protocol but this value is effect to ban node and this look critical it's obvious to set to 100.

src/test/sigma_partialspend_mempool_tests.cpp
thebevrishot: obvious

src/test/sigma_partialspend_mempool_tests.cpp
thebevrishot: obvious

thebevrishot:
@ultimaweapon  fixed

src/zerocoin_v3.cpp
```diff
@@ -167,6 +168,12 @@ bool CheckSpendZcoinTransactionV3(
 					return false;
 			}
 
+			// check duplicated serials in same transaction.
```
martun: This change is not necessary. We have a LARGER problem here. What we must do is move zerocoinTxInfoV3->zcTransactions.insert(hashTx); from line 182 to the end of this function. The problem we have here is the first spend will be verified by "CheckZerocoinSpendSerialV3", and we'll verify that spend is not re-used. Then "zerocoinTxInfoV3->zcTransactions.insert(hashTx);" happens, and for all subsequent spends their serials are not checked. I.E. the attacker can actually create a transaction, where the first spend is valid, and the other spends re-use the same serial, and it will pass :) Looks like we have the same problem for the old zerocoin, and we're lucky no-one used it against us.

thebevrishot: we ever be attacked with this bug before. and we fix this way on Zerocoin. your solution look good.  I'll try.

thebevrishot: @martun  If do that the transaction will be accepted to mempool but will be rejected when create a block. I think we should check in mempool level.

martun: @thebevrishot No, why? During adding to mempool, this line will run: zerocoinTxInfoV3->zcTransactions.insert(hashTx);

Then this transaction will not be re-checked while being added to the blocks.

thebevrishot: function AcceptToMemoryPoolWorker do not pass zerocoinTxInfoV3 then `zerocoinTxInfoV3->zcTransactions.insert(hashTx);` would not be call.

riordant: @martun Was that not the same attack that occurred following the MTP launch?

martun: @riordant Yes, this is similiar to that attack, but I do not believe you have fixed the code properly that time.

I think even while adding the mempool we still want to check if the serial number happened before, but we can set isVerifyDB, such that serials will not be recorded as spent.

And this line still needs to get out of the for loop: zerocoinTxInfoV3->zcTransactions.insert(hashTx);

Otherwise all the other spends, except the first one will not be checked while being added to the blocks.

thebevrishot: for the case serial duplicate with confirmed serial is checked in function `AcceptToMemoryPoolWorker`(line: 1357). 

I agree with move `zerocoinTxInfoV3->zcTransactions.insert(hashTx);` out of loop. It do not make sense to insert single hashTx many time.

nopslide: @FreakyMaryk please create testcases that cover this kind of attack as well

src/zerocoin_v3.cpp
```diff
@@ -107,8 +107,10 @@ bool CheckSpendZcoinTransactionV3(
 		int nHeight,
 		bool isCheckWallet,
 		CZerocoinTxInfoV3 *zerocoinTxInfoV3) {
+	int txHeight = chainActive.Height();
```
ultimaweapon: We need to lock `cs_main` before access `chainActive`.

thebevrishot: hmm, I got this after merge. but seem like it never be used.

PR #371, opened 2019-04-22
Add LGTM configuration

ultimaweapon:
This will make https://lgtm.com work correctly on this project.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,27 @@
+path_classifiers:
+  docs:
+    - "exclude: /"
+    - "ZNODE.md"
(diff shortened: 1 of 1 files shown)


riordant:
Untested ACK

ultimaweapon:
After take a look in LGTM logs it fail to build our project so that why it still cannot analyze our C++ code. I will take a look deeply into the problems.

ultimaweapon:
I can't make it work for C++. Seem like it need to merge into default branch first so I will close this PR and move it to `master` instead. Then we can merge `master` into `sigma`.

PR #373, opened 2019-04-22
Fixed 2 deadlocks

a-bezrukov:
Deadlocks appeared due to cs_wallet was acquired first and then cs_main was second. The usual way is vice versa.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -11,7 +11,6 @@ using namespace std;
  * CChain implementation
  */
 void CChain::SetTip(CBlockIndex *pindex) {
-    std::lock_guard<std::mutex> _(chain_mutex);
     if (pindex == NULL) {
         vChain.clear();
         return;
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -11,6 +11,7 @@ using namespace std;
  * CChain implementation
  */
 void CChain::SetTip(CBlockIndex *pindex) {
+    std::lock_guard<std::mutex> _(chain_mutex);
     if (pindex == NULL) {
         vChain.clear();
         return;
--- a/src/chain.h
+++ b/src/chain.h
@@ -23,6 +23,7 @@
 
 #include <vector>
 #include <unordered_set>
+#include <mutex>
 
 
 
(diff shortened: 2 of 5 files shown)


a-bezrukov:
@ultimaweapon, this was my initial idea - that I should avoid introducing a new sync object and acquire the cs_main before cs_wallet. I would need to put LOCK(cs_main) into CWalletDB::LoadWallet (walletdb.cpp:744), and by that cover almost all LoadWallet with cs_main, which is a lot of code and it takes pretty much time during startup. I thought that introducing a new sync object is a ways less evil, than creating a bottle neck like this.

ultimaweapon:
Do you know why it cause dead lock? From what I saw the only call path is from startup which GUI should not concurrent try to lock both `cs_main` and `cs_wallet`. In that case I think it is okay to lock at that level since it will happen only at startup, which should not effect any performance since there is no parallel initialization.

a-bezrukov:
That's what the debug outputs: 
`
2019-04-24 14:26:36 Previous lock order was:
2019-04-24 14:26:36  (1) cs_main  main.cpp:3848
2019-04-24 14:26:36  (1) cs_main  wallet/wallet.cpp:1144
2019-04-24 14:26:36  (2) cs_wallet  wallet/wallet.cpp:1144
2019-04-24 14:26:36 Current lock order is:
2019-04-24 14:26:36  (2) pwallet->cs_wallet  wallet/walletdb.cpp:768
2019-04-24 14:26:36  (1) cs_main  zerocoin.cpp:570
`
It's easy to reproduce if run on a wallet with transactions (I download it from the insight)

ultimaweapon:
Okay I see how it prevent dead lock now. So it not actually a dead lock but just potential due to it lock in different order. I still think we should lock `cs_main` from wallet site due to add additional lock just for this purpose is overkill. Maybe need to ask other people to see, @martun @thebevrishot what do you guys think?

a-bezrukov:
It is not a potential deadlock :) It locks 80% of the time and I need to kill -9 as nothing else working. Assertion is failing when it is --enable-debug, if not it just hangs.

martun:
I think having a separate mutex for this is the right thing to do. Looks nicer, does not prevent us from doing something else. On the other hand locking cs_main all the time does not look nice.

ultimaweapon:
Okay if you guys think it is the right way I will remove my request changes so this PR can be merge after one additional approve.

thebevrishot:
take a look. do not agree to.

reubenyap:
As per the discussion, do not use any additional mutexes and to instead just lock two mutexes at once.

PR #374, opened 2019-04-23
Add LGTM configuration

ultimaweapon:
This is replacement for #371 due to I can't make it work on non-default branch.


ultimaweapon:
This required for Sigma but it will merge into `master` first then we will merge `master` into `sigma`.

ultimaweapon:
@riordant @a-bezrukov could you guys please review? We can ignore the CI fail for `master` branch since it is know issue.

PR #375, opened 2019-04-23
Disable mint functionalities until Sigma is activated

ultimaweapon:


thebevrishot:
please provide result image.

src/wallet/rpcwallet.cpp
```diff
@@ -2714,6 +2714,16 @@ UniValue mint(const UniValue& params, bool fHelp)
             "\nAutomatically choose denominations to mint by amount\n" +
             HelpRequiringPassphrase());
 
+    // Ensure Sigma mints is already accepted by network so users will not lost their coins
+    // due to other nodes will treat it as garbage data.
+    {
+        LOCK(cs_main);
```
martun: Is it necessary to lock cs_main here?

ultimaweapon:
@a-bezrukov okay good idea. I agree.

src/wallet/rpcwallet.cpp
ultimaweapon: Yes because we want to access `chainActive`.

src/zerocoin_v3.cpp
```diff
@@ -51,6 +51,17 @@ static bool CheckZerocoinSpendSerialV3(
 	return true;
 }
 
+bool IsSigmaAllowed()
+{
+	LOCK(cs_main);
```
martun: After we merge this one, this lock is not necessary any more right? https://github.com/zcoinofficial/zcoin/pull/373/files

ultimaweapon: If we continue with the way in that PR, yes.

PR #378, opened 2019-04-24
Fix db read/write on sigma data

thebevrishot:
https://trello.com/c/OfleGZDg/431-sigma-testnet-does-not-accept-testbootstrapzcoinio-when-reindexing-it-restarts-from-scratch-despite-having-all-the-blocks-in-tes

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -466,7 +466,7 @@ public:
             READWRITE(spentSerials);
 	    }
 
-        if (!(nType & SER_GETHASH)) {
+        if (!(nType & SER_GETHASH) && nHeight >= Params().GetConsensus().nSigmaStartBlock) {
             READWRITE(mintedPubCoinsV3);
             READWRITE(spentSerialsV3);
         }


src/chain.h
```diff
@@ -466,7 +466,7 @@ class CDiskBlockIndex : public CBlockIndex
             READWRITE(spentSerials);
 	    }
 
-        if (!(nType & SER_GETHASH)) {
+        if (!(nType & SER_GETHASH) && nHeight >= Params().GetConsensus().nSigmaStartBlock) {
             READWRITE(mintedPubCoinsV3);
```
martun: Is it a problem if these maps are empty? I believe they were empty, and must be still serialized without a problem no?

thebevrishot: serialization support empty map. you can take a look serialization function at line 1056 in file `src/serialize.h`.

martun: ok, got it. thanks!

PR #380, opened 2019-04-25
Always check for Zerocoin transaction

ultimaweapon:
This will make it behavior to be the same as `master` branch. Actually this is the source of problems why we make a mistake when review Zerocoin disable release.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1181,6 +1181,7 @@ bool CheckTransaction(
 			    return state.DoS(10, false, REJECT_INVALID, "bad-txns-prevout-null");
 		    }
 	    }
+
         if (tx.IsZerocoinV3SigmaTransaction()) {
             if (!CheckZerocoinTransactionV3(
                     tx,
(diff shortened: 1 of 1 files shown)


martun:
Can you describe why this is necessary? Maybe add a comment near the change to describe it?

ultimaweapon:
@martun this will make it behavior to be the same as previous (see here https://github.com/zcoinofficial/zcoin/blob/master/src/main.cpp#L1165). I think we don't need to put a comment in the code since the comment will not correct once it merge into master.

martun:
@ultimaweapon Yes, but the code is more logical without this change :) Maybe we try to fix somewhere else?

ultimaweapon:
I agree with you that it look better when there is a guard. The only issue for me is it alter current behavior for Zerocoin on master branch. What I think is we should not alter it in Sigma branch due to it will unexpected by other people that are not working on Sigma and also make Sigma team misunderstand too like it happen with our Zerocoin disable patch. So if we want to alter it behavior we should make it to master branch instead of Sigma branch.

PR #381, opened 2019-04-25
Postponing Sigma on testnet

a-bezrukov:


ultimaweapon:
Do you have any reason why we need to do that?

a-bezrukov:
To test ZC mints and spends

ultimaweapon:
I see. If we want to do this way we need to allow previous Sigma mint but not new mints to be process by `CheckZerocoinTransactionV3` too otherwise the state will be incorrect. I think we have more better way. Instead we will remove Zerocoin transition code and allow Zerocoint until block 100,000 for testnet.

ultimaweapon:
After discussion found out that we already deployed this patch and no any Sigma mint until current height. So we need to merge this.

ultimaweapon:
After discussion we don't need this anymore.

PR #385, opened 2019-04-29
Update zmq to 4.3.1

riordant:
Addresses https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-6250
Credit @Archimedes1618: "although the version you use is lower than the affected ones,  it might be worth thinking about using a newer version"
This is the same version Bitcoin now uses

### DIFF
--- a/src/zmq/zmqnotificationinterface.cpp
+++ b/src/zmq/zmqnotificationinterface.cpp
@@ -72,10 +72,14 @@ CZMQNotificationInterface* CZMQNotificationInterface::CreateWithArguments(const 
 // Called at startup to conditionally set up ZMQ socket(s)
 bool CZMQNotificationInterface::Initialize()
 {
+    int major = 0, minor = 0, patch = 0;
+    zmq_version(&major, &minor, &patch);
+    LogPrint("zmq", "zmq: version %d.%d.%d\n", major, minor, patch);
+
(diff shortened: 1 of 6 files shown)


ultimaweapon:
The patch look good to me. I'm currently building it to see if it success on all platforms.

src/zmq/zmqnotificationinterface.cpp
```diff
@@ -118,7 +122,7 @@ void CZMQNotificationInterface::Shutdown()
             LogPrint("zmq", "   Shutdown notifier %s at %s\n", notifier->GetType(), notifier->GetAddress());
             notifier->Shutdown();
         }
-        zmq_ctx_destroy(pcontext);
+        zmq_ctx_term(pcontext);
 
         pcontext = 0;
```
thebevrishot: better to use `nullptr`.

riordant: Ok. This line was taken verbatim from the Bitcoin Core upgrade

riordant:
@ultimaweapon This PR is passing on the Jenkins page, but says it's failing here. Do you know why?

ultimaweapon:
@riordant it required all of checks to be successfully in order to get a green check mark. What we need is only `pr-merge` which already succeeded.

PR #389, opened 2019-04-30
small fix rpc

thebevrishot:
https://trello.com/c/MTlEcteu/413-seems-like-we-did-not-validate-the-situation-with-spend-many-like-d05
https://trello.com/c/og4WYbuU/338-remove-old-sigma-mint-spend-transaction-creation-functions
https://trello.com/c/pt16UgX4/373-add-description-for-rpc-mint


src/wallet/rpcwallet.cpp
```diff
@@ -2716,7 +2716,7 @@ UniValue mint(const UniValue& params, bool fHelp)
 
     if (fHelp || params.size() != 1)
         throw std::runtime_error(
-            "mint amount\n"
+            "mint <amount>(could not smaller than 0.1)\n"
```
riordant: `mint <amount>(0.1, 0.5, 1, 10, 100)`

src/wallet/rpcwallet.cpp
```diff
@@ -2840,79 +2840,6 @@ UniValue mintzerocoin(const UniValue& params, bool fHelp)
 
 }
 
-UniValue mintzerocoinV3(const UniValue& params, bool fHelp)
```
riordant: Agree. Broken functions

src/wallet/rpcwallet.cpp
```diff
@@ -3501,6 +3192,10 @@ UniValue spendmany(const UniValue& params, bool fHelp) {
 
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
+    // Only account "" have sigma coins.
```
riordant: Disagree. Should remove `account` option altogether. Otherwise it's a useless parameter

src/wallet/rpcwallet.cpp
ultimaweapon: We can put any value and it will automatically divide into supported mints. So that why it can be any value as long as it can multiply with the lowest denomination.

ultimaweapon: I think we should use the same format as Bitcoin RPC. So it should be something like `mint amount\n`

src/wallet/rpcwallet.cpp
```diff
@@ -2716,7 +2716,7 @@ UniValue mint(const UniValue& params, bool fHelp)
 
     if (fHelp || params.size() != 1)
         throw std::runtime_error(
-            "mint amount\n"
+            "mint <amount>(could not smaller than 0.1)\n"
             "\nAutomatically choose denominations to mint by amount\n" +
```
ultimaweapon: Other Bitcoin RPC don't have `\n` at the end.

src/wallet/rpcwallet.cpp
```diff
@@ -2716,7 +2716,7 @@ UniValue mint(const UniValue& params, bool fHelp)
...
             HelpRequiringPassphrase());
```
ultimaweapon: Other Bitcoin RPC have additional `\n` after `HelpRequiringPassphrase()`.

ultimaweapon: We need to add more description too. Take a look at `sendmany` RPC for some example.

src/wallet/rpcwallet.cpp
```diff
@@ -3501,6 +3192,10 @@ UniValue spendmany(const UniValue& params, bool fHelp) {
 
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
+    // Only account "" have sigma coins.
+    if (!params[0].isNull() && !params[0].get_str().empty())
```
ultimaweapon: I think we should remove null checking since it confusing due to this one it not optional parameter.

src/wallet/rpcwallet.cpp
ultimaweapon: The reason we make this RPC to use the same format as `sendmany` because we want to make it the same except just RPC name. That why we still keep that parameter even we did not use it. @reubenyap do you have any idea?

src/wallet/rpcwallet.cpp
riordant: Ah ok, sure. In that case, I would just like to change it to be grammatically correct:
`"mint <amount>(No less than 0.1)\n"`

src/wallet/rpcwallet.cpp
riordant: @ultimaweapon Why must it be the same? By that logic, users will think that they HAVE the option of using a different account. At the very least, the error message should be changed to something along the lines of `Account is not usable with Sigma coins` or so

src/wallet/rpcwallet.cpp
riordant: Shouldn't this also be in a `try` block? If `get_str()` fails it will throw

src/wallet/rpcwallet.cpp
ultimaweapon: The benefit we have to make it the same is:

1. User just need to change the only RPC name to switch between normal send and private send.
2. It will easier for scripting too since it just need to change only RPC name to switch between normal send and private send.

Usually the RPC is for advance user and programming so I think it make sense to go with this way. The account name for `sendmany` also already deprecated and it accept only empty string too. The reason it still there instead of remove is because backward compatibility. The solution that Bitcoin use is they provide the description about this parameter that it is already deprecated and accept only empty string.

I believe the reason we use insufficient fund error because we want to mimic the behavior of `sendmany`.

riordant: Ok, from the perspective of ease-of-use when scripting, I agree with you. People using it manually (ie. most likely programmers) will figure it out I guess.

src/wallet/rpcwallet.cpp
riordant: Agree, the latter check is enough. We are already assuming that each param element is the correct type and exists

src/wallet/rpcwallet.cpp
thebevrishot: GREAT! thanks for correction.

src/wallet/rpcwallet.cpp
thebevrishot: thanks Putta for explanation.

src/wallet/rpcwallet.cpp
thebevrishot: After discussion it's better to add to arguments description instead.

thebevrishot: make sense.

src/wallet/rpcwallet.cpp
thebevrishot: some Bitcoin RPC command have `\n` but seem like most of them have no.

src/wallet/rpcwallet.cpp
```diff
@@ -2716,9 +2716,18 @@ UniValue mint(const UniValue& params, bool fHelp)
...
     if (fHelp || params.size() != 1)
         throw std::runtime_error(
-            "mint <amount>(could not smaller than 0.1)\n"
-            "\nAutomatically choose denominations to mint by amount\n" +
-            HelpRequiringPassphrase());
+            "mint amount\n"
+            "\nAutomatically choose denominations to mint by amount."
+            + HelpRequiringPassphrase() + "\n"
+            "\nArguments:\n"
+            "1. \"amount\"      (numeric or string, required) The amount in " + CURRENCY_UNIT + " to mint but must not less than 0.1\n"
```
ultimaweapon: I think `must not less than 0.1` is not correct. It must be something like `must be able to multiply with 0.1`.

src/wallet/rpcwallet.cpp
```diff
@@ -3193,7 +3202,8 @@ UniValue spendmany(const UniValue& params, bool fHelp) {
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
     // Only account "" have sigma coins.
-    if (!params[0].isNull() && !params[0].get_str().empty())
+    string strAccount = AccountFromValue(params[0]);
```
ultimaweapon: Namespace issue.

thebevrishot:
@riordant Could you please review again?

src/wallet/rpcwallet.cpp
martun: Guys do we need to delete this? I can try to fix them. We don't have any RPC functions to substitute these right?

thebevrishot: we use `mint` instead.

riordant: Exactly. `mint` and `spendmany` are enough

FreakyMaryk:
@thebevrishot Looks like build failed, cause you broke the python test, please fix it

PR #390, opened 2019-05-01
Added consensus parameters to change dandelion params for testnet and regtests

martun:


src/net.cpp
```diff
@@ -1971,7 +1971,8 @@ bool OpenNetworkConnection(
         LOCK(cs_vNodes);
         // Dandelion: new outbound connection
         CNode::vDandelionOutbound.push_back(pnode);
-        if (CNode::vDandelionDestination.size() < DANDELION_MAX_DESTINATIONS) {
+        auto consensus = Params().GetConsensus();
```
a-bezrukov: Here a non-const copy of Consensus are created and the `consensus` object can be changed later. `const auto &` works ways better here as it will store the reference that `const Consensus::Params& GetConsensus() const` returns.

martun: Agree!

src/wallet/wallet.cpp
```diff
@@ -2143,7 +2143,7 @@ CAmount CWallet::SelectMintCoinsForAmount(
         const std::vector<sigma::CoinDenominationV3>& denominations,
         std::vector<sigma::CoinDenominationV3>& coinsOut) {
     CAmount val = required;
-    for (int i = 0; i < denominations.size(); i++)
```
ultimaweapon: Nice one for fixing warning!

src/main.cpp
```diff
@@ -7027,15 +7027,11 @@ bool static ProcessMessage(CNode *pfrom, string strCommand,
                              stempool.size(),
                              stempool.DynamicMemoryUsage() / 1000);
                     int64_t nCurrTime = GetTimeMicros();
-                    int64_t nEmbargo = 1000000 * DANDELION_EMBARGO_MINIMUM +
-                        PoissonNextSend(nCurrTime, DANDELION_EMBARGO_AVG_ADD);
+                    auto consensus = Params().GetConsensus();
```
ultimaweapon: The `consensus` will be copied. Will be better to change to `auto&`.

PR #393, opened 2019-05-03
Fix CheckBlock function return true when throw exception

thebevrishot:
https://trello.com/c/qpDbib7F/429-fix-checkblock-to-return-false-when-there-is-exception


riordant:
LGTM

ultimaweapon:
@thebevrishot Just notice the one is fail is branch building not merging.

PR #394, opened 2019-05-03
Fix unspent mints status

a-bezrukov:
This restores mints status to unspent after zapwallettxes removed the unconfirmed spends. The spend serial information is given by the blockchain.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -5586,24 +5586,63 @@ bool CWallet::InitLoadWallet() {
         nWalletDBUpdated++;
 
         // Restore wallet transaction metadata after -zapwallettxes=1
-        if (GetBoolArg("-zapwallettxes", false) && GetArg("-zapwallettxes", "1") != "2") {
-            CWalletDB walletdb(walletFile);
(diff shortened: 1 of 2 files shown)


a-bezrukov:
I dedicated quite some time to testing it. My main test scenario was:

src/wallet/wallet.cpp
```diff
@@ -5586,24 +5586,63 @@ bool CWallet::InitLoadWallet() {
         nWalletDBUpdated++;
 
         // Restore wallet transaction metadata after -zapwallettxes=1
-        if (GetBoolArg("-zapwallettxes", false) && GetArg("-zapwallettxes", "1") != "2") {
-            CWalletDB walletdb(walletFile);
+        if (GetBoolArg("-zapwallettxes", false)) {
```
martun: We probably need to do something similar for sigma. Could you please create a PR for sigma branch, or just add a trello card?

a-bezrukov: Here it is: https://trello.com/c/KNSHlfOd/487-zapwallettxes-should-correctly-restore-status-of-sigma-spends

PR #398, opened 2019-05-03
Refactor, rename all v3 to sigma

martun:


src/wallet/rpcwallet.cpp
```diff
@@ -3213,7 +3213,7 @@ UniValue spendallzerocoin(const UniValue& params, bool fHelp) {
     return  hasUnspendableMints;
 }
 
-UniValue spendzerocoinV3(const UniValue& params, bool fHelp) {
+UniValue spendSigma(const UniValue& params, bool fHelp) {
```
riordant: In keeping with the standard for RPC functions, they should all be lower case.

thebevrishot:
@martun please fix conflicts.

src/wallet/rpcwallet.cpp
martun: done

PR #400, opened 2019-05-05
Fixed Znode sync issue for preventing mining

nopslide:
I setup 13 nodes for local mainnet, turnout I cannot mining due to znode not synce. I don't have coin to create 10 znodes. 
After looking at the code, saw this portion of code make me stuck.


riordant:
I think @psolstice is best to review this one

qashban4:

PR #401, opened 2019-05-06
Checking that there are znodes in network

nopslide:
Otherwise, it will prevent setgernerate command to create blocks

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -1121,6 +1121,7 @@ void static ZcoinMiner(const CChainParams &chainparams) {
                 do {
                     bool fvNodesEmpty;
                     bool fHasZnodesWinnerForNextBlock;
+                    bool fHasZnodes;
                     const Consensus::Params &params = chainparams.GetConsensus();
                     {
                         LOCK(cs_vNodes);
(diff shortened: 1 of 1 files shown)


reubenyap:
Not needed according to @psolstice

nopslide:
I think we need to take care in case Zcoin doesn't have any znodes running, and it will preventing solo miner to run command for creating blocks here.

src/miner.cpp
```diff
@@ -1129,10 +1129,15 @@ void static ZcoinMiner(const CChainParams &chainparams) {
...
-                                chainActive.Height() < params.nZnodePaymentsStartBlock ||
-                                mnodeman.GetNextZnodeInQueueForPayment(chainActive.Height(), true, nCount);
+                        // Check you actually have znodes in network
+			            if(mnodeman.CountEnabled() > 0){
+                        	fHasZnodesWinnerForNextBlock =
+                                	params.IsRegtest() ||
+                                	chainActive.Height() < params.nZnodePaymentsStartBlock ||
+                                	mnodeman.GetNextZnodeInQueueForPayment(chainActive.Height(), true, nCount);
+                        }else{
+				            fHasZnodesWinnerForNextBlock = true; // we don't have Znode winner causes we don't have Znode on network
```
thebevrishot: Please use new flag instead. Setting `fHasZnodesWinnerForNextBlock` as true isn't make sense.

riordant: @thebevrishot added extra param, please re-review

qashban4:

PR #402, opened 2019-05-06
Pay reward back to miners when don't have znode

nopslide:
Reward should belong to the miner who are creating a block during network have zero znodes enabled

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -529,8 +529,10 @@ CBlockTemplate* BlockAssembler::CreateNewBlock(
         if (nHeight >= chainparams.GetConsensus().nZnodePaymentsStartBlock) {
             const Consensus::Params &params = chainparams.GetConsensus();
             CAmount znodePayment = GetZnodePayment(chainparams.GetConsensus(), nHeight > 0 && nBlockTime >= params.nMTPSwitchTime);
-            coinbaseTx.vout[0].nValue -= znodePayment;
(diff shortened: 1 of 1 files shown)


src/miner.cpp
```diff
@@ -529,8 +529,10 @@ CBlockTemplate* BlockAssembler::CreateNewBlock(
         if (nHeight >= chainparams.GetConsensus().nZnodePaymentsStartBlock) {
             const Consensus::Params &params = chainparams.GetConsensus();
             CAmount znodePayment = GetZnodePayment(chainparams.GetConsensus(), nHeight > 0 && nBlockTime >= params.nMTPSwitchTime);
-            coinbaseTx.vout[0].nValue -= znodePayment;
-            FillBlockPayments(coinbaseTx, nHeight, znodePayment, pblock->txoutZnode, pblock->voutSuperblock);
+            if(mnodeman.CountEnabled() > 0){
```
martun: I'm not familiar with znodes. Can it happen, that mnodeman.CountEnabled() > 0, but they did not agree on the znode winner, and we still want to pay it to miner?

nopslide: it checked either we have znodes running on network or not

reubenyap:
Not to touch this until Znode revamp.

nopslide:
it should not be close, but open until we release sigma and look back at it later.

qashban4:

PR #403, opened 2019-05-06
 script check fix for sigma && validation checks

mattt21:


src/main.cpp
```diff
@@ -1449,7 +1449,7 @@ bool AcceptToMemoryPoolWorker(
             view.SetBackend(dummy);
         } // LOCK
 
-        if (!tx.IsZerocoinSpend() && fCheckInputs) {
+        if (!tx.IsZerocoinSpend() && !tx.IsZerocoinSpendV3() && fCheckInputs) {
```
ultimaweapon: We need Sigma spend to go into this block due to it have to pay fee (it not feeless like Zerocoin spend).

src/primitives/transaction.cpp
```diff
@@ -205,7 +205,7 @@ double CTransaction::ComputePriority(double dPriorityInputs, unsigned int nTxSiz
 
 bool CTransaction::IsCoinBase() const
 {
-    return (vin.size() == 1 && vin[0].prevout.IsNull() && (vin[0].scriptSig.size() == 0 || vin[0].scriptSig[0] != OP_ZEROCOINSPEND));
+    return (vin.size() == 1 && vin[0].prevout.IsNull() && (vin[0].scriptSig.size() == 0 || (vin[0].scriptSig[0] != OP_ZEROCOINSPEND && vin[0].scriptSig[0] != OP_ZEROCOINSPENDV3)));
```
ultimaweapon: The reason we did not check if `scriptSig` is Sigma spend or not because `vin[0].prevout.IsNull()` will be false so it does not need to check `scriptSig`.

PR #404, opened 2019-05-07
Fixed every-time resync

a-bezrukov:


src/zerocoin_v3.cpp
```diff
@@ -313,9 +313,9 @@ bool CheckZerocoinTransactionV3(
     auto& consensus = Params().GetConsensus();
 
     // nHeight have special mode which value is INT_MAX so we need this.
-    int realHeight;
+    int realHeight = consensus.nSigmaStartBlock;
 
-    {
+    if(!(isVerifyDB && nHeight == INT_MAX)) {
```
martun: Can't understand this change. I can see that this function is called from 3 places, once when adding to Mempool, and twice again when adding to block (not sure why we double-check this :) ). If we're adding to mempool, then isVerifyDB=false, and nHeight = INT_MAX. When called from **ConnectBlock**, isVerifyDB=false and nHeight=pindex->nHeight (real height of the block). When called from CheckBlock, isVerifyDB=false/true, height is the real height of the block.

So if we're checking a transaction in a block, then nHeight != INT_MAX, I.E. we will enter this if statement, if transaction tries to enter mempool, then isVerifyDB == false, and it will again enter this if statement.

Looks like this if statement will always be entered, I.E. your better delete it.

a-bezrukov: It is called from walletdb.cpp:544, and this exact call made it rechecking everytime: `if (!(CheckTransaction(wtx, state, wtx.GetHash(), true, INT_MAX, false) && (wtx.GetHash() == hash)`

src/zerocoin_v3.cpp
```diff
@@ -313,9 +313,9 @@ bool CheckZerocoinTransactionV3(
     auto& consensus = Params().GetConsensus();
 
     // nHeight have special mode which value is INT_MAX so we need this.
-    int realHeight;
+    int realHeight = consensus.nSigmaStartBlock;
```
thebevrishot: I think it's unnecessary because If realHeight < 50k zerocoin mints checking in block line 326 will not check.

a-bezrukov: Right!

src/zerocoin.cpp
```diff
@@ -564,9 +564,9 @@ bool CheckZerocoinTransaction(const CTransaction &tx,
                               CZerocoinTxInfo *zerocoinTxInfo)
 {
     // nHeight have special mode which value is INT_MAX so we need this.
-    int realHeight;
+    int realHeight = params.nSigmaStartBlock;
 
-    {
+    if(!(isVerifyDB && nHeight == INT_MAX)) {
```
ultimaweapon: I think this seem like a hack. After investigate I found the expression `isVerifyDB && nHeight == INT_MAX` will be `true` only when loading wallet. Can we do something like this instead?

```cpp
bool isWalletCheck = (isVerifyDB && nHeight == INT_MAX);
```

and then change bottom logics to something like this without initialize `realHeight` with `params.nSigmaStartBlock`:

```cpp
if (!isWalletCheck && realHeight > params.nSigmaStartBlock + params.nZerocoinV2MintGracefulPeriod) {
```

```cpp
if (!isWalletCheck && realHeight > params.nSigmaStartBlock + params.nZerocoinV2SpendGracefulPeriod) {
```

a-bezrukov: The way it is implemented now saves re-entering the cs_main. I am keeping this property.

a-bezrukov:
@martun Yes, we need it still. Merged, thanks for reviewing!

PR #406, opened 2019-05-07
Allow sigma network to do Znode

nopslide:
In activeznode.cpp, there is following condition that does not allow znode to enable

```
if (infoMn.nProtocolVersion < MIN_ZNODE_PAYMENT_PROTO_VERSION_1
                || infoMn.nProtocolVersion > MIN_ZNODE_PAYMENT_PROTO_VERSION_2) 
```

so changing value in MIN_ZNODE_PAYMENT_PROTO_VERSION_2 to match with Protocol Verison

### DIFF
--- a/src/znode-payments.h
+++ b/src/znode-payments.h
@@ -24,7 +24,7 @@ static const int MNPAYMENTS_SIGNATURES_TOTAL            = 10;
 // V1 - Last protocol version before update
 // V2 - Newest protocol version
 static const int MIN_ZNODE_PAYMENT_PROTO_VERSION_1 = 90024;
-static const int MIN_ZNODE_PAYMENT_PROTO_VERSION_2 = 90026;
+static const int MIN_ZNODE_PAYMENT_PROTO_VERSION_2 = 90027;
 
 extern CCriticalSection cs_vecPayees;
 extern CCriticalSection cs_mapZnodeBlocks;


nopslide:
@psolstice need ur input

reubenyap:
Not needed according to @psolstice

nopslide:
I don't agree with this issue closing, 
If you take a look at https://github.com/zcoinofficial/zcoin/blob/sigma/src/version.h#L12
`static const int PROTOCOL_VERSION = 90027;`

along with following this line https://github.com/zcoinofficial/zcoin/blob/sigma/src/znode.cpp#L35

`nProtocolVersion(PROTOCOL_VERSION),`


so with the code in activeznode.cpp

`if (infoMn.nProtocolVersion < MIN_ZNODE_PAYMENT_PROTO_VERSION_1
                || infoMn.nProtocolVersion > MIN_ZNODE_PAYMENT_PROTO_VERSION_2) {
            nState = ACTIVE_ZNODE_NOT_CAPABLE;
            strNotCapableReason = "Invalid protocol version";
            LogPrintf("CActiveZnode::ManageStateRemote -- %s: %s\n", GetStateString(), strNotCapableReason);
            return;
        }`

if you take a look at https://github.com/zcoinofficial/zcoin/blob/sigma/src/znode-payments.h#L26
static const int MIN_ZNODE_PAYMENT_PROTO_VERSION_1 = 90024;
static const int MIN_ZNODE_PAYMENT_PROTO_VERSION_2 = 90026;

you will see (90027 < 90024) || (90027 > 90026), with this value, it will be TRUE, and it will show "Invalid protocol version"

nopslide:
This problem linked to this issue no. 
https://github.com/zcoinofficial/zcoin/issues/413

When u testing on testnet, this problem won't show, coz it is not reach the block height.

psolstice:
#446 does the same more effectively

PR #407, opened 2019-05-08
Fix sigma states corrupted when disconnect block

ultimaweapon:
This PR also convert indentation for `src/zerocoin_v3.cpp` from tabs to spaces too. So try to review each commit will be more easy to read.

### DIFF
--- a/src/zerocoin_v3.cpp
+++ b/src/zerocoin_v3.cpp
@@ -30,140 +30,140 @@ sigma::ParamsV3* ZCParamsV3 = sigma::ParamsV3::get_default();
 static CZerocoinStateV3 zerocoinStateV3;
 
 static bool CheckZerocoinSpendSerialV3(
-		CValidationState &state,
-		CZerocoinTxInfoV3 *zerocoinTxInfoV3,
-		const Scalar &serial,
-		int nHeight,
-		bool fConnectTip) {
-	// check for zerocoin transaction in this block as well
-	if (zerocoinTxInfoV3 &&
-			!zerocoinTxInfoV3->fInfoIsComplete &&
-			zerocoinTxInfoV3->spentSerials.find(serial) != zerocoinTxInfoV3->spentSerials.end())
-		return state.DoS(0, error("CTransaction::CheckTransaction() : two or more spends with same serial in the same block"));
+        CValidationState &state,
+        CZerocoinTxInfoV3 *zerocoinTxInfoV3,
+        const Scalar &serial,
+        int nHeight,
+        bool fConnectTip) {
+    // check for zerocoin transaction in this block as well
+    if (zerocoinTxInfoV3 &&
+            !zerocoinTxInfoV3->fInfoIsComplete &&
+            zerocoinTxInfoV3->spentSerials.find(serial) != zerocoinTxInfoV3->spentSerials.end())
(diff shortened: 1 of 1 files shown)


src/zerocoin_v3.cpp
```diff
@@ -648,10 +642,15 @@ void CZerocoinStateV3::RemoveBlock(CBlockIndex *index) {
 			mintedPubCoins.erase(coinIt);
 		}
 	}
+
+	index->mintedPubCoinsV3.clear();
```
martun: Is variable index used anywhere else, after we remove the block?

src/zerocoin_v3.cpp
```diff
@@ -471,12 +462,7 @@ bool ConnectBlockZCV3(
 				zerocoinStateV3.AddSpend(serial.first);
 			}
 		}
-        // Shows if V3 sigma mints are now allowed.
-        bool V3MintsAllowed = (pindexNew->nHeight >= Params().GetConsensus().nSigmaStartBlock);
```
martun: Can you describe me how this was causing the sync problems? I'm just curious to know.

ultimaweapon:
@psolstice you can click on each commit to review it one by one.

src/zerocoin_v3.cpp
ultimaweapon: Yes it not destroy and it will be reuse later so that why it make state corrupted.

psolstice:
Still I don't like the idea of tabs/spaces fixes mixed with the relevant changes

src/zerocoin_v3.cpp
martun: Aha, so it's not cleared, and it can be used only in the case where we RETURN to this branch of the blockchain right? In that case this index is re-used, and we need to clean it up. 

Can you add a long comment to the code describing it? Because if someone want to read this code in the future, that would help a lot. It was not obvious to me at all.

src/zerocoin_v3.cpp
psolstice: This is exactly the problem I warned about. If block gets disconnected and gets connected again and mints are not cleared they will be doubled

psolstice: Once the block header is downloaded CBlockIndex is retained and reused if referenced later. Even in case when the block is disconnected

ultimaweapon: Nope this one is not the problems. But rejecting sigma mint before activation block is not the right way as we discussed before so I remove it. We don't need to check here because `pblock->zerocoinTxInfoV3->mints` will always empty if sigma is not activated yet.

src/zerocoin_v3.cpp
ultimaweapon: Okay good idea. I will add comment here.

ultimaweapon:
@psolstice what is the problems with that? There is 2 choices if we want to go that way:

1. Discard the latest commit and force push to this PR. This will make it harder for other people that already review due to they cannot continue review just a new commits, they need to review the whole PR again. This also make PR history look weird too when someone take a look on this PR later.
2. Close this PR and open a new one. Other people that already review also need to review the whole PR again.

src/zerocoin_v3.cpp
psolstice: I'm pretty sure I was writing comments to different line of code and now it's here. Weird

PR #410, opened 2019-05-08
Getrawtransaction show denominations

a-bezrukov:
This change will allow users to see the Sigma spend denominations in Insight

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -92,8 +92,7 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
             } catch (CBadTxIn&) {
                 throw JSONRPCError(RPC_DATABASE_ERROR, "An error occurred during processing the Sigma spend information");
             }
-            in.push_back(Pair("anonymityGroup", int64_t(txin.prevout.n)));
-            in.push_back(Pair("groupId", int64_t(pubcoinId)));
+            in.push_back(Pair("anonymityGroup", int64_t(pubcoinId)));
             fillStdFields(in, txin);
 
             in.push_back(Pair("value", ValueFromAmount(spend->getIntDenomination())));
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -61,8 +61,6 @@ void ScriptPubKeyToJSON(const CScript& scriptPubKey, UniValue& out, bool fInclud
 
 namespace {
     void fillStdFields(UniValue & out, CTxIn const & txin) {
-        out.push_back(Pair("txid", txin.prevout.hash.GetHex()));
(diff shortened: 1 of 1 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -60,6 +59,17 @@ void ScriptPubKeyToJSON(const CScript& scriptPubKey, UniValue& out, bool fInclud
     out.push_back(Pair("addresses", a));
 }
 
+namespace {
+    void fillStdFields(UniValue & out, CTxIn const & txin) {
+        out.push_back(Pair("txid", txin.prevout.hash.GetHex()));
```
martun: I think for sigma spends prevout == null. This must crash. No?

a-bezrukov: It's filled with zerroes. It is not a pointer in CTxIn class...

src/rpc/rawtransaction.cpp
```diff
@@ -60,6 +59,17 @@ void ScriptPubKeyToJSON(const CScript& scriptPubKey, UniValue& out, bool fInclud
...
+        out.push_back(Pair("vout", (int64_t)txin.prevout.n));
```
ultimaweapon: My idea is we replace `txid` and `vout` fields to something like `anonymityGroup` will be more useful since `txid` will be always zeroes for Sigma spend. @a-bezrukov @martun what do you guys think?

a-bezrukov: I think it's a good idea. Please tell me which fields we need. Easy available there are CoinSpendV3 fields and groupId.

ultimaweapon: Anonymity group for Sigma is in `txin.prevout.n`.

a-bezrukov: Done

src/rpc/rawtransaction.cpp
```diff
@@ -88,16 +86,21 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
             std::unique_ptr<sigma::CoinSpendV3> spend;
+            uint32_t pubcoinId;
             try {
-                std::tie(spend, std::ignore) = ParseSigmaSpend(txin);
+                std::tie(spend, pubcoinId) = ParseSigmaSpend(txin);
             } catch (CBadTxIn&) {
                 throw JSONRPCError(RPC_DATABASE_ERROR, "An error occurred during processing the Sigma spend information");
             }
+            in.push_back(Pair("anonymityGroup", int64_t(txin.prevout.n)));
+            in.push_back(Pair("groupId", int64_t(pubcoinId)));
```
ultimaweapon: Actually `txin.prevout.n` and `pubcoinId` will be the same value. So my idea is we should keep only `pubcoinId` but rename it name to `anonymityGroup`.

PR #414, opened 2019-05-13
Add missing Sigma check

riordant:
Adds missing check for `TX_ZEROCOINMINTV3` in `ismine.cpp`.
Fixes warning:
```
script/ismine.cpp:61:13: warning: enumeration
      value 'TX_ZEROCOINMINTV3' not handled in
      switch [-Wswitch]
    switch (whichType)
            ^
script/ismine.cpp:61:13: note: add missing switch
      cases
    switch (whichType)
```


riordant:
Well as long as it worked for previous mints I think it's ok for now.

PR #421, opened 2019-05-21
Fix sigma page

thebevrishot:
- spend entries isn't cleared after spend success
- error message isn't show when create spend transaction with duplicated addresses.


src/qt/sigmapage.cpp
```diff
@@ -258,6 +258,11 @@ void SigmaPage::on_sendButton_clicked()
     // process sendStatus and on error generate message shown to user
     processSpendCoinsReturn(sendStatus);
 
+    if (sendStatus.status == WalletModel::OK)
```
ultimaweapon: Brace for this one should be the same line.

thebevrishot: fixed!

PR #429, opened 2019-05-29
Add sigma spend limit per transaction

thebevrishot:


src/test/sigma_mintspend_numinputs.cpp
```diff
@@ -34,15 +34,16 @@ BOOST_AUTO_TEST_CASE(sigma_mintspend_numinputs)
...
+    int denominationIndexB = 1; // 0.5
 
     sigma::CSigmaState *sigmaState = sigma::CSigmaState::GetState();
     auto& consensus = ::Params().GetConsensus();
 
-    // Create 2000 new empty blocks to get some funds. nMaxSigmaInputPerBlock == 500, so 
-    // we want to create 500 coins of each denomination. For denomination 100 we need 50.000 xzc.
-    CreateAndProcessEmptyBlocks(2000, scriptPubKey);
+    // Create 200 new empty blocks to get some funds. nMaxSigmaInputPerBlock == 50, so
+    CreateAndProcessEmptyBlocks(200, scriptPubKey);
```
martun: Looks like the comment said we need more funds than we get in 200 blocks.

PR #432, opened 2019-05-31
Small fix

martun:


reubenyap:
We're still calling the file zerocoin_v3.cpp?

martun:
@reubenyap We will rename it in a future commit. It must be done in a separate commit, because otherwise git does not understand that it's the same file, just moved.

PR #434, opened 2019-05-31
Make error message more clear

catchingknives:
Old message seemed to refer to Zcoin as "Zerocoin" and wasn't as clear. Note that this is only part of the log message.


reubenyap:
LGTM

PR #437, opened 2019-06-03
Disable znode persistent state by default

psolstice:


justanwar:
Leaving it at `true` for now (no change)

PR #439, opened 2019-06-04
Add coin control for mints & spends

riordant:
Adds the same coin control feature to the "Sigma" tab. When minting, users can choose coins as they would at the "Send" page. However when spending, users can now choose individual mints to spend.


thebevrishot:
@riordant please fix conflicts.

src/qt/sigmadialog.cpp
```diff
@@ -351,13 +409,180 @@ void SigmaPage::removeEntry(SendCoinsEntry* entry)
...
+        {
+            ui->labelCoinControlChangeLabel->setText(tr("Warning: Invalid Bitcoin address"));
+        }
+        else // Valid address
+        {
+            CKeyID keyid;
+            addr.GetKeyID(keyid);
+            if (!walletModel->havePrivKey(keyid)) // Unknown change address
+            {
+                ui->labelCoinControlChangeLabel->setText(tr("Warning: Unknown change address"));
```
thebevrishot: Seem like we not allow unknow change address right? because it not set `CoinControlDialog::coinControl->destChange = addr.Get();` I wonder this is warning not error.

src/qt/sigmadialog.cpp
```diff
@@ -286,19 +337,26 @@ void SigmaPage::on_sendButton_clicked()
...
+    }
+
     // now send the prepared transaction
     WalletModel::SendCoinsReturn sendStatus = walletModel->sendSigma(currentTransaction, selectedCoins, changes);
     // process sendStatus and on error generate message shown to user
     processSpendCoinsReturn(sendStatus);
 
     if (sendStatus.status == WalletModel::OK) {
         accept();
+        CoinControlDialog::coinControl->UnSelectAll();
```
thebevrishot: need to call this after mint too.

src/qt/walletmodel.cpp
```diff
@@ -727,8 +727,14 @@ void WalletModel::listCoins(std::map<QString, std::vector<COutput> >& mapCoins)
         }
 
         CTxDestination address;
-        if(!out.fSpendable || !ExtractDestination(cout.tx->vout[cout.i].scriptPubKey, address))
+        if(cout.tx->IsZerocoinMint() || cout.tx->IsSigmaMint()){
+            mapCoins[QString::fromStdString("(mint)")].push_back(out);
```
thebevrishot: Is it possible to separate `(sigma mint) ` and `(zerocoin mint)`? I think it should be better. How do you think?

src/qt/sigmadialog.cpp
riordant: Ok sure

src/qt/walletmodel.cpp
riordant: Do you mean have only Sigma mint? As they are already separate now

thebevrishot: ok, I forgot we disable zerocoin now, have only 'mint' is make sense.

src/qt/sigmadialog.cpp
thebevrishot: same behavior as bitcoin and not important issue. then I will close this.

riordant: Ok I see.

src/qt/walletmodel.cpp
```diff
@@ -699,7 +699,7 @@ bool WalletModel::isSpent(const COutPoint& outpoint) const
 void WalletModel::listCoins(std::map<QString, std::vector<COutput> >& mapCoins) const
```
ultimaweapon: Same as previous issue that we should not mix both UTXO and Mints.

src/qt/sigmadialog.cpp
```diff
@@ -351,13 +412,180 @@ void SigmaPage::removeEntry(SendCoinsEntry* entry)
...
+    }
+}
+
+// Coin Control: settings menu - coin control enabled/disabled by user
+void SigmaDialog::coinControlFeatureChanged(bool checked)
+{
+    ui->frameCoinControl->setVisible(checked);
+
+    if (!checked && walletModel) // coin control features disabled
+        CoinControlDialog::coinControl->SetNull();
```
ultimaweapon: Seem like we share coin control UI with send tab. That mean if someone open it in normal send tab it will also effect sigma tab and wise versa.

src/qt/sigmadialog.cpp
```diff
@@ -55,14 +58,49 @@ SigmaPage::SigmaPage(const PlatformStyle *platformStyle, QWidget *parent) :
 
     addEntry();
 
+    // init coin control section
+    GUIUtil::setupAddressWidget(ui->lineEditCoinControlChange, this);
+
+    connect(ui->addButton, SIGNAL(clicked()), this, SLOT(addEntry()));
+    connect(ui->clearButton, SIGNAL(clicked()), this, SLOT(clear()));
```
ultimaweapon: Those above two lines are duplicated with line 97 and 98.

src/qt/coincontroldialog.cpp
```diff
@@ -741,6 +742,10 @@ void CoinControlDialog::updateView()
             // address
             CTxDestination outputAddress;
             QString sAddress = "";
+            if(!treeMode && nIsMint){
```
ultimaweapon: We should not mix normal UTXO and Mints together by default. The available coins to show should depend on what type of transaction:

- Normal send: show only normal UTXO.
- Mint: show both normal UTXO and Mint but not able to mix it together.
- Spend: show only mints.

src/qt/sigmadialog.cpp
```diff
@@ -351,13 +412,180 @@ void SigmaPage::removeEntry(SendCoinsEntry* entry)
...
+    CoinControlDialog dlg(platformStyle);
+    dlg.setModel(walletModel);
+    dlg.exec();
+    coinControlUpdateLabels();
+}
+
+// Coin Control: checkbox custom change address
+void SigmaDialog::coinControlChangeChecked(int state)
+{
+    if (state == Qt::Unchecked)
```
ultimaweapon: We should allow custom change only when it is mint.

src/qt/sigmadialog.h
```diff
@@ -31,14 +32,28 @@ public Q_SLOTS:
...
+    void coinControlChangeEdited(const QString &);
 
 private:
-    Ui::SigmaPage *ui;
+    Ui::SigmaDialog *ui;
     ClientModel *clientModel;
     WalletModel *walletModel;
     bool isNewRecipientAllowed;
     const PlatformStyle *platformStyle;
+    QDialog *sigmaDialog;
```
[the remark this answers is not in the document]
riordant: Agreed

src/qt/coincontroldialog.cpp
riordant: Wouldn't it be better to only show normal UTXO for Mint?

src/wallet/wallet.cpp
```diff
@@ -69,6 +69,8 @@ CFeeRate CWallet::fallbackFee = CFeeRate(DEFAULT_FALLBACK_FEE);
 
 const uint256 CMerkleTx::ABANDON_HASH(uint256S("0000000000000000000000000000000000000000000000000000000000000001"));
 
+CCoinControl g_coincontrol;
```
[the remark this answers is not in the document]
riordant: What is the problem with it as global variable? It saves having to pass the variable all the way through to `GetAvailableCoins`. But if you think that's better I can change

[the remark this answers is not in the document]
ultimaweapon: The problems with this as a global variable is it need to be reset when we don't need coin control feature before calling the function, which will make the function hard to use because the caller need to know there is a global variable that control function behavior, which is unexpected by other people if they don't know how that function was implemented.

src/qt/coincontroldialog.cpp
ultimaweapon: That is what I think before but after discuss with @thebevrishot users might want to merge their smaller denomination to the bigger one (e.g. input evevel `0.1` to create `1`).

reubenyap: When minting why do you want to show mint for selection? Mint for mint?

ultimaweapon: @reubenyap please see my above comment.

reubenyap: I think this is a rare case but sure. Go ahead if the fix is easy.

src/zerocoin_v3.cpp
```diff
@@ -566,6 +566,57 @@ bool ConnectBlockSigma(
     return true;
 }
 
+bool ZerocoinGetOutPoint(COutPoint& outPoint, GroupElement pubCoinValue) {
```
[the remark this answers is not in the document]
riordant: Agreed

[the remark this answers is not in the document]
riordant: Agreed

src/qt/sigmadialog.cpp
riordant: You mean when spending a mint right?

src/wallet/txbuilder.h
```diff
@@ -36,10 +36,10 @@ class TxBuilder
     explicit TxBuilder(CWallet& wallet) noexcept;
     virtual ~TxBuilder();
 
-    CWalletTx Build(const std::vector<CRecipient>& recipients, CAmount& fee);
+    CWalletTx Build(const std::vector<CRecipient>& recipients, CAmount& fee, const CCoinControl *coinControl = NULL);
```
ultimaweapon: This will cause more dependencies at the abstraction level. Also cause more requirements to the derived class. We should pass `CCoinControl` to constructor of `SigmaSpendBuilder` instead.

src/qt/sigmadialog.cpp
ultimaweapon: Nope. When we spend it will never changes as a normal output, it will always changes as mints. That mean if we allow user to specify custom change address for spend they might confuse.

src/wallet/wallet.cpp
```diff
@@ -2619,11 +2621,14 @@ void CWallet::AvailableCoins(vector <COutput> &vCoins, bool fOnlyConfirmed, cons
...
 
             for (unsigned int i = 0; i < pcoin->vout.size(); i++) {
-                if ((pcoin->vout[i].scriptPubKey.IsZerocoinMint() || pcoin->vout[i].scriptPubKey.IsSigmaMint()) && !fConsiderMints){
-                    continue;
-                }
                 bool found = false;
-                if (nCoinType == ONLY_DENOMINATED) {
+                if(nCoinType == ALL_COINS){
+                    // We are now taking ALL_COINS to mean everything sans mints
+                    found = !(pcoin->vout[i].scriptPubKey.IsZerocoinMint() || pcoin->vout[i].scriptPubKey.IsSigmaMint());
```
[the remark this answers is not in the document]
riordant: It's functionally correct at least. The nice thing about this is we don't have to change other logic that uses `ALL_COINS` (Bitcoin Core stuff actually). So instead I introduce two new `AvailableCoinsType` values, `ONLY_MINTS` and `WITH_MINTS`. 
If `WITH_MINTS` passed, normal UTXOs + mints are considered. If `ONLY_MINTS`, we only return mints. And `ALL_COINS` is the same as before, just returns normal UTXOs.

src/zerocoin_v3.cpp
```diff
@@ -566,22 +566,12 @@ bool ConnectBlockSigma(
...
-    BOOST_FOREACH(sigma::CoinDenomination denomination, denominations){
-        sigma::PublicCoin pubCoin(pubCoinValue, denomination);
-        auto mintedCoinHeightAndId = sigmaState->GetMintedCoinHeightAndId(pubCoin);
-        mintHeight = mintedCoinHeightAndId.first;
-        coinId = mintedCoinHeightAndId.second;
-        if(mintHeight!=-1 && coinId!=-1)
-            break;
-    }
-
+    auto mintedCoinHeightAndId = sigmaState->GetMintedCoinHeightAndId(pubCoin);
```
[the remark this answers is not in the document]
riordant: Because we were passing only `GroupElement`, now full `pubCoin` (good suggestion from @ultimaweapon )

src/qt/sigmadialog.cpp
```diff
@@ -183,17 +183,17 @@ void SigmaDialog::on_mintButton_clicked()
...
             QMessageBox::Ok, QMessageBox::Ok);
         return;
     }
 
     if (amount % smallestDenominationValue != 0) {
         amount -= amount % smallestDenominationValue;
         auto reply = QMessageBox::question(
-            this, tr("Amount to mint is impossible."),
-            tr("Amount to mint must be multiple of 0.05 XZC. Do you want to spend %1 XZC?"
+            this, tr("Amount to mint is too low."),
```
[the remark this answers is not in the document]
riordant: Ok. "impossible" does not sound grammatically correct though. @reubenyap ?

src/wallet/txbuilder.h
riordant: Sounds good.

src/wallet/wallet.cpp
```diff
@@ -2290,7 +2292,7 @@ bool CWallet::GetCoinsToSpend(
             _("Required amount exceed value spend limit"));
     }
 
-    std::list<CSigmaEntry> coins = GetAvailableCoins();
+    std::list<CSigmaEntry> coins = GetAvailableCoins(coinControl);
```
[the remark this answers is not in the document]
riordant: But not all coins can be used depending on requirements, my understanding was that the user selected coins that could be considered?

src/qt/sigmadialog.h
```diff
@@ -24,6 +25,8 @@ class SigmaDialog : public QDialog
     void setClientModel(ClientModel *model);
     void setWalletModel(WalletModel *model);
 
+    static bool mintTabSelected;
```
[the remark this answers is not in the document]
ultimaweapon: Okay. The idea from me is add a new property to `SigmaCoinControlDialog` to indicate what is it current mode (e.g. normal UTXO or mints). Then from `SigmaDialog` just inspect if it is the right mode for the current tab and change it if it not the right one.

src/qt/sigmadialog.cpp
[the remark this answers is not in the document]
riordant: @Panu `Unable to Mint`?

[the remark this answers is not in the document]
thebevrishot: yep, that's make sense.

src/qt/sigmadialog.h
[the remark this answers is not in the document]
riordant: Ok, changed to reflect this

src/wallet/wallet.cpp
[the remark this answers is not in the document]
thebevrishot: all coins that be shown on coin control, should be usable. and I think user should expect the behavior like normal coin control which use all coins that user select.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
thebevrishot: ok, make sense.

src/zerocoin_v3.cpp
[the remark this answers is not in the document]
thebevrishot: I see

src/qt/sigmadialog.cpp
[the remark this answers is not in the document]
riordant: Done

src/wallet/wallet.cpp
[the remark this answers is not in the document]
riordant: Sorry I don't understand. All coins are shown to user at Coin Control screen. Now assume the user wants to spend 30 XZC, and selects two spends of 100 XZC each. Obviously only one can be used? All coins will be used if necessary

src/wallet/txbuilder.h
```diff
@@ -31,12 +31,13 @@ class TxBuilder
...
 
 public:
     explicit TxBuilder(CWallet& wallet) noexcept;
     virtual ~TxBuilder();
 
-    CWalletTx Build(const std::vector<CRecipient>& recipients, CAmount& fee, const CCoinControl *coinControl = NULL);
+    CWalletTx Build(const std::vector<CRecipient>& recipients, CAmount& fee);
 
 protected:
     virtual CAmount GetInputs(std::vector<std::unique_ptr<InputSigner>>& signers, CAmount required, const CCoinControl *coinControl = NULL) = 0;
```
ultimaweapon: We don't need `coinControl` for this method anymore.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
thebevrishot: coin control in bitcoin(also for our normal coins) will use all coins that user select for example if user choose 2 coins of 100 XZC. and send just 0.1 XZC. bitcoin still use 2 coins of 100 XZC to spend. I think user will expect same thing with coin control for sigma.

src/wallet/txbuilder.h
riordant: Yes, it has been removed

src/wallet/wallet.cpp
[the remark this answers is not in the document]
riordant: Ok, looking at fix for this

src/wallet/txbuilder.h
ultimaweapon: I mean the one in `GetInputs`, not `Build`.

riordant: Agree. Fixed

src/wallet/wallet.cpp
```diff
@@ -2191,11 +2191,13 @@ std::list<CSigmaEntry> CWallet::GetAvailableCoins(const CCoinControl *coinContro
     std::list<CSigmaEntry> coins;
     CWalletDB(strWalletFile).ListSigmaPubCoin(coins);
 
+    std::set<COutPoint> lockedCoins = setLockedCoins;
```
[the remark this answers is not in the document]
riordant: When trying to capture `setLockedCoins` in `remove_if`, it returns `error: capture of non-variable`. Similarly capturing `this` in `remove_if` caused undefined behaviour. So the only way I found that worked was to copy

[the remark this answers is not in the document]
ultimaweapon: Do you have the source that it is undefined behavior? I can't remember it is undefined behavior.

src/zerocoin_v3.cpp
```diff
@@ -595,21 +595,7 @@ bool ConnectBlockSigma(
...
-    if(mintHeight==-1 && coinId==-1)
-        return false;
-
-    // get block containing mint
-    CBlockIndex *mintBlock = chainActive[mintHeight];
-    CBlock block;
-    if(!ReadBlockFromDisk(block, mintBlock, ::Params().GetConsensus()))
-        LogPrintf("can't read block from disk.\n");
-
+bool GetOutPointFromBlock(COutPoint& outPoint, GroupElement pubCoinValue, CBlock block){
```
ultimaweapon: `CBlock` is too large to pass by value. This one must be passed by reference, so please change it to `const CBlock&`. Also will be better to fix this for `pubCoinValue` too.

src/zerocoin_v3.cpp
```diff
@@ -636,6 +622,62 @@ bool GetOutPoint(COutPoint& outPoint, sigma::PublicCoin pubCoin) {
     return false;
 }
 
+bool GetOutPoint(COutPoint& outPoint, sigma::PublicCoin pubCoin) {
```
ultimaweapon: Same as previous issue, should pass by reference.

src/zerocoin_v3.cpp
```diff
@@ -636,6 +622,62 @@ bool GetOutPoint(COutPoint& outPoint, sigma::PublicCoin pubCoin) {
...
+    // get block containing mint
+    CBlockIndex *mintBlock = chainActive[mintHeight];
+    CBlock block;
+    if(!ReadBlockFromDisk(block, mintBlock, ::Params().GetConsensus()))
+        LogPrintf("can't read block from disk.\n");
+
+    return GetOutPointFromBlock(outPoint, pubCoinValue, block);
+}
+
+bool GetOutPoint(COutPoint& outPoint, uint256 pubCoinValueHash) {
```
ultimaweapon: Same here.

src/zerocoin_v3.cpp
```diff
@@ -636,6 +622,62 @@ bool GetOutPoint(COutPoint& outPoint, sigma::PublicCoin pubCoin) {
...
+    // get block containing mint
+    CBlockIndex *mintBlock = chainActive[mintHeight];
+    CBlock block;
+    if(!ReadBlockFromDisk(block, mintBlock, ::Params().GetConsensus()))
+        LogPrintf("can't read block from disk.\n");
+
+    return GetOutPointFromBlock(outPoint, pubCoin.value, block);
+}
+
+bool GetOutPoint(COutPoint& outPoint, GroupElement pubCoinValue) {
```
ultimaweapon: Same here.

src/hdmint/tracker.cpp
```diff
@@ -372,8 +372,9 @@ bool CHDMintTracker::UpdateStatusInternal(const std::set<uint256>& setMempool, C
     uint256 hashPubcoin = sigma::GetPubCoinValueHash(mint.pubCoinValue);
     //! Check whether this mint has been spent and is considered 'pending' or 'confirmed'
     // If there is not a record of the block height, then look it up and assign it
-    uint256 txidMint;
-    bool isMintInChain = ZerocoinGetSigmaMintTxHash(txidMint, mint.pubCoinValue);
+    COutPoint outPoint;
+    bool isMintInChain = GetOutPoint(outPoint, mint.pubCoinValue);
```
ultimaweapon: We already know the denomination for this so call the pubcoin version will be better to remove trying every denominations.

src/hdmint/tracker.cpp
```diff
@@ -372,8 +372,9 @@ bool CHDMintTracker::UpdateStatusInternal(const std::set<uint256>& setMempool, C
...
+    uint256 txidMint = outPoint.hash;
```
ultimaweapon: Can we change it to reference instead?

src/hdmint/wallet.cpp
```diff
@@ -169,8 +169,10 @@ void CHDMintWallet::SyncWithChain(bool fGenerateMintPool)
                 continue;
             }
 
-            uint256 txHash;
-            if (sigma::ZerocoinGetSigmaMintTxHash(txHash, pubCoinValueHash)) {
+            
+            COutPoint outPoint;
+            if (sigma::GetOutPoint(outPoint, pubCoinValueHash)) {
+                uint256 txHash = outPoint.hash;
```
ultimaweapon: Same here.

riordant:
@ultimaweapon Agree with all the above. Testing changes now

PR #442, opened 2019-06-05
Detect late upgrade

a-bezrukov:
This requests reindexing to restore Sigma state if it is detected that client upgrade happened after the Sigma HF.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1636,15 +1636,15 @@ bool AppInit2(boost::thread_group &threadGroup, CScheduler &scheduler) {
 
                 pblocktree = new CBlockTreeDB(nBlockTreeDBCache, false, fReindex);
 
-	            if (!fReindex) {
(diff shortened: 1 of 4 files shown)


src/init.cpp
```diff
@@ -1662,6 +1662,18 @@ bool AppInit2(boost::thread_group &threadGroup, CScheduler &scheduler) {
                     break;
                 }
 
+                if (!fReindex) {
+                    CBlockIndex *tip = chainActive.Tip();
+                    if (tip && tip->nHeight >= chainparams.GetConsensus().nSigmaStartBlock) {
+                        const uint256* phash = tip->phashBlock;
+                        if (pblocktree->GetBlockIndexVersion(*phash) < CLIENT_VERSION) {
```
psolstice: We need to compare it with the first version having sigma (0.13.8+ or whatever it would be), not the current version

PR #444, opened 2019-06-05
Sigma start block adjusted

a-bezrukov:


thebevrishot:
I try to calculate with 300 sec per block, the block 175600 is about  June 30. the block on July 11, 8:00 am should be 178588.

a-bezrukov:
I calculated based on block frequency during the last month.

thebevrishot:
Could you please provide more concrete logic?

PR #448, opened 2019-06-06
Fix mining invalid spends, also don't check spend proof while adding to mempool

martun:


reubenyap:
On hold for this as may not be needed.

PR #450, opened 2019-06-10
Fix mark spendable any coin which have at least 6 confirmations.

thebevrishot:
current behavior
- mark a coin as spendable only if have 6 confirms which is not a single coin in the group.

expected behavior
- mark any coin as spendable only if have 6 confirms
- If coins in wallet not enough to spend will should error not enough balance and then show all coins that mark as spendable but can't not spend because of privacy issue. 

for more information about notification can take a look more in the card https://trello.com/c/M403vjcG/472-fix-sigma-tab-show-wrong-balance
the situation is have only 2 coins in global state(1 and 0.5) and try to spend 1.3 XZC. Message error will tell user what denomination(s) need to wait more other user to mint.

### DIFF
--- a/src/hdmint/tracker.cpp
+++ b/src/hdmint/tracker.cpp
@@ -444,7 +444,7 @@ bool CHDMintTracker::UpdateStatusInternal(const std::set<uint256>& setMempool, C
 
         // Check that the mint has correct used status
         if (mint.isUsed != isUsed) {
-            LogPrintf("%s : Set mint %s isUsed to %d\n", __func__, hashPubcoin.GetHex(), isUsed);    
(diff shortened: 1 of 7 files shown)


thebevrishot:
don't merge now, need to wait https://github.com/zcoinofficial/zcoin/pull/439 merge first.

levonpetrosyan93:
Please fix conflicts.

thebevrishot:
@a-bezrukov @levonpetrosyan93 Merge sigma to this is too complex. Then I just apply same logic as before and fix some function. please review.

levonpetrosyan93:
LGTM

PR #451, opened 2019-06-10
Fix coingroup id test

thebevrishot:
this test use only denomination 0.05.
after add 15001st mint which is in different block that contain 15000th. group id should increase.
and then remove block that contain 15001st block id should decrease back.

### DIFF
--- a/src/test/sigma_state_tests.cpp
+++ b/src/test/sigma_state_tests.cpp
@@ -214,9 +214,6 @@ BOOST_AUTO_TEST_CASE(sigma_addmint_more_than_restriction_in_one)
     sigmaState->Reset();
 }
 
-// This is a correct case, but should be commented till fix of code
-// Cause code of creating another coin group is not in the state class itself
-
 // Checking AddMint ZC_SPEND_V3_COINSPERID+1 coins on different blocks should have two group id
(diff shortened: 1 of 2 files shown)


src/test/sigma_state_tests.cpp
```diff
@@ -214,22 +214,23 @@ BOOST_AUTO_TEST_CASE(sigma_addmint_more_than_restriction_in_one)
...
+    sigma::CoinDenomination testDenomination = sigma::CoinDenomination::SIGMA_DENOM_0_05;
+
+    std::vector<CBlockIndex> indexs(ZC_SPEND_V3_COINSPERID + 1);
     for (int i = 0; i <= ZC_SPEND_V3_COINSPERID; ++i){
-        index = CreateBlockIndex(i+1);
+        indexs[i] = CreateBlockIndex(i+1);
         auto params = sigma::Params::get_default();
-        const sigma::PrivateCoin privcoin(params);
+        const sigma::PrivateCoin privcoin(params, testDenomination);
         pubcoin = privcoin.getPublicCoin();
```
ultimaweapon: We should fix this too. `getPublicCoin()` return a const reference to `PublicCoin` so we should take this advantage for performance due to this loop iterate a lot.

thebevrishot:
@martun done, please take a look.

PR #457, opened 2019-06-14
Add coin per group limit.

thebevrishot:
push mints on new block to next group if current_coins + new_coins > 16000

### DIFF
--- a/src/zerocoin_v3.cpp
+++ b/src/zerocoin_v3.cpp
@@ -616,8 +616,7 @@ void CSigmaState::AddMintsToStateAndBlockIndex(
 
         SigmaCoinGroupInfo &coinGroup = coinGroups[make_pair(denomination, mintCoinGroupId)];
 
-        if (coinGroup.nCoins < ZC_SPEND_V3_COINSPERID // there's still space in the accumulator
-            && coinGroup.nCoins + mintsWithThisDenom.size() <= ZC_SPEND_V3_COINSPERID_LIMIT) {
+        if (coinGroup.nCoins + mintsWithThisDenom.size() <= ZC_SPEND_V3_COINSPERID_LIMIT) {
             if (coinGroup.nCoins == 0) {
                 // first group of coins for given denomination
                 assert(coinGroup.firstBlock == nullptr);
--- a/src/zerocoin_v3.cpp
+++ b/src/zerocoin_v3.cpp
@@ -600,29 +600,26 @@ bool CSigmaState::AddMintsToStateAndBlockIndex(
         CBlockIndex *index,
         const CBlock* pblock) {
 
-    std::unordered_map<sigma::CoinDenomination, size_t> blockDenomCounts;
+    std::unordered_map<sigma::CoinDenomination, std::vector<sigma::PublicCoin>> blockDenomMints;
     for (const auto& mint : pblock->sigmaTxInfo->mints) {
-        blockDenomCounts[mint.getDenomination()]++;
+        blockDenomMints[mint.getDenomination()].push_back(mint);
     }
 
-    for (const auto& pubCoin : pblock->sigmaTxInfo->mints) {
-        sigma::CoinDenomination denomination = pubCoin.getDenomination();
+    for (const auto& it : blockDenomMints) {
+        const sigma::CoinDenomination denomination = it.first;
+        const std::vector<sigma::PublicCoin>& mintsWithThisDenom = it.second;
 
         if (latestCoinIds[denomination] < 1)
             latestCoinIds[denomination] = 1;
-        int	mintCoinGroupId = latestCoinIds[denomination];
+        auto mintCoinGroupId = latestCoinIds[denomination];
 
-        // ZC_SPEND_V3_COINSPERID = 15.000, yet the actual limit of coins per accumlator is 16.000.
-        // We need to cut at 15.000, such that we always have enough space for new mints. Mints for
-        // each block will end up in the same accumulator.
+
         SigmaCoinGroupInfo &coinGroup = coinGroups[make_pair(denomination, mintCoinGroupId)];
         int coinsPerId = ZC_SPEND_V3_COINSPERID;
         int coinsPerIdLimit = ZC_SPEND_V3_COINSPERID_LIMIT;
-        if ((coinGroup.nCoins < coinsPerId // there's still space in the accumulator
-            && coinGroup.nCoins + blockDenomCounts[denomination] < coinsPerIdLimit)
-            || coinGroup.lastBlock == index // or we have already placed some coins from current block.
-            ) {
-            if (coinGroup.nCoins++ == 0) {
+        if (coinGroup.nCoins < coinsPerId // there's still space in the accumulator
+            && coinGroup.nCoins + mintsWithThisDenom.size() <= coinsPerIdLimit) {
+            if (coinGroup.nCoins == 0) {
                 // first group of coins for given denomination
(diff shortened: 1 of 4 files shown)


src/zerocoin_v3.cpp
```diff
@@ -604,51 +596,64 @@ void CSigmaTxInfo::Complete() {
...
+
+        // ZC_SPEND_V3_COINSPERID = 15.000, yet the actual limit of coins per accumlator is 16.000.
+        // We need to cut at 15.000, such that we always have enough space for new mints. Mints for
+        // each block will end up in the same accumulator.
+        SigmaCoinGroupInfo &coinGroup = coinGroups[make_pair(denomination, mintCoinGroupId)];
+        int coinsPerId = ZC_SPEND_V3_COINSPERID;
+        int coinsPerIdLimit = ZC_SPEND_V3_COINSPERID_LIMIT;
+        if ((coinGroup.nCoins < coinsPerId // there's still space in the accumulator
+            && coinGroup.nCoins + blockDenomCounts[denomination] < coinsPerIdLimit)
+            || coinGroup.lastBlock == index // or we have already placed some coins from current block.
```
ultimaweapon: Seem like this one still allow mint per group to exceed limit.

src/zerocoin_v3.cpp
```diff
@@ -604,51 +596,64 @@ void CSigmaTxInfo::Complete() {
...
-        newCoinGroup.firstBlock = newCoinGroup.lastBlock = index;
-        newCoinGroup.nCoins = 1;
-    }
-    CMintedCoinInfo coinInfo;
-    coinInfo.denomination = denomination;
-    coinInfo.id = mintCoinGroupId;
-    coinInfo.nHeight = index->nHeight;
-    mintedPubCoins.insert(std::make_pair(pubCoin, coinInfo));
-    return mintCoinGroupId;
+    return 1;
```
martun: return true?

src/zerocoin_v3.cpp
```diff
@@ -604,51 +596,64 @@ void CSigmaTxInfo::Complete() {
...
+            latestCoinIds[denomination] = 1;
+        int	mintCoinGroupId = latestCoinIds[denomination];
+
+        // ZC_SPEND_V3_COINSPERID = 15.000, yet the actual limit of coins per accumlator is 16.000.
+        // We need to cut at 15.000, such that we always have enough space for new mints. Mints for
+        // each block will end up in the same accumulator.
+        SigmaCoinGroupInfo &coinGroup = coinGroups[make_pair(denomination, mintCoinGroupId)];
+        int coinsPerId = ZC_SPEND_V3_COINSPERID;
+        int coinsPerIdLimit = ZC_SPEND_V3_COINSPERID_LIMIT;
+        if ((coinGroup.nCoins < coinsPerId // there's still space in the accumulator
```
martun: Why is the first part of "if" necessary? Check in "oinGroup.nCoins + blockDenomCounts[denomination] < coinsPerIdLimit" checks that all the mints in this block will fit in the same accumulator right? And I believe there must be coinsPerId used, not coinsPerIdLimit.

src/zerocoin_v3.cpp
```diff
@@ -604,51 +596,64 @@ void CSigmaTxInfo::Complete() {
...
+            && coinGroup.nCoins + blockDenomCounts[denomination] < coinsPerIdLimit)
```
levonpetrosyan93: I think better to do this check for each denomination, not for each coin.

src/zerocoin_v3.cpp
levonpetrosyan93: Agree with @ultimaweapon.

src/zerocoin_v3.cpp
thebevrishot: you're right. thanks.

src/zerocoin_v3.cpp
thebevrishot: good idea. thanks.

src/zerocoin_v3.cpp
thebevrishot: I use 15 000 as soft cap(`coinsPerID`) and 16 000(`coinsPerIdLimit`) as hard cap. 

If before add new mints coin in group is greater than or equal 15000, then new mints should be added to next group. which is first case. `coinGroup.nCoins < coinsPerId`

and if mints in group is less than 15000 but new mints + old mints > 16000, then new mints should also be push to new group. which is second case `coinGroup.nCoins + new mints amount < coinsPerIdLimit`

src/zerocoin_v3.cpp
thebevrishot: quite not understand how it allow but have a bug here I have fixed. please check.

src/zerocoin_v3.cpp
```diff
@@ -600,29 +600,26 @@ bool CSigmaState::AddMintsToStateAndBlockIndex(
...
         SigmaCoinGroupInfo &coinGroup = coinGroups[make_pair(denomination, mintCoinGroupId)];
         int coinsPerId = ZC_SPEND_V3_COINSPERID;
         int coinsPerIdLimit = ZC_SPEND_V3_COINSPERID_LIMIT;
-        if ((coinGroup.nCoins < coinsPerId // there's still space in the accumulator
-            && coinGroup.nCoins + blockDenomCounts[denomination] < coinsPerIdLimit)
-            || coinGroup.lastBlock == index // or we have already placed some coins from current block.
-            ) {
-            if (coinGroup.nCoins++ == 0) {
+        if (coinGroup.nCoins < coinsPerId // there's still space in the accumulator
+            && coinGroup.nCoins + mintsWithThisDenom.size() <= coinsPerIdLimit) {
```
ultimaweapon: Seem like this one will cause signesss comparison mismatch.

src/zerocoin_v3.cpp
```diff
@@ -637,23 +634,27 @@ bool CSigmaState::AddMintsToStateAndBlockIndex(
...
+            coinInfo.id = mintCoinGroupId;
+            coinInfo.nHeight = index->nHeight;
+            mintedPubCoins.insert(std::make_pair(mint, coinInfo));
+
+            LogPrintf("AddMintsToStateAndBlockIndex: mint added denomination=%d, id=%d\n", denomination, mintCoinGroupId);
+            index->sigmaMintedPubCoins[{denomination, mintCoinGroupId}].push_back(mint);
+        }
     }
-    return 1;
+    return true;
```
ultimaweapon: Seem like we don't need a returned value anymore. So change this method to `void` instead will be better.

thebevrishot:
@ultimaweapon some thing like `coinGroup.lastBlock->nHeight <= index->nHeight` ?

src/zerocoin_v3.cpp
thebevrishot: make sense.

src/zerocoin_v3.cpp
thebevrishot: fixed

src/zerocoin_v3.cpp
martun: Wait, I did not understand why we need 2 caps any more. We must do the following
1. Know how many coins are already in the last accumulator.
2. Count how many new mints are in the new block.
3. If these mints + number of coins already in the accumulator < 16.000, then put them all into the same accumulator, if not, then close the current accumulator and put them all into the next accumulator. 

This must be the whole logic, no soft cap.

thebevrishot: I see, that's make sense.

PR #458, opened 2019-06-14
Removing all sigma spends referencing the current block hash when the…

martun:
… block is removed.

### DIFF
--- a/src/zerocoin_v3.cpp
+++ b/src/zerocoin_v3.cpp
@@ -23,8 +23,6 @@
 
 #include <ios>
 
-extern CTxMemPool mempool;
-
 namespace sigma {
 
 // Set up the Sigma Params object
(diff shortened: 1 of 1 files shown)


src/zerocoin_v3.cpp
```diff
@@ -22,6 +23,8 @@
 
 #include <ios>
 
+extern CTxMemPool mempool;
+
```
ultimaweapon: I think we don't need this due to we already included `txmempool.h`

src/zerocoin_v3.cpp
```diff
@@ -460,8 +463,41 @@ bool CheckSigmaTransaction(
     return true;
 }
 
-void DisconnectTipSigma(CBlock & /*block*/, CBlockIndex *pindexDelete) {
+void RemoveSigmaSpendsReferencingBlock(CBlockIndex* blockIndex) {
+    std:vector<CTransaction> txn_to_remove;
+    for (CTxMemPool::txiter mi = mempool.mapTx.begin(); mi != mempool.mapTx.end(); ++mi) {
```
ultimaweapon: We need to lock before accessing `mempool`

levonpetrosyan93: We are locking upper, for example in DisconnectBlocks() . Why do we need to lock here? May be better to call AssertLockHeld() here?

psolstice: theoretically it's ok to have it run only under `cs_main` lock which is acquired before the call. But adding one more lock won't hurt

psolstice: also have a minor question to @martun : what about dandelion pool? don't we need to delete txs from there as well?

martun: Having them left in stempool does not hurt, but I will change the code to remove also from there. Will also add additional locks, they will not hurt.

PR #460, opened 2019-06-16
Reject Sigma Double Mint

levonpetrosyan93:


src/zerocoin_v3.cpp
```diff
@@ -355,6 +355,10 @@ bool CheckSigmaMintTransaction(
     if (hasCoin) {
        LogPrintf("CheckSigmaMintTransaction: double mint, tx=%s\n",
                 txout.GetHash().ToString());
+        return state.DoS(100,
```
martun: Are we sure we want function CheckSigmaMintTransaction to fail, in case it's called twice?

levonpetrosyan93: Yes I see this function can be called twice, it can be at miner. let me add a condition before return.

ultimaweapon:
Sometime ago I found it called twice but I cannot find a second reference anymore. Not sure which way we should go.

psolstice:
I think there is a need for `fStatefulCheck` flag here. So the double mint is checked only if it is set to true

PR #462, opened 2019-06-17
Preventing surges in sigma

a-bezrukov:


a-bezrukov:
Ready to review. The test failure caused by the HD mint test failure.

src/hash_functions.cpp
```diff
@@ -31,4 +31,21 @@ std::size_t CPublicCoinHash::operator ()(const sigma::PublicCoin& coin) const no
     return result;
 }
 
+
+CMintedCoinInfo CMintedCoinInfo::make(CoinDenomination denomination,  int coinGroupId, int nHeight) {
```
riordant: Probably best to change filename as these are the only objects present and are unrelated to hash functions

src/zerocoin_v3.cpp
```diff
@@ -584,8 +591,8 @@ bool ConnectBlockSigma(
             }
 
             if (!fJustCheck) {
-                pindexNew->sigmaSpentSerials.insert(serial.first);
-                sigmaState.AddSpend(serial.first);
+                pindexNew->sigmaSpentSerials.insert(serial);
+                sigmaState.AddSpend(serial.first, serial.second.denomination, serial.second.coinGroupId);
```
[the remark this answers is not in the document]
a-bezrukov: AddSpend now used in 12 places, 10 of them are tests. If we implemented the change, most of the code would look like this:
`sigmaState->AddSpend(coinSerial, CSpendCoinInfo::make(denom, groupId));` or like this
`CSpendCoinInfo coinInfo = CSpendCoinInfo::make(denom, groupId);
sigmaState->AddSpend(coinSerial, coinInfo);`

As it seems to me, both would make readability worse. On the other hand, a caller of this code should not know about CSpendCoinInfo, as the CSpendCoinInfo is used for internal storage, and it may introduce unnecessary include directive which will make it slower to compile.

Let's please leave it as is to not introduce code coupling.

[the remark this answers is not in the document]
riordant: Ok, but in the case of `AddMint` it is expected that the caller should know?

src/coin_containers.h
```diff
@@ -0,0 +1,59 @@
+#ifndef HASH_FUNCTIONS_H__
+#define HASH_FUNCTIONS_H__
+
```
psolstice: When renaming a file you need to change the constants at the top

src/zerocoin_v3.cpp
```diff
@@ -732,11 +739,90 @@ void CSigmaTxInfo::Complete() {
...
+/******************************************************************************/
+
+CSigmaState::Containers::Containers(std::atomic<bool> & surgeCondition)
+: surgeCondition(surgeCondition)
+{}
+
+void CSigmaState::Containers::AddMint(sigma::PublicCoin const & pubCoin, CMintedCoinInfo const & coinInfo) {
+    mintedPubCoins.insert(std::make_pair(pubCoin, coinInfo));
+    mintMetaInfo[coinInfo.coinGroupId][coinInfo.denomination] += 1;
+    CheckSurgeCondition(coinInfo.coinGroupId, coinInfo.denomination);
```
psolstice: How adding a mint can create a surge condition? Same for removing a spend

src/zerocoin_v3.cpp
```diff
@@ -732,11 +739,90 @@ void CSigmaTxInfo::Complete() {
...
+    if( result ) {
+        std::ostringstream ostr;
+        ostr << "Turning sigma surge protection ON: groupId: " << groupId << ", denomination: " << denom << '\n';
+        error(ostr.str().c_str());
+    }
+
+    for(metainfo_container_t::const_iterator smi = spendMetaInfo.begin(); smi != spendMetaInfo.end() && !result; ++smi) {
+        for(std::map<CoinDenomination, size_t>::const_iterator di = smi->second.begin(); di != smi->second.end() && !result; ++di) {
+            if(di->second > mintMetaInfo[smi->first][di->first]) {
+                result = true;
```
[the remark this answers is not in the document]
psolstice: why are we walking over all the spend meta infos when instructed to check for only one particular groupId/denom?

src/zerocoin_v3.cpp
```diff
@@ -186,7 +187,7 @@ bool CheckSigmaSpendTransaction(
     for (const CTxIn &txin : tx.vin)
     {
         std::unique_ptr<sigma::CoinSpend> spend;
-        uint32_t pubcoinId;
```
levonpetrosyan93: good rename

src/zerocoin_v3.cpp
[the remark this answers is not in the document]
levonpetrosyan93: I think we don't need this loops. Checking only for provided groupId and denom is enough.

src/zerocoin_v3.cpp
```diff
@@ -965,10 +1043,10 @@ int CSigmaState::GetCoinSetForSpend(
 
 std::pair<int, int> CSigmaState::GetMintedCoinHeightAndId(
         const sigma::PublicCoin& pubCoin) {
-    auto coinIt = mintedPubCoins.find(pubCoin);
+    auto coinIt = containers.GetMints().find(pubCoin);
```
levonpetrosyan93: Why not write auto coinIt = GetMints().find(pubCoin); here? more readable I think.

PR #463, opened 2019-06-18
Fix functions GetHex and SetHex of Scalar

thebevrishot:


nopslide:
Is it safe to fix the secp256k1 lib?

Poramin Insom

thebevrishot:
@nopslide this class just a c++ wrapper of secp256k1_scalar which create by us. And it work incorrectly.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -269,10 +269,10 @@ unsigned char* Scalar::deserialize(unsigned char* buffer) {
...
     unsigned char buffer[32];
-    std::stringstream ss;
-
     serialize(buffer);
 
+    std::stringstream ss;
+    ss << std::hex;
     for (int i = 0; i < 32; ++i) {
         ss << buffer[i] / 16;
         ss << buffer[i] % 16;
```
ultimaweapon: From my perspective this is hard to read. My idea is we should use hex table instead (like my answer here https://stackoverflow.com/a/16827872/1829232). But maybe it is just only me so if other people think the current way is more better then we should go with it.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -269,10 +269,10 @@ unsigned char* Scalar::deserialize(unsigned char* buffer) {
 
 std::string Scalar::GetHex() const {
     unsigned char buffer[32];
-    std::stringstream ss;
-
     serialize(buffer);
 
+    std::stringstream ss;
+    ss << std::hex;
     for (int i = 0; i < 32; ++i) {
```
ultimaweapon: Change hard-coded size to size of `buffer` will be better. My idea is change `buffer` to `std::array` will be better so we can foreach on it or get it size easily.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -281,19 +281,14 @@ std::string Scalar::GetHex() const {
     return ss.str();
 }
 
-void Scalar::SetHex(const std::string& str) const {
+void Scalar::SetHex(const std::string& str) {
```
ultimaweapon: Better to check if `str` have the correct length.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -281,19 +281,14 @@ std::string Scalar::GetHex() const {
...
-    for (int i = 0; i < 32; i+=2)
-        buffer[i] =  str[i] * 16 + str[i + 1];
-
-    int overflow = 0;
-
-    secp256k1_scalar_set_b32(reinterpret_cast<secp256k1_scalar *>(value_), buffer, &overflow);
-
-    if (overflow) {
-        throw "Scalar: decoding overflowed";
+    for (int i = 0; i < 32; i++) {
```
ultimaweapon: Same as previous issue that we should not hard-coded buffer size.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -281,19 +281,14 @@ std::string Scalar::GetHex() const {
...
+        buffer[i] = strtol(str.substr(2 * i, 2).c_str(), NULL, 16);
```
ultimaweapon: We should check if `strtol` success.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -281,19 +281,14 @@ std::string Scalar::GetHex() const {
...
     }
+
+    deserialize(buffer);
```
ultimaweapon: This is quite dangerous if we change the way how it serialized. So better to use `secp256k1_scalar_set_b32` as before.

thebevrishot: make sense

src/secp256k1/src/cpp/Scalar.cpp
thebevrishot: use `::isxdigit` instead.

src/secp256k1/src/cpp/Scalar.cpp
thebevrishot: make sense

src/secp256k1/src/cpp/Scalar.cpp
thebevrishot: fixed

src/secp256k1/src/cpp/Scalar.cpp
thebevrishot: fixed

src/secp256k1/src/cpp/Scalar.cpp
thebevrishot: fixed. but use bare value instead look up hex table.

src/secp256k1/src/cpp/Scalar.cpp
```diff
@@ -268,27 +270,43 @@ unsigned char* Scalar::deserialize(unsigned char* buffer) {
...
 void Scalar::SetHex(const std::string& str) {
-    unsigned char buffer[32];
+    if (str.size() != 64) {
+        throw "Scalar: decoding invalid length";
+    }
+
+    std::array<unsigned char, 32> buffer;
+
+    for (std::size_t i = 0; i < buffer.size(); i++) {
+        auto hexs = str.substr(2 * i, 2).c_str();
```
ultimaweapon: This is a bug. `substr` return a new `std::string` that mean it will destroy immediately once a call to `c_str` is completed.

PR #464, opened 2019-06-20
Fix hdmint test.

thebevrishot:
As I understand this test need to make sure user will get same coins when regenerate by recreate same mints. The test fail because we disable duplicated mint on sigma state. Then I fix it by remove stateful assertion. I believe the remaining test is enough.


riordant:
I see what you're doing here. If duplicated mint is disabled however, if user was to restore mints from the chain, will it work?

PR #474, opened 2019-06-25
Remint improvements

a-bezrukov:


src/wallet/wallet.cpp
```diff
@@ -4387,6 +4386,11 @@ bool CWallet::CreateZerocoinToSigmaRemintModel(string &stringError, int version,
     // currently we don't support zerocoin mints v1
     assert(version == ZEROCOIN_TX_VERSION_2);
 
+    if (!znodeSync.IsBlockchainSynced()) {
```
ultimaweapon: I think we don't need this due to we already have below check. For me it also look weird too that wallet depend on znode.

src/wallet/wallet.cpp
```diff
@@ -4398,6 +4402,13 @@ bool CWallet::CreateZerocoinToSigmaRemintModel(string &stringError, int version,
 
     LOCK2(cs_main, cs_wallet);
 
+    const Consensus::Params &params = Params().GetConsensus();
+    if (chainActive.Height() < params.nSigmaStartBlock ||
```
ultimaweapon: We already have `sigma::IsSigmaAllowed` to check if sigma already activated.

src/qt/zc2sigmapage.cpp
```diff
@@ -114,3 +136,27 @@ void Zc2SigmaPage::selectionChanged() {
...
 }
+
+void Zc2SigmaPage::updateAvailableRemints() {
+    model->updateRows();
+}
+
+bool Zc2SigmaPage::showZc2SigmaPage() {
+    {
+        LOCK(cs_main);
+        if(!znodeSync.IsBlockchainSynced()) {
```
ultimaweapon: If we proceed with the method that I just commented we don't need this.

src/qt/zc2sigmapage.cpp
```diff
@@ -114,3 +136,27 @@ void Zc2SigmaPage::selectionChanged() {
...
+            return false;
+        }
+    }
+
+    {
+        LOCK(cs_main);
+        const Consensus::Params &params = Params().GetConsensus();
+        if (chainActive.Height() < params.nSigmaStartBlock ||
```
ultimaweapon: Will be better to use `sigma::IsSigmaAllowed` instead to centralize logic also it more readable too.

src/qt/zc2sigmapage.h
```diff
@@ -29,21 +31,26 @@ QT_END_NAMESPACE
...
     ~Zc2SigmaPage();
-
     void createModel();
 
+    static bool showZc2SigmaPage();
+
 private:
     Ui::Zc2SigmaPage *ui;
     std::shared_ptr<Zc2SigmaModel> model;
+    std::shared_ptr<QTimer> tmrAvailMints;
```
ultimaweapon: Same as previous issue. We can subscribe to the same event that I commented before.

src/wallet/wallet.cpp
a-bezrukov: Agree

src/qt/bitcoingui.h
```diff
@@ -125,6 +126,8 @@ class BitcoinGUI : public QMainWindow
     RPCConsole *rpcConsole;
     HelpMessageDialog *helpMessageDialog;
 
+    std::shared_ptr<QTimer> checkZc2SigmaPageTimer;
```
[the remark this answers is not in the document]
a-bezrukov: Agree

src/qt/zc2sigmapage.h
a-bezrukov: Not this time as the user can remint his ZC and we'd need to exclude this from the list before the next block comes

src/wallet/wallet.cpp
psolstice: this is not really a dependency on znode. This is a ready-to-use function that returns true if the blockchain is synced. There isn't another one.
If we don't use something like this the client can create remint while syncing/reindexing and the current block height is within the limits

src/wallet/wallet.cpp
psolstice: Sure but this check is cleaner because it shows that height lies within specified window (x >= start && x < start+length)

src/wallet/wallet.cpp
ultimaweapon: I see.

PR #475, opened 2019-06-26
Fix coins balance updating issue on gui

thebevrishot:
- Add wallet notify on HD coins update.
- Fix process crash when try to spend more than 35 coins.

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -879,11 +879,8 @@ WalletModel::SendCoinsReturn WalletModel::prepareSigmaSpendTransaction(
     } catch (const InsufficientFunds& err) {
         return AmountExceedsBalance;
     } catch (const std::runtime_error& err) {
-        if (0 == std::strcmp(
-            err.what(),
-            std::runtime_error(_("Can not choose coins within limit.")).what())) {
+        if (_("Can not choose coins within limit.") == err.what())
             return ExceedLimit;
-        }
         throw err;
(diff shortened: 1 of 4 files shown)


src/qt/walletmodel.cpp
```diff
@@ -878,6 +878,13 @@ WalletModel::SendCoinsReturn WalletModel::prepareSigmaSpendTransaction(
         *newTx = wallet->CreateSigmaSpendTransaction(sendRecipients, fee, selectedCoins, changes, coinControl);
     } catch (const InsufficientFunds& err) {
         return AmountExceedsBalance;
+    } catch (const std::runtime_error& err) {
+        if (0 == std::strcmp(
+            err.what(),
+            std::runtime_error(_("Can not choose coins within limit.")).what())) {
+            return ExceedLimit;
+        }
+        throw err;
```
psolstice: why simple text comparison is so weird here?

psolstice: I mean why this
``if (`_("error message") == err.what())```
won't work?

thebevrishot: It don't work because return value of `what` is a pointer to a null-terminated string.

https://en.cppreference.com/w/cpp/error/exception/what

thebevrishot: thanks for fix. I don't know std::string and null-terminated string can compare this way before.

https://en.cppreference.com/w/cpp/string/basic_string/operator_cmp

PR #479, opened 2019-07-01
Handle disconnectTip for sigma.

thebevrishot:
Fix following problem when a block is disconnected.
1. GUI will not recalculate balance of private coin.
- solve this one by add condition to recalculate sigma balance function to let it work when ever current block number is less that cached block bumber.
2. hdmint tracker not reset state of mint.
- solve this one by calling `UpdateSpendStateFromBlock` when every disconnectTip is called.


src/main.cpp
```diff
@@ -3342,6 +3342,13 @@ bool static DisconnectTip(CValidationState &state, const CChainParams &chainpara
     if (!ReadBlockFromDisk(block, pindexDelete, chainparams.GetConsensus()))
         return AbortNode(state, "Failed to read block");
 
+    // retrieve all mints
+    block.sigmaTxInfo = std::make_shared<sigma::CSigmaTxInfo>();
+    for (auto const& tx : block.vtx) {
+        CheckTransaction(tx, state, tx.GetHash(), false, pindexDelete->pprev->nHeight,
```
levonpetrosyan93: Why we need to call CheckTransaction() here?

thebevrishot: CheckTransaction provide logic to retrieve mint and serial if we pass sigmaTxInfo with. Also I need to use same logic to retrieve instead of retrieve it manually.

levonpetrosyan93: What you mean retrieve?

thebevrishot: actually, just get mint/spend list from block.

levonpetrosyan93: Ok got it.

PR #480, opened 2019-07-01
Remint in insight

a-bezrukov:


src/script/standard.cpp
```diff
@@ -98,6 +98,16 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, vector<vector<unsi
         return true;
     }
 
+    // Zerocoin to Sigma remint
+    if (scriptPubKey.IsZerocoinRemint())
+    {
+        typeRet = TX_ZEROCOINTOSIGMAREMINT;
```
thebevrishot: please add scriptPubkey size assertion.

a-bezrukov: Sure

thebevrishot: Wait!, sorry I just notice that. As I understand function `Solver` is use only with output.

a-bezrukov: From it's signature, it can be used with any CScript object. There is nothing wrong in using it with inputs. Actually, the code here is not used as I found a shorter way of handing this. Nonetheless, this code is useful for debugging.

thebevrishot: The first parameter name and type of the second parameter which is `enum txnouttype` indicate clearly what kind of script need to be handled by the function. Also I don't think debugging purpose is good enough reason for add unused code.

a-bezrukov: Ok, removed

a-bezrukov: Here you are

thebevrishot: LGTM. thanks for removing.

PR #482, opened 2019-07-02
HD Mint upgrades and fixes

riordant:
- Add "zapwalletmints" conf setting
- Fixes for encrpytion of wallet - now storing all previous mints indexed by pubcoin hash
- Various efficiency improvements
- Code refactor
- Use `CMintMeta` objects over `CSigmaEntry` when showing coins in GUI - this allows for data retrieval when wallet is locked (mint secret data is now part of the HDChain and so requires unlock to access) and means that we do not have to regenerate mints for every access
- Add walletDB functions to get pairs of serial hashes and pubCoins (for easier regeneration of mint from chain)

Note that due to WalletDB updates, **existing testnet HD Sigma mints (ie. those generated after the merge of https://github.com/zcoinofficial/zcoin/pull/392) will not be accessible.** The amount of code required to be added to make these accessible is not worth the time at this point. Zerocoins and non-HD Sigma mints are unaffected by this.


src/hdmint/wallet.cpp
```diff
@@ -324,22 +315,29 @@ bool CHDMintWallet::SeedToZerocoin(const uint512& seedZerocoin, GroupElement& co
...
 
-CKeyID CHDMintWallet::GetZerocoinSeedID(uint32_t n){
+CKeyID CHDMintWallet::GetZerocoinSeedData(int32_t nCount, uint160 hashSeedMaster){
     // Get CKeyID for n from mintpool
-    std::pair<CKeyID,uint32_t> mintPoolEntry;
-    if(!mintPool.Get(n, mintPoolEntry)){
+    uint256 hashPubcoin;
+    std::pair<uint256,MintPoolEntry> mintPoolEntryPair;
+
+    while(!mintPool.Get(nCount, hashSeedMaster, mintPoolEntryPair)){
```
thebevrishot: If caller pass invalid hashSeedMaster by accident then this loop will run forever.

src/hdmint/wallet.cpp
```diff
@@ -324,22 +315,29 @@ bool CHDMintWallet::SeedToZerocoin(const uint512& seedZerocoin, GroupElement& co
...
         GenerateMintPool();
-        mintPool.Get(n, mintPoolEntry);
     }
 
-    return mintPoolEntry.first;
+    return get<1>(mintPoolEntryPair.second);
 }
 
-uint512 CHDMintWallet::GetZerocoinSeed(uint32_t n, CKeyID& seedId)
+uint512 CHDMintWallet::CreateZerocoinSeed(int32_t& n, CKeyID& seedId, bool checkIndex)
```
thebevrishot: If we do not modify n, please do not pass by reference.

src/hdmint/wallet.h
```diff
@@ -18,40 +18,39 @@ class CHDMint;
...
-    int static const COUNT_LAST_USED_DEFAULT = 0;
+    int static const COUNT_DEFAULT = 0;
 
     CHDMintWallet(std::string strWalletFile);
 
-    bool SetHashSeedMaster(const uint160& hashSeedMaster, bool fResetCount=false);
-    void SyncWithChain(bool fGenerateMintPool = true);
-    uint32_t GenerateHDMint(sigma::CoinDenomination denom, sigma::PrivateCoin& coin, CHDMint& dMint, bool fGenerateOnly = false);
-    bool GenerateMint(const uint32_t& nCount, const sigma::CoinDenomination denom, CKeyID seedId, sigma::PrivateCoin& coin, CHDMint& dMint);
+    bool ReadIsCrypted();
```
thebevrishot: Never define.

src/hdmint/wallet.h
```diff
@@ -18,40 +18,39 @@ class CHDMint;
...
+    bool WriteIsCrypted(bool isCrypted);
```
thebevrishot: This one too.

src/wallet/wallet.cpp
```diff
@@ -7148,6 +7130,16 @@ DBErrors CWallet::ZapWalletTx(std::vector <CWalletTx> &vWtx) {
     return DB_LOAD_OK;
 }
 
+DBErrors CWallet::ZapSigmaMints() {
+    if (!fFileBacked)
+        return DB_LOAD_OK;
+    DBErrors nZapSigmaMintRet = CWalletDB(strWalletFile, "cr+").ZapSigmaMints(this);
+    if (nZapSigmaMintRet != DB_LOAD_OK)
+        return nZapSigmaMintRet;
```
thebevrishot: Need to log something here?

src/wallet/walletdb.cpp
```diff
@@ -977,6 +983,31 @@ DBErrors CWalletDB::ZapSelectTx(CWallet *pwallet, vector <uint256> &vTxHashIn, v
     return DB_LOAD_OK;
 }
 
+DBErrors CWalletDB::ZapSigmaMints(CWallet *pwallet) {
```
thebevrishot: please rename function to more descriptive name.

src/hdmint/mintpool.cpp
```diff
@@ -7,109 +7,55 @@
...
-    if (pMint.second >= nCountLastGenerated)
-        nCountLastGenerated = pMint.second;
 
     if (fVerbose)
-        LogPrintf("%s : add %s count %d to mint pool\n", __func__, pMint.first.GetHex().substr(0, 6), pMint.second);
+        LogPrintf("%s : add %s count %d to mint pool\n", __func__, pMint.first.GetHex().substr(0, 6), get<2>(pMint.second));
 }
 
-bool CMintPool::Has(const CKeyID& seedId)
+bool SortSmallest(const pair<uint256, MintPoolEntry> a, const pair<uint256, MintPoolEntry> b)
```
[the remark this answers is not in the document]
riordant: Agree

src/hdmint/mintpool.cpp
```diff
@@ -7,109 +7,55 @@
...
     for (auto pMint : *(this)) {
-        if(pMint.first==seedId){
-           result = pMint;
-           return true; 
-        }
+        listMints.emplace_back(pMint);
     }
-    return false;
+
+    listMints.sort(SortSmallest);
```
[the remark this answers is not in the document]
riordant: I've updated it to sort by the third argument of `mintPoolEntry`, the count. (It's not a requirement that it be sorted anyway)

src/hdmint/tracker.cpp
```diff
@@ -481,76 +482,50 @@ bool CHDMintTracker::MintMetaToZerocoinEntries(std::list <CSigmaEntry>& entries,
...
-        CMintMeta meta = GetMetaFromPubcoin(hashPubcoin);
-        if(UpdateStatusInternal(setMempool, meta)){
-            updatedMeta.emplace_back(meta);
+        // Check hashPubcoin in db
+        if(walletdb.ReadMintPoolPair(hashPubcoin, hashSeedMasterEntry, seedId, nCount)){
+            // If found in db but not in memory - this is likely a resync
+            if(!HasPubcoinHash(hashPubcoin)){
+                MintPoolEntry mintPoolEntry(hashSeedMasterEntry, seedId, nCount);
+                mintPoolEntries.push_back(std::make_pair(hashPubcoin, mintPoolEntry));
+                continue;
```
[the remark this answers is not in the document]
riordant: This part is only called if the mint is found in memory as well as db. If it's found in DB but NOT in memory, we use `SyncWithChain` to create an entry for the mint and store it (see the `SetMintSeedSeen` function). This is what will happen during a `zapsigmamint` or if the wallet corrupted and a backup is used to "catch up" with the chain

src/hdmint/tracker.cpp
```diff
@@ -559,13 +534,27 @@ void CHDMintTracker::UpdateSpendStateFromBlock(const sigma::spend_info_container
...
+        // Check serialHash in db
+        if(walletdb.ReadSerialHash(spentSerialHash, hashPubcoin)){
+            // If found in db but not in memory - this is likely a resync
+            if(!Get(spentSerialHash, meta)){
+                if(!walletdb.ReadMintPoolPair(hashPubcoin, hashSeedMasterEntry, seedId, nCount)){
+                    continue;
+                }
+                MintPoolEntry mintPoolEntry(hashSeedMasterEntry, seedId, nCount);
+                mintPoolEntries.push_back(std::make_pair(hashPubcoin, mintPoolEntry));
+                continue;
```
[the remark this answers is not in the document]
riordant: Same comment as above.

src/hdmint/wallet.cpp
```diff
@@ -50,91 +48,81 @@ bool CHDMintWallet::SetHashSeedMaster(const uint160& hashSeedMaster, bool fReset
...
+    nCountLastGenerated = COUNT_DEFAULT;
 
-    if (fResetCount)
+    if (fResetCount){
         walletdb.WriteZerocoinCount(nCountLastUsed);
-    else if (!walletdb.ReadZerocoinCount(nCountLastUsed))
-        nCountLastUsed = COUNT_LAST_USED_DEFAULT;
-
-    mintPool.Reset();
+        walletdb.WriteZerocoinSeedCount(nCountLastUsed);
```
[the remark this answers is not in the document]
riordant: Yes, as discussed, updated.

src/hdmint/wallet.cpp
```diff
@@ -145,9 +133,12 @@ void CHDMintWallet::SyncWithChain(bool fGenerateMintPool)
...
-        for (pair<CKeyID, uint32_t> pMint : listMints) {
+        std::set<uint256> setChecked;
+        if(listMints==boost::none){
+            listMints = list<pair<uint256, MintPoolEntry>>();
+            mintPool.List(listMints.get());
+        }
+        for (pair<uint256, MintPoolEntry>& pMint : listMints.get()) {
             LOCK(cs_main);
             if (setChecked.count(pMint.first))
                 return;
```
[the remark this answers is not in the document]
riordant: Correct, I think continue is best. There is no breaking change if the same coin is added twice.

src/hdmint/wallet.cpp
riordant: `hashSeedMaster` will always be set by this point. However I will add a check that it is initialized.

src/hdmint/wallet.h
riordant: Removed

src/hdmint/wallet.h
riordant: Removed

src/wallet/walletdb.cpp
riordant: What do you suggest? It's just derived from `ZapWalletTx`..

src/hdmint/wallet.cpp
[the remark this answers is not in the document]
thebevrishot: ok

src/wallet/walletdb.cpp
thebevrishot: After discussion, I agree with add comment instead.

src/wallet/wallet.cpp
riordant: Added log

src/wallet/walletdb.cpp
riordant: After discussion, added description of function operation to the header file. (Keeping similar format to `ZapWalletTx` for consistency)

src/hdmint/mintpool.cpp
```diff
@@ -7,109 +7,55 @@
...
-
-void CMintPool::Reset()
-{
-    clear();
-    nCountLastGenerated = 0;
-    nCountLastRemoved = 0;
-}
-
-bool CMintPool::Front(std::pair<CKeyID, uint32_t>& pMint)
+bool CMintPool::Front(pair<uint256, MintPoolEntry> pMint)
```
levonpetrosyan93: I could not find any use of this. If it is necessary, than we need to pass by reference I think.

src/hdmint/mintpool.cpp
```diff
@@ -7,109 +7,55 @@
...
 {
     if (empty())
         return false;
     pMint = *begin();
     return true;
 }
 
-bool CMintPool::Next(pair<CKeyID, uint32_t>& pMint)
+bool CMintPool::Next(pair<uint256, MintPoolEntry> pMint)
```
levonpetrosyan93: Same here.

src/hdmint/tracker.cpp
```diff
@@ -475,85 +476,50 @@ bool CHDMintTracker::UpdateStatusInternal(const std::set<uint256>& setMempool, C
...
 
+    if(mintPoolEntries.size()>0)
+        zwalletMain->SyncWithChain(false, mintPoolEntries);
+
     //overwrite any updates
     for (CMintMeta& meta : updatedMeta)
         UpdateState(meta);
 }
 
 void CHDMintTracker::UpdateSpendStateFromBlock(const sigma::spend_info_container& spentSerials){
```
levonpetrosyan93: I see we have duplication of code in this 2 functions. Would be better somehow prevent that.

src/hdmint/mintpool.cpp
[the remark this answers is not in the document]
thebevrishot: ok, cool.

src/hdmint/tracker.cpp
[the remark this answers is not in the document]
thebevrishot: ok

src/hdmint/wallet.cpp
thebevrishot: it's safe from infinity loop now.

src/hdmint/wallet.cpp
```diff
@@ -315,27 +313,33 @@ bool CHDMintWallet::SeedToZerocoin(const uint512& seedZerocoin, GroupElement& co
...
     std::pair<uint256,MintPoolEntry> mintPoolEntryPair;
 
-    while(!mintPool.Get(nCount, hashSeedMaster, mintPoolEntryPair)){
+    if(!mintPool.Get(nCount, hashSeedMaster, mintPoolEntryPair)){
         // Top up mintpool if empty
         GenerateMintPool();
+        if(!mintPool.Get(nCount, hashSeedMaster, mintPoolEntryPair)){
+            ResetCount();
+            throw ZerocoinException("Unable to retrieve mint seed ID");
+        }
```
thebevrishot: Another issue, seem like `GenerateMintPool` generate only next 20 mints if current state is 0 but we need to find at index 1000 it'll never be find. I think we should pass required nCount to function `GenerateMintPool`.

src/wallet/wallet.cpp
thebevrishot: cool

src/hdmint/wallet.cpp
thebevrishot: Actually I except something like `int32_t n`.

riordant: It's passing by const reference now?

src/hdmint/tracker.cpp
riordant: Created new function `UpdateFromBlock`

src/hdmint/wallet.cpp
riordant: Updated following discussion

riordant:
@thebevrishot  @levonpetrosyan93 Updated to include fixes for your comments, please review.

PR #487, opened 2019-07-04
Disable Exodus from building by default

ultimaweapon:


TomaszWaszczyk:
Exodus in done? What is the status of Exodus?

thebevrishot:
@TomaszWaszczyk Exodus is done long time ago. This one is just an improvement.

PR #490, opened 2019-07-05
RPC tests: encrypted wallet, zapsigmamints

FreakyMaryk:


qa/rpc-tests/sigma_remint.py
```diff
@@ -56,6 +56,12 @@ def run_test(self):
         assert len(sigma_mint) == 12, 'Should be 12 sigma mints after remint, but was: {}' \
                 .format(len(sigma_mint))
 
+        sigma_mints = set([mint['amount'] for mint in sigma_mint])
+        expected_mints = set([Decimal('25.00000000'), Decimal('10.00000000'), Decimal('100.00000000'), Decimal('1.00000000')])
```
thebevrishot: why use set instead of list?

qa/rpc-tests/sigma_remint_lockedwallet.py
```diff
@@ -0,0 +1,122 @@
...
+                self.nodes[0].remintzerocointosigma(denom)
+            except JSONRPCException as e:
+                assert False, "Could not remint denomination {} with next exception {}." \
+                    .format(denom, e.error['message'])
+
+        self.nodes[0].generate(50)
+
+        sigma_mint = self.nodes[0].listunspentsigmamints()
+
+        assert len(sigma_mint) == len(zcoin_denoms)+1, \
```
thebevrishot: more descriptive if you count amount of remints instead of using `len(zcoin_denoms)+1`.

qa/rpc-tests/sigma_remint_lockedwallet.py
```diff
@@ -0,0 +1,122 @@
...
+            'Looks like sigma mints unspendable after remint on encrypted wallet.'
+
+        # check that we are able to mint/spend
+        self.nodes[0].mint(1)
+        self.nodes[0].generate(10)
+
+        sigma_mint = self.nodes[0].listunspentsigmamints()
```
thebevrishot: could you use `sigma_mints` instead?

qa/rpc-tests/sigma_zapsigmamints.py
```diff
@@ -0,0 +1,72 @@
...
+    def setup_nodes(self):
+        # This test requires mocktime
+        enable_mocktime()
+        return start_nodes(self.num_nodes, self.options.tmpdir)
+
+    def run_test(self):
+        getcontext().prec = 6
+        self.sync_all()
+
+        zcoin_denoms = [1, 10, 25, 50, 100]
```
thebevrishot: why use zerocoin denomination?

qa/rpc-tests/sigma_remint.py
FreakyMaryk: Cause we checked the len before. Here just quick check that all denoms shown

qa/rpc-tests/sigma_remint_lockedwallet.py
FreakyMaryk: Yes, but what's the difference?

qa/rpc-tests/sigma_zapsigmamints.py
FreakyMaryk: Yes, can rename here

thebevrishot: I mean denominations. I expect something like `denoms = [0.05, 0.1, 0.5, 1, 10, 25, 100]`

qa/rpc-tests/sigma_remint_lockedwallet.py
thebevrishot: it's list, then the name should be plural.

qa/rpc-tests/sigma_remint.py
thebevrishot: ok

PR #493, opened 2019-07-06
Fix wallet swell

a-bezrukov:


src/wallet/wallet.cpp
```diff
@@ -1310,7 +1278,7 @@ isminetype CWallet::IsMine(const CTxOut &txout) const {
             return ISMINE_NO;
         }
 
-        return db.HasZerocoinEntry(pub) ? ISMINE_SPENDABLE : ISMINE_NO;
+        return CZerocoinState::GetZerocoinState()->HasCoin(pub) ? ISMINE_SPENDABLE : ISMINE_NO;
```
psolstice: If the check is performed before ConnectBlock the answer is always "no". If the check is performed after ConnectBlock the answer is always "yes". Doesn't seem right to me

src/wallet/wallet.cpp
```diff
@@ -1188,25 +1188,7 @@ isminetype CWallet::IsMine(const CTxIn &txin) const {
...
-            SER_NETWORK,
-            PROTOCOL_VERSION
-        );
-
-        libzerocoin::CoinSpend spend(v2 ? ZCParamsV2 : ZCParams, data);
-
-        if (db.HasCoinSpendSerialEntry(spend.getCoinSerialNumber())) {
-            return ISMINE_SPENDABLE;
-        }
+        return ISMINE_NO;
```
psolstice: To be clear: do we just give up on marking zerocoin spends as ours even if they are ours?

src/wallet/wallet.cpp
levonpetrosyan93: How this solves wallet swelling?

src/wallet/wallet.cpp
reubenyap: Our own Zerocoin spends should still be identified.

a-bezrukov: Yes, we are not identifying our spends. This is the behaviour that we had since the beginning. Only spends to own address is identified.

src/wallet/wallet.cpp
```diff
@@ -1300,18 +1268,7 @@ CAmount CWallet::GetDebit(const CTxIn &txin, const isminefilter &filter) const {
...
-
-        try {
-            pub = ParseZerocoinMintScript(txout.scriptPubKey);
-        } catch (std::invalid_argument&) {
-            return ISMINE_NO;
-        }
-
-        return db.HasZerocoinEntry(pub) ? ISMINE_SPENDABLE : ISMINE_NO;
-    } else if (txout.scriptPubKey.IsSigmaMint()) {
+    if (txout.scriptPubKey.IsSigmaMint()) {
```
psolstice: This one will cover zerocoin to sigma remints too. Need to test this case to see if nothing breaks

PR #496, opened 2019-07-09
List from mempool fix

riordant:
Fixes issue where if the user closes wallet without spend mined (ie. mempool entry), on restart, spend still displays as "Spendable"

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -866,6 +866,9 @@ void ThreadImport(std::vector <boost::filesystem::path> vImportFiles) {
     if (!GetBoolArg("-disablewallet", false)) {
         zwalletMain->SyncWithChain();
     }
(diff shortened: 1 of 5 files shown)


src/init.cpp
```diff
@@ -866,6 +866,9 @@ void ThreadImport(std::vector <boost::filesystem::path> vImportFiles) {
     if (!GetBoolArg("-disablewallet", false)) {
         zwalletMain->SyncWithChain();
     }
+    if (GetBoolArg("-zapwallettxes", false)) {
```
a-bezrukov: Is this for debugging?

riordant: No, if `zapwallettxes` is used it zaps the mempool, so we need to perform an update of the meta objects. If you look at `InitLoadWallet` is does something similar - it performs an update of the status mint objects if `zapwallettxes` is used

PR #497, opened 2019-07-09
Disable sigma on legacy wallet

thebevrishot:
- disable rpc
- disable zwallet
- disble GUI and show message below on sigma and remint tab.

```
We have detected that your wallet does not use a HD seed. A HD seed allows you to restore all your private keys and also Sigma mints even when restoring from an old backup.

To have Sigma private transactions functionality, a HD wallet is required. We highly recommend that you create a new wallet and move your funds over to the new wallet version or use the dumpwallet in conjunction with importprivkey commands to move the keys to the new wallet.
```

GUI when use nonhd wallet are uploaded to link below.

https://trello.com/c/6avQbIke/197-ask-users-to-upgrade-old-wallet-to-hd-wallet

### DIFF
--- a/src/hdmint/tracker.cpp
+++ b/src/hdmint/tracker.cpp
@@ -475,7 +475,7 @@ bool CHDMintTracker::UpdateMetaStatus(const std::set<uint256>& setMempool, CMint
 }
 
 void CHDMintTracker::UpdateFromBlock(const std::list<std::pair<uint256, MintPoolEntry>>& mintPoolEntries, const std::vector<CMintMeta>& updatedMeta){
-    if(mintPoolEntries.size()>0)
+    if(mintPoolEntries.size()>0 && zwalletMain)
         zwalletMain->SyncWithChain(false, mintPoolEntries);
 
     //overwrite any updates
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -566,10 +566,14 @@ static void NotifyZerocoinChanged(WalletModel *walletmodel, CWallet *wallet, con
                               Q_ARG(QString, QString::fromStdString(pubCoin)),
                               Q_ARG(QString, QString::fromStdString(isUsed)),
                               Q_ARG(int, status));
-    QMetaObject::invokeMethod(walletmodel, "updateSigmaCoins", Qt::QueuedConnection,
+
+    // disable sigma
+    if (pwalletMain->IsHDSeedAvailable()) {
+        QMetaObject::invokeMethod(walletmodel, "updateSigmaCoins", Qt::QueuedConnection,
                               Q_ARG(QString, QString::fromStdString(pubCoin)),
                               Q_ARG(QString, QString::fromStdString(isUsed)),
                               Q_ARG(int, status));
+    }
 }
 
 static void NotifyTransactionChanged(WalletModel *walletmodel, CWallet *wallet, const uint256 &hash, ChangeType status)
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -139,7 +139,10 @@ void WalletModel::pollBalanceChanged()
             transactionTableModel->updateConfirmations();
 
         // check sigma
-        checkSigmaAmount(false);
+        // support only hd
+        if (pwalletMain->IsHDSeedAvailable()) {
+            checkSigmaAmount(false);
+        }
     }
 }
 
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -397,7 +397,11 @@ void WalletView::gotoSigmaPage()
 
 void WalletView::gotoZc2SigmaPage()
 {
-    setCurrentWidget(zc2SigmaPage);
+    if (pwalletMain->IsHDSeedAvailable()) {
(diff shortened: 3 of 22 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -75,6 +75,16 @@ bool EnsureWalletIsAvailable(bool avoidException)
     return true;
 }
 
+bool EnsureHDSeedIsAvailable(bool avoidException = false) {
```
ultimaweapon: This look weird to have a flags to control exception should raise or not. We can move the caller after help message and remove this flag.

src/wallet/rpcwallet.cpp
```diff
@@ -75,6 +75,16 @@ bool EnsureWalletIsAvailable(bool avoidException)
...
+    if (!pwalletMain || !pwalletMain->IsHDSeedAvailable()) {
```
ultimaweapon: No need to check `pwalletMain` is null due to we are in wallet code. If it null that mean something wrong.

src/qt/walletview.cpp
```diff
@@ -235,7 +242,9 @@ void WalletView::setClientModel(ClientModel *clientModel)
     sendZcoinView->setClientModel(clientModel);
     znodeListPage->setClientModel(clientModel);
     exoAssetsPage->setClientModel(clientModel);
-    sigmaView->setClientModel(clientModel);
+    if (pwalletMain->IsHDSeedAvailable()) {
```
ultimaweapon: We should check if `sigmaView` is not null instead so the logic to control the visibility of `sigmaView` will be in the single place.

src/qt/walletview.cpp
```diff
@@ -256,7 +265,9 @@ void WalletView::setWalletModel(WalletModel *walletModel)
     overviewPage->setWalletModel(walletModel);
     receiveCoinsPage->setModel(walletModel);
     zerocoinPage->setModel(walletModel->getAddressTableModel());
-    sigmaView->setWalletModel(walletModel);
+    if (pwalletMain->IsHDSeedAvailable()) {
```
ultimaweapon: Same here.

src/qt/walletview.cpp
```diff
@@ -386,7 +397,11 @@ void WalletView::gotoSigmaPage()
 
 void WalletView::gotoZc2SigmaPage()
 {
-    setCurrentWidget(zc2SigmaPage);
+    if (pwalletMain->IsHDSeedAvailable()) {
+        setCurrentWidget(zc2SigmaPage);
+    } else {
+        setCurrentWidget(sigmaPage);
```
ultimaweapon: Better to provide comment here so the other people will not confuse why if no HD seed we will switch to sigma page instead of remint page.

src/hdmint/tracker.cpp
```diff
@@ -475,7 +475,7 @@ bool CHDMintTracker::UpdateMetaStatus(const std::set<uint256>& setMempool, CMint
 }
 
 void CHDMintTracker::UpdateFromBlock(const std::list<std::pair<uint256, MintPoolEntry>>& mintPoolEntries, const std::vector<CMintMeta>& updatedMeta){
-    if(mintPoolEntries.size()>0)
+    if(mintPoolEntries.size()>0 && zwalletMain)
```
ultimaweapon: No need to check here since tracker should not get invoked any methods if HD mints is not activated.

src/wallet/sigmaspendbuilder.cpp
```diff
@@ -161,7 +161,9 @@ CAmount SigmaSpendBuilder::GetChanges(std::vector<CTxOut>& outputs, CAmount amou
 
         sigma::PrivateCoin newCoin(params, denomination, ZEROCOIN_TX_VERSION_3);
         hdMint.SetNull();
-        zwalletMain->GenerateMint(denomination, newCoin, hdMint);
+        if (zwalletMain) {
```
ultimaweapon: Same as previous issue. We should not store `hdMint` if it not enabled.

src/qt/walletmodel.cpp
```diff
@@ -139,7 +139,10 @@ void WalletModel::pollBalanceChanged()
             transactionTableModel->updateConfirmations();
 
         // check sigma
-        checkSigmaAmount(false);
+        // support only hd
+        if (pwalletMain->IsHDSeedAvailable()) {
```
ultimaweapon: Will be better to checking in `checkSigmaAmount` instead so it will not need to check everywhere that invoked it.

src/qt/walletmodel.cpp
```diff
@@ -563,10 +566,14 @@ static void NotifyZerocoinChanged(WalletModel *walletmodel, CWallet *wallet, con
                               Q_ARG(QString, QString::fromStdString(pubCoin)),
                               Q_ARG(QString, QString::fromStdString(isUsed)),
                               Q_ARG(int, status));
-    QMetaObject::invokeMethod(walletmodel, "updateSigmaCoins", Qt::QueuedConnection,
+
+    // disable sigma
+    if (pwalletMain->IsHDSeedAvailable()) {
```
ultimaweapon: Same here.

PR #499, opened 2019-07-10
Add exodus sigma activated checking function and consensus param

thebevrishot:
https://trello.com/c/9RW5PQZ7/529-add-new-consensus-param-to-control-token-mint-spend-feature-1


thebevrishot:
wrong place

PR #500, opened 2019-07-10
Add sigma feature flag on exodus

thebevrishot:


src/exodus/rules.h
```diff
@@ -121,6 +123,9 @@ class CConsensusParams
     //! Block to activate the waiting period for enabling managed property address freezing
     int FREEZENOTICE_FEATURE_BLOCK;
 
+    //! Block to activate sigma
+    int EXODUS_SIGMA_STARTING_BLOCK;
```
ultimaweapon: It name should be `SIGMA_EATURE_BLOCK`.

PR #502, opened 2019-07-10
importprivkey -> importwallet recommendation

catchingknives:
Change recommended command from `importprivkey` (which only imports individual private keys) to `importwallet` (which imports the whole wallet).


a-bezrukov:
The revision it was branched from had an issue

PR #504, opened 2019-07-10
Fix segmentation fault for non-HD wallet

ultimaweapon:
The reason for remove `qa/rpc-tests/enable_hdsigma.py` is because it should already covered by the other tests that use those RPCs already.


src/wallet/sigmaspendbuilder.cpp
```diff
@@ -105,6 +105,10 @@ static std::unique_ptr<SigmaSpendSigner> CreateSigner(const CSigmaEntry& coin)
 
 SigmaSpendBuilder::SigmaSpendBuilder(CWallet& wallet, const CCoinControl *coinControl) : TxBuilder(wallet)
 {
+    if (!zwalletMain) {
+        throw std::logic_error("Sigma feature required HD wallet");
```
thebevrishot: should be just `require`

thebevrishot: also I think better to create guard function instead. because we use this several time.

src/wallet/wallet.cpp
```diff
@@ -2098,16 +2096,18 @@ std::vector<CRecipient> CWallet::CreateSigmaMintRecipients(
     std::vector<sigma::PrivateCoin>& coins,
     vector<CHDMint>& vDMints)
 {
+    if (!zwalletMain) {
```
thebevrishot: same here

src/wallet/wallet.cpp
```diff
@@ -2207,10 +2207,14 @@ CAmount CWallet::SelectSpendCoinsForAmount(
 }
 
 std::list<CSigmaEntry> CWallet::GetAvailableCoins(const CCoinControl *coinControl, bool includeUnsafe) const {
+    if (!zwalletMain) {
```
thebevrishot: same here

src/wallet/sigmaspendbuilder.cpp
ultimaweapon: I'm not grammar mastery so let's ask other people. @reubenyap what do you think?

ultimaweapon: @riordant just provided a different one `Sigma feature requires HD wallet`.

reubenyap: @riordant suggestion is okay

PR #508, opened 2019-07-12
Dump/Import wallet RPC Updates

riordant:
- Output Key metadata in `dumpwallet`
- Restore Key metadata in `importwallet`
- Use Key metadata to restore `MintPool` objects from previous wallet, and sync mints with chain.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -444,6 +444,8 @@ UniValue importwallet(const UniValue& params, bool fHelp)
 
     bool fGood = true;
 
+    bool fMintUpdate = false;
+
     int64_t nFilesize = std::max((int64_t)1, (int64_t)file.tellg());
     file.seekg(0, file.beg);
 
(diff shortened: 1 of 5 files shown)


src/wallet/rpcdump.cpp
```diff
@@ -679,7 +710,13 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
             } else {
                 file << "change=1";
```
thebevrishot: when export hd sigma this one should printed valid value or should not be shown at all.

riordant: @thebevrishot Could you clarify what you mean by this? Do you mean that if it is a mint it should show `change=2`?

thebevrishot: sorry, I'm confused this line with nChange in hdkeypath.

riordant: Ah sure, ok.

PR #515, opened 2019-07-14
Sigma Cleanup

riordant:
- Remove Mint entry from coin control following spend
- Consider HD mints in `AbandonTransaction` function
- General refactoring of `Tracker` class

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1091,7 +1091,7 @@ bool CWallet::AbandonTransaction(const uint256 &hashTx) {
             uint256 hashSerial = primitives::GetSerialHash(serial);
             CMintMeta meta;
             if(zwalletMain->GetTracker().GetMetaFromSerial(hashSerial, meta)){
-                meta.isUsed = true;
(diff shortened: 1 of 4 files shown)


src/wallet/wallet.cpp
```diff
@@ -1089,25 +1088,16 @@ bool CWallet::AbandonTransaction(const uint256 &hashTx) {
...
-
-                    // erase zerocoin spend entry
-                    CSigmaSpendEntry spendEntry;
-                    spendEntry.coinSerial = serial;
-                    walletdb.EraseCoinSpendSerialEntry(spendEntry);
-                }
+            uint256 hashSerial = primitives::GetSerialHash(serial);
+            CMintMeta meta;
+            if(zwalletMain->GetTracker().GetMetaFromSerial(hashSerial, meta)){
+                meta.isUsed = true;
```
ultimaweapon: We should mark it as unused, is not it?

riordant: Yes correct. Updated

PR #516, opened 2019-07-14
Patching Sigma Generator creation

levonpetrosyan93:
A concern was brought up on the public verifiability of the randomness of g in the Sigma generator. While we aren't aware of how this can be exploited as there is no discrete log relationship between g and h, the team agreed that it was best practice to have the generation of g publicly verifiable.

This PR makes g as a hash of the base point of secp256k1 library.

### DIFF
--- a/src/sigma/params.cpp
+++ b/src/sigma/params.cpp
@@ -11,7 +11,7 @@ Params* Params::get_default() {
         //fixing generator G;
         GroupElement g;
 
-        if(::Params().GenConsensus().IsMain()) {
+        if(::Params().GetConsensus().IsMain()) {
             unsigned char buff[32] = {0};
             GroupElement base;
             base.set_base_g();
--- a/src/sigma/params.cpp
+++ b/src/sigma/params.cpp
@@ -11,7 +11,7 @@ Params* Params::get_default() {
         //fixing generator G;
         GroupElement g;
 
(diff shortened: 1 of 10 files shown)


src/sigma/params.cpp
```diff
@@ -10,6 +11,13 @@ Params* Params::get_default() {
         //fixing generator G;
         GroupElement g("9216064434961179932092223867844635691966339998754536116709681652691785432045",
                        "33986433546870000256104618635743654523665060392313886665479090285075695067131");
+
+        if((::Params().NetworkIDString() == CBaseChainParams::MAIN)) {
```
psolstice: `::Params().GenConsensus().IsMain()` is the preferred way across the code now

src/sigma/params.cpp
```diff
@@ -10,6 +11,13 @@ Params* Params::get_default() {
...
+            unsigned char buff[32] = {0};
+            g.sha256(buff);
```
psolstice: this is still something suspicious. Yes, it's hash but from value that is of unknown origin. Can we just take a hash of something simpler?

levonpetrosyan93: Yes,agree will replace it with base point now.

src/sigma/params.cpp
levonpetrosyan93: Ok thank you will change.

PR #520, opened 2019-07-16
Ban 0.13.8.1 from the network

psolstice:


nopslide:
This is temporary solution right ? Coz someone can rename it from client side.

psolstice:
they can but there is no rational reason to do that

src/zerocoin_params.h
```diff
@@ -42,6 +42,9 @@ static const int64_t DUST_HARD_LIMIT = 1000;   // 0.00001 XZC mininput
 #define ZC_SIGMA_STARTING_BLOCK         182030 //Approx July 23rd, 2019, 8:00 AM UTC
 #define ZC_SIGMA_TESTNET_STARTING_BLOCK 50000
 
+// The block number after which old sigma clients are banned.
+#define ZC_OLD_SIGMA_BAN_BLOCK          181850 //Approx July 21st, 2019, 4:00 AM UTC
```
levonpetrosyan93: Need to update this number also I think

a-bezrukov: This number is OK - July 21st, early Monday morning.

thebevrishot: I think, It should be about 181540

a-bezrukov: The comment was wrong. It should be 22nd early morning.

thebevrishot: then make sense

PR #528, opened 2019-07-18
Fixing CS assertion

a-bezrukov:


a-bezrukov:
CheckFinalTx requires cs_main to be held

PR #529, opened 2019-07-19
Add flag to control sigma feature for token

ultimaweapon:


src/exodus/sp.h
```diff
@@ -122,6 +133,20 @@ class CMPSPInfo : public CDBBase
...
             READWRITE(manual);
             READWRITE(historicalData);
+
+            if (ser_action.ForRead()) {
+                // If it is EOF when trying to read additional field that mean it is data before we introduced it.
+                try {
+                    READWRITE(sigmaStatus);
+                } catch (std::ios_base::failure&) {
+                    // Assume it is EOF due to no other better way to check.
+                    sigmaStatus = static_cast<uint8_t>(SigmaStatus::SoftDisabled);
```
thebevrishot: butter to call `s.eof()`  instead.

ultimaweapon: The problems is it may not consistent because not all stream support it (e.g. network stream). Actually the better way is to throw a specific exception to indicated it EOF but right now STL did not have it so we have 2 choices:

1. Create a new exception class for that purpose.
2. Use what we already had.

The problems with the first choice is it need to fix all of stream to throw that exception, which will be incompatible with Bitcoin (e.g. when merging Bitcoin we may fail to apply our patch to a new type of stream). Also it will not compatible with STL stream too. So that why I think the second choice is the best one for now.

PR #530, opened 2019-07-22
Fix memory

a-bezrukov:


src/txmempool.cpp
```diff
@@ -752,8 +752,9 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list <CTransaction
         if (it != mapNextTx.end()) {
             const CTransaction &txConflict = *it->second;
             if (txConflict != tx) {
+                uint256 const txConflictHash = txConflict.GetHash();
```
ultimaweapon: Better to change to `uint256 const &txConflictHash` or revert it back to prevent copy constructor.

src/zerocoin_v3.cpp
```diff
@@ -750,8 +750,9 @@ void CSigmaState::Containers::RemoveMint(sigma::PublicCoin const & pubCoin) {
     mint_info_container::const_iterator iter = mintedPubCoins.find(pubCoin);
     if (iter != mintedPubCoins.end()) {
         mintMetaInfo[iter->second.coinGroupId][iter->second.denomination] -= 1;
+        CMintedCoinInfo tmpMintInfo(iter->second);
```
ultimaweapon: This will slower than the previous version due to it need to construct `CMintedCoinInfo`. Is there any reason why we do this?

src/zerocoin_v3.cpp
```diff
@@ -765,8 +766,9 @@ void CSigmaState::Containers::RemoveSpend(Scalar const & serial) {
     spend_info_container::const_iterator iter = usedCoinSerials.find(serial);
     if (iter != usedCoinSerials.end()) {
         spendMetaInfo[iter->second.coinGroupId][iter->second.denomination] -= 1;
+        CSpendCoinInfo tmpSpendInfo(iter->second);
```
ultimaweapon: Same here.

src/txmempool.cpp
psolstice: Nope, you cannot do the reference here. Also isn't this a bug in the bitcoin code?

src/zerocoin_v3.cpp
```diff
@@ -750,8 +750,9 @@ void CSigmaState::Containers::RemoveMint(sigma::PublicCoin const & pubCoin) {
     mint_info_container::const_iterator iter = mintedPubCoins.find(pubCoin);
     if (iter != mintedPubCoins.end()) {
         mintMetaInfo[iter->second.coinGroupId][iter->second.denomination] -= 1;
+        CMintedCoinInfo tmpMintInfo(iter->second);
         mintedPubCoins.erase(iter);
-        CheckSurgeCondition(iter->second.coinGroupId, iter->second.denomination);
+        CheckSurgeCondition(tmpMintInfo.coinGroupId, tmpMintInfo.denomination);
```
psolstice: If removal of a mint creates surge condition it would mean some problem other than surge (e.g. incorrect sigma state). Probably it deserves an assert. The only thing that can lead to the surge is new spend

src/zerocoin_v3.cpp
psolstice: @ultimaweapon the copy is done because erase call below can invalidate iterator

src/zerocoin_v3.cpp
```diff
@@ -765,8 +766,9 @@ void CSigmaState::Containers::RemoveSpend(Scalar const & serial) {
     spend_info_container::const_iterator iter = usedCoinSerials.find(serial);
     if (iter != usedCoinSerials.end()) {
         spendMetaInfo[iter->second.coinGroupId][iter->second.denomination] -= 1;
+        CSpendCoinInfo tmpSpendInfo(iter->second);
         usedCoinSerials.erase(iter);
-        CheckSurgeCondition(iter->second.coinGroupId, iter->second.denomination);
+        CheckSurgeCondition(tmpSpendInfo.coinGroupId, tmpSpendInfo.denomination);
```
psolstice: Removal of spend can never lead to surge condition

levonpetrosyan93: But it can remove the condition.

src/txmempool.cpp
ultimaweapon: What is the problems? `txConflict.GetHash()` already return `const uint256&`.

ultimaweapon: @a-bezrukov just explain what is going on here. I think will be better to put a comment here why we need to copy to prevent somebody accidentally change it in the future.

src/zerocoin_v3.cpp
psolstice: Point taken but just to prevent forks when surging

PR #536, opened 2019-07-23
Fix issue with parsing legacy HD keypath

riordant:
Fixes issue where old keypath format does not parse correctly in `importwallet`. Also adds information on hardened components to `CKeyMetadata`

### DIFF
--- a/src/wallet/walletdb.h
+++ b/src/wallet/walletdb.h
@@ -58,6 +58,9 @@ enum DBErrors
     DB_NEED_REWRITE
 };
 
+// {value, isHardened}
+typedef pair<int64_t,bool> Component;
+
 /* simple HD chain data model */
 class CHDChain
 {
@@ -107,8 +110,8 @@ public:
     int nVersion;
     int64_t nCreateTime; // 0 means unknown
     std::string hdKeypath; //optional HD/bip32 keypath
-    int64_t nChange; // HD/bip32 keypath change counter
-    int64_t nChild; // HD/bip32 keypath child counter
+    Component nChange; // HD/bip32 keypath change counter
+    Component nChild; // HD/bip32 keypath child counter
     CKeyID hdMasterKeyID; //id of the HD masterkey used to derive this key
 
     CKeyMetadata()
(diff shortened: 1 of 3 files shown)


src/wallet/walletdb.h
```diff
@@ -124,9 +125,11 @@ class CKeyMetadata
...
         if(hdKeypath=="m")
             return false;
-        boost::split(nComponents, hdKeypath, boost::is_any_of("/"), boost::token_compress_on);
-        nChange = boost::lexical_cast<int64_t>(nComponents[4]);
-        nChild = boost::lexical_cast<int64_t>(nComponents[5]);
+        std::string hdKeypathNonHardened = hdKeypath;
+        boost::erase_all(hdKeypathNonHardened, "'");
+        boost::split(nComponents, hdKeypathNonHardened, boost::is_any_of("/"), boost::token_compress_on);
+        nChange = boost::lexical_cast<int64_t>(nComponents[nComponents.size()-2]);
+        nChild = boost::lexical_cast<int64_t>(nComponents[nComponents.size()-1]);
```
thebevrishot: I think we should support or return false when we found hardened key instead of treat it as non-hardened.

riordant: Yes this is good suggestion, and I have actually extended this code to include a "Keypath" class, which will be needed when we add BIP39 support. However for now, `nChange` and `nChild` are only used by the BIP44 path, so this works without a big change.

thebevrishot: Then we should return false in hardened case. To let caller handle it for this release.

riordant: I don't understand what you mean. Do you mean adding `isHardened` books for `nChange` and `nChild`?

riordant: @thebevrishot Added information related to "hardened" values.

src/wallet/walletdb.h
```diff
@@ -58,6 +58,9 @@ enum DBErrors
     DB_NEED_REWRITE
 };
 
+// {value, isHardened}
+typedef pair<int64_t,bool> Component;
```
thebevrishot: Four bytes is enough. We can use uint32_t follow the function `CPubKey CWallet::GenerateNewKey(uint32_t nChange)`.

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys

riordant: Agree. Updated

src/wallet/walletdb.h
```diff
@@ -125,8 +129,17 @@ class CKeyMetadata
...
             return false;
         boost::split(nComponents, hdKeypath, boost::is_any_of("/"), boost::token_compress_on);
-        nChange = boost::lexical_cast<int64_t>(nComponents[4]);
-        nChild = boost::lexical_cast<int64_t>(nComponents[5]);
+        std::string nChangeStr = nComponents[nComponents.size()-2];
+        std::string nChildStr  = nComponents[nComponents.size()-1];
+
+        nChange.second = (nChangeStr.find("'") != string::npos);
+        boost::erase_all(nChangeStr, "'");
+        nChange.first = boost::lexical_cast<int64_t>(nChangeStr);
```
thebevrishot: sorry, I forgot to point here to be updated too.

src/wallet/walletdb.h
```diff
@@ -125,8 +129,17 @@ class CKeyMetadata
...
+
+        nChild.second = (nChildStr.find("'") != string::npos);
+        boost::erase_all(nChildStr, "'");
+        nChild.first = boost::lexical_cast<int64_t>(nChildStr);
```
thebevrishot: same issue

riordant: Correct, updated.

PR #545, opened 2019-07-26
Implement denomination creation for Exodus

### LINKED ISSUE #533: Implement denomination creation
This need to be a complete implementation (consensus level, wallet level, etc.). Store available denominations for each property in `CMPSPInfo`.

- [x] Implementation [6]
- [x] Review [1]

ultimaweapon:
## PR intention
Resolve #533


src/exodus/createpayload.cpp
```diff
@@ -549,5 +549,25 @@ std::vector<unsigned char> CreatePayload_ExodusAlert(uint16_t alertType, uint32_
     return payload;
 }
 
+std::vector<unsigned char> CreatePayload_CreateDenomination(uint32_t propertyId, uint64_t value)
+{
+    std::vector<unsigned char> payload;
+
+    uint16_t messageType = 1025;
```
thebevrishot: better to use `EXODUS_TYPE_CREATE_DENOMINATION` instead.

src/exodus/createpayload.h
```diff
@@ -39,5 +39,6 @@ std::vector<unsigned char> CreatePayload_MetaDExCancelEcosystem(uint8_t ecosyste
 std::vector<unsigned char> CreatePayload_ExodusAlert(uint16_t alertType, uint32_t expiryValue, const std::string& alertMessage);
 std::vector<unsigned char> CreatePayload_DeactivateFeature(uint16_t featureId);
 std::vector<unsigned char> CreatePayload_ActivateFeature(uint16_t featureId, uint32_t activationBlock, uint32_t minClientVersion);
+std::vector<unsigned char> CreatePayload_CreateDenomination(uint32_t propertyId, uint64_t value);
```
thebevrishot: I believe type of `value` should be int64_t instead because it's need to be compatible with `CAmount`.

src/exodus/rpctx.cpp
```diff
@@ -1481,6 +1481,53 @@ UniValue exodus_sendalert(const UniValue& params, bool fHelp)
...
+UniValue exodus_sendcreatedenomination(const UniValue& params, bool fHelp)
+{
+    if (fHelp || params.size() != 3) {
+        throw std::runtime_error(
+            "exodus_sendcreatedenomination \"fromaddress\" propertyid \"value\"\n"
+            "\nCreate a new denomination for the given property.\n"
+            "\nArguments:\n"
+            "1. fromaddress          (string, required) the address to send from\n"
+            "2. propertyid           (number, required) the property to create a new denomination\n"
+            "3. value                (string, required) the value of denomination to create\n"
```
thebevrishot: better to also support number.

src/exodus/tx.cpp
```diff
@@ -797,6 +800,26 @@ bool CMPTransaction::interpret_UnfreezeTokens()
...
 }
 
+/** Tx 1025 */
+bool CMPTransaction::interpret_CreateDenomination()
+{
+    if (pkt_size < 16) {
+        return false;
+    }
+
+    memcpy(&property, &pkt[4], 4);
```
thebevrishot: better to use `sizeof(property)` instead.

src/exodus/tx.cpp
```diff
@@ -797,6 +800,26 @@ bool CMPTransaction::interpret_UnfreezeTokens()
...
+    swapByteOrder32(property);
+    memcpy(&nValue, &pkt[8], 8);
```
thebevrishot: same issue

src/exodus/tx.cpp
```diff
@@ -2335,6 +2361,59 @@ int CMPTransaction::logicMath_UnfreezeTokens()
...
+        PrintToLog("%s(): rejected: type %d or version %d not permitted for property %d at block %d\n",
+                __func__,
+                type,
+                version,
+                property,
+                block);
+        return PKT_ERROR_TOKENS - 22;
+    }
+
+    if (!nValue || nValue > MAX_INT_8_BYTES) {
```
thebevrishot: if change nValue to int64_t this should be changed to `< 0`

src/exodus/createpayload.cpp
ultimaweapon: The problems is it will look ugly because other functions in the files hard-coded like this. But I agree we should not hard-coded like this so I will fix all of other functions.

src/exodus/createpayload.h
ultimaweapon: Same as previous issue. If we want to fix this one we should fix the other too, which need a lot of effort so I think we should fix it later.

src/exodus/rpctx.cpp
ultimaweapon: We can't due to it will have problems with JavaScript client.

src/exodus/tx.cpp
ultimaweapon: If we use this way we need to use it for `pkt[4]` too, which will loo ugly for the late member (e.g. `memcpy(&nValue, &pkt[sizeof(type) + sizeof(version) + sizeof(property)], sizeof(nValue));`). My idea is we should keep this way for now and find a better way and refactor it later.

src/exodus/tx.cpp
ultimaweapon: Same as previous comment.

src/exodus/tx.cpp
ultimaweapon: We can't do that for now due to it need to check a lot. Will be better to refactor it later.

src/exodus/createpayload.h
thebevrishot: agree

PR #550, opened 2019-07-26
Minor memory fix

a-bezrukov:
## PR intention
Fixes a minor memory leak.

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -117,11 +117,12 @@ static bool multiUserAuthorized(std::string strUserPass)
             std::string strSalt = vFields[1];
             std::string strHash = vFields[2];
 
-            unsigned int KEY_SIZE = 32;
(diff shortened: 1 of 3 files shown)


a-bezrukov:
> Take a fix from the bitcoin code instead
> https://github.com/bitcoin/bitcoin/blob/master/src/httprpc.cpp#L114

They have `static` keyword for the buffer. Our code does not have anything to prevent the race condition on the array. I understand why you requested this change but I don't know what is better for you in this situation - leave it as implemented (without unnecessary copying from char array to vector), or apply their code without the `static` keyword. Please let me know.

psolstice:
Look carefully at their code
```
        static const unsigned int KEY_SIZE = 32;
        unsigned char out[KEY_SIZE];
```
static is only for size (rightfully so though I'd prefer constexpr). Buffer is not static

PR #556, opened 2019-07-29
Enabling ZSTD

levonpetrosyan93:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1150,8 +1150,30 @@ if test x$need_bundled_univalue = xyes; then
   AC_CONFIG_SUBDIRS([src/univalue])
 fi
 
+PKG_CHECK_MODULES([ZSTD],
+                  [libzstd >= 1.1],
+                  have_zstd=yes,
+                  have_zstd=no)
+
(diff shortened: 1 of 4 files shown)


ultimaweapon:
We don't need this anymore. When I was created #248 I don't know release get build, sorry for not updated the issue. If we really want to enable ZSTD for Tor we need to add it to `depends/packages` too, which I don't think it is worth.

levonpetrosyan93:
@ultimaweapon In this PR ZSTD is being enabled when it is installed, I other cases it works as before.

FreakyMaryk:
@Putta, could you pls approve that we need this changes which Levon did? Or we can just close this and not add this changes?

reubenyap:
If this fixes when zstd is installed I don't see any harm in installing it.

ultimaweapon:
What is the point to have it on developer machines but not in release? Currently it already disable ZSTD so even there is ZSTD installed in developer machines it will not try to use it.

ultimaweapon:
@reubenyap the problems is it take some effort. If you think it worth to have it then we should include it with this PR.

PR #566, opened 2019-08-01
Add RPC function to regenerate mint pool if incorrect indexes used

riordant:
## PR intention
- FIxes an issue where wallets, generated with HD mint prior to 0.13.8.2, have mints that are always stuck in "pending" on the Sigma page. This is due to incorrect indexes being generated for the mint pool database functions in the 0.13.8.1 release. The fix is simply to regenerate the mint pool from the known values, and add with the correct indexes.

- Also adds increased logging level generally for HD mint code.

## Code changes brief
- Add RPC function `regeneratemintpool`
- Modification of `RegenerateMintPoolEntry` to return a uint256 pair - `<hashPubcoin, hashSerial>`


src/hdmint/wallet.cpp
```diff
@@ -142,12 +151,30 @@ bool CHDMintWallet::LoadMintPoolFromDB()
...
         mintPool.Add(mintPoolPair);
+    }
 
     return true;
 }
 
+bool CHDMintWallet::GetSerialForPubcoin(const std::vector<std::pair<uint256, GroupElement>>& serialPubcoinPairs, const uint256& hashPubcoin, uint256& hashSerial)
+{
+    bool fFound = false;
+    for(auto serialPubcoinPair : serialPubcoinPairs){
```
a-bezrukov: May it be a const ref, not a copy?

src/hdmint/wallet.cpp
```diff
@@ -142,12 +151,30 @@ bool CHDMintWallet::LoadMintPoolFromDB()
...
+        GroupElement pubcoin = serialPubcoinPair.second;
```
a-bezrukov: Same here

riordant: Agreed.

src/hdmint/wallet.cpp
riordant: Agreed

src/hdmint/wallet.cpp
riordant: Could you clarify what you mean here?

a-bezrukov: pubcoin is a copy now. I think making it ``` GroupElement const & pubcoin = serialPubcoinPair.second``` will be better performance-wise.

levonpetrosyan93: I think it can be like this  const GroupElement& pubcoin.

levonpetrosyan93: Will be more readable if we keep this line, only make it const reference.

PR #567, opened 2019-08-01
RPC Wallet tests fix

FreakyMaryk:
## PR intention
RPC Wallet tests fix for:
* wallet-dump
* walletbackup
* wallet-hd

## Code changes brief
Time of tests executing was increased due to tests long-run (restarting/sync/dampwallet)
There are some TODO in this tests which are can be removed during review.


ultimaweapon:
@FreakyMaryk could you provide the example of wallet dump file?

FreakyMaryk:
@ultimaweapon 
[wallet.unencrypted.dump.txt](https://github.com/zcoinofficial/zcoin/files/3460799/wallet.unencrypted.dump.txt)

thebevrishot:
@riordant `# - As wallet encryption creates a new master seed, it adds another set of hdmint keys.` Could new keys can be used? Is this a bug?

riordant:
@thebevrishot Not a bug. When you generate a new wallet, it is in unencrypted state, and has a particular master seed. When you encrypt, it generates a new master seed. Both seeds generate 21 keys to be used for mints. The newer keys will always be used, that is, the keys for latest master seed will always be used for new mints.

qa/rpc-tests/wallet-dump.py
```diff
@@ -96,12 +106,19 @@ def run_test (self):
         self.nodes[0].walletpassphrase('test', 10)
         # Should be a no-op:
         self.nodes[0].keypoolrefill()
-        self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.encrypted.dump")
+        try:
+            self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.encrypted.dump")
```
ultimaweapon: Same here.

qa/rpc-tests/wallet-hd.py
```diff
@@ -36,7 +37,11 @@ def run_test (self):
 
         # Import a non-HD private key in the HD wallet
         non_hd_add = self.nodes[0].getnewaddress()
-        self.nodes[1].importprivkey(self.nodes[0].dumpprivkey(non_hd_add))
+        try:
+            self.nodes[1].importprivkey(self.nodes[0].dumpprivkey(non_hd_add))
```
ultimaweapon: Same here.

qa/rpc-tests/walletbackup.py
```diff
@@ -125,11 +126,28 @@ def run_test(self):
         logging.info("Backing up")
         tmpdir = self.options.tmpdir
         self.nodes[0].backupwallet(tmpdir + "/node0/wallet.bak")
-        self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.dump")
+        
+        try:
+            self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.dump")
```
ultimaweapon: Same here.

qa/rpc-tests/walletbackup.py
```diff
@@ -125,11 +126,28 @@ def run_test(self):
...
+        except Exception as ex:
+            key = parse_tmp_dumpwallet_code_from_warning(ex.error['message'])
+            self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.dump", key)
+   
         self.nodes[1].backupwallet(tmpdir + "/node1/wallet.bak")
-        self.nodes[1].dumpwallet(tmpdir + "/node1/wallet.dump")
+        
+        try:
+            self.nodes[1].dumpwallet(tmpdir + "/node1/wallet.dump")
```
ultimaweapon: Same here.

qa/rpc-tests/walletbackup.py
```diff
@@ -125,11 +126,28 @@ def run_test(self):
...
+        except Exception as ex:
+            key = parse_tmp_dumpwallet_code_from_warning(ex.error['message'])
+            self.nodes[1].dumpwallet(tmpdir + "/node1/wallet.dump", key)
+
         self.nodes[2].backupwallet(tmpdir + "/node2/wallet.bak")
-        self.nodes[2].dumpwallet(tmpdir + "/node2/wallet.dump")
+        
+        try:
+            self.nodes[2].dumpwallet(tmpdir + "/node2/wallet.dump")
```
ultimaweapon: Same here.

qa/rpc-tests/wallet-hd.py
```diff
@@ -12,6 +12,7 @@
 )
 import os
 import shutil
+from test_framework.test_helper import *
```
thebevrishot: import issue

qa/rpc-tests/wallet-hd.py
```diff
@@ -72,7 +77,7 @@ def run_test (self):
         for _ in range(num_hd_adds):
             hd_add_2 = self.nodes[1].getnewaddress()
             hd_info_2 = self.nodes[1].validateaddress(hd_add_2)
-            assert_equal(hd_info_2["hdkeypath"], "m/0'/0'/"+str(_+1)+"'")
+            assert_equal(hd_info_2["hdkeypath"], "m/44'/1'/0'/0/"+str(_+1))
```
thebevrishot: space error should be `..." + str...`

qa/rpc-tests/walletbackup.py
```diff
@@ -35,6 +35,7 @@
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
+from test_framework.test_helper import *
```
thebevrishot: import issue

qa/rpc-tests/wallet-dump.py
```diff
@@ -25,8 +25,8 @@ def read_dump(file_name, addrs, hd_master_addr_old):
                 # key = key_label.split(" ")[0]
                 keytype = key_label.split(" ")[2]
                 if len(comment) > 1:
-                    addr_keypath = comment.split(" addr=")[1]
-                    addr = addr_keypath.split(" ")[0]
+                    addr_keypath = key_label.split(" ")[3].lower()
+                    addr = comment.split(" addr=")[1].split(" ")[0].rstrip()
```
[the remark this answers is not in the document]
FreakyMaryk: Regexp is greedy

qa/rpc-tests/wallet-dump.py
```diff
@@ -5,7 +5,7 @@
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (start_nodes, start_node, assert_equal, bitcoind_processes)
-
+from test_framework.test_helper import *
```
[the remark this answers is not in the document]
FreakyMaryk: Agree

qa/rpc-tests/wallet-hd.py
FreakyMaryk: You mean formatting

thebevrishot: yep

qa/rpc-tests/wallet-dump.py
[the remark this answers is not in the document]
thebevrishot: It's ok to use Regexp just to find by key. You should got the result exactly one otherwise it should fail. I think it'll be more easy to understand.

qa/rpc-tests/wallet-dump.py
```diff
@@ -20,35 +21,41 @@ def read_dump(file_name, addrs, hd_master_addr_old):
...
-                            found_addr += 1
-                            break
-                        elif keytype == "change=1":
-                            found_addr_chg += 1
-                            break
-                        elif keytype == "reserve=1":
-                            found_addr_rsv += 1
-                            break
+
+                lparts = re.findall(r"[a-zA-Z0-9\=\/\:\-']{2,}", line)
```
ultimaweapon: `[^#^\s]{2,}` will be more easy to read. Also IMO just `parts` instead of `lparts` look better.

qa/rpc-tests/wallet-dump.py
```diff
@@ -81,12 +92,19 @@ def run_test (self):
...
+            key = get_dumpwallet_otp (ex.error['message'])
+            self.nodes[0].dumpwallet(tmpdir + "/node0/wallet.unencrypted.dump", key)
+        assert key, 'Import wallet did not raise exception when was called first time without one-time code.'
 
         found_addr, found_addr_chg, found_addr_rsv, hd_master_addr_unenc = \
             read_dump(tmpdir + "/node0/wallet.unencrypted.dump", addrs, None)
         assert_equal(found_addr, test_addr_count)  # all keys must be in the dump
-        assert_equal(found_addr_chg, 50)  # 50 blocks where mined
+
+        assert_equal(found_addr_chg, 50 + hdmint_key_count) # 50 block were mined + hdmint keys
```
ultimaweapon: @riordant do you know why it treat HD mint key as a change key? This does not seem right to me. For me it should be `reserve=1` instead of `change=1` for HD mint key in the pool.

qa/rpc-tests/wallet-dump.py
```diff
@@ -96,12 +114,21 @@ def run_test (self):
...
+        assert key, 'Import wallet did not raise exception when was called first time without one-time code.'
 
         found_addr, found_addr_chg, found_addr_rsv, hd_master_addr_enc = \
             read_dump(tmpdir + "/node0/wallet.encrypted.dump", addrs, hd_master_addr_unenc)
         assert_equal(found_addr, test_addr_count)
-        assert_equal(found_addr_chg, 90 + 1 + 50)  # old reserve keys are marked as change now
+        
+        # - old reserve keys are marked as change now
+        # - As wallet encryption creates a new master seed, it adds another set of hdmint keys.
+        assert_equal(found_addr_chg, 90 + 1 + 50 + (hdmint_key_count * 2))
```
ultimaweapon: @riordant same question here.

riordant: @ultimaweapon what question is that? Your last comment seems unrelated

qa/rpc-tests/wallet-dump.py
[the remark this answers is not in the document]
thebevrishot: I expect something like `re.findall('(hdKeypath)(\=)(\w+[\/\'\d]*)', str)`,  `re.findall('(addr)(\=)(\w+)', str)`, and `re.findall('(inactivehdmaster|hdmaster)(\=)(\d)', str)`. current solution is strict with order of values but restoring logic isn't.

https://github.com/zcoinofficial/zcoin/pull/508/files#diff-522490d83dce5375d423b23886e4125eL476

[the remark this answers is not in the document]
FreakyMaryk: As I see it is same now. Such way of getting values will be test also format of wallet data instead of only searching values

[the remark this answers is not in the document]
FreakyMaryk: And also pls check Putta's comment below

qa/rpc-tests/wallet-dump.py
FreakyMaryk: Do not agree. Cause in my regexp you can see valid symbols in dumpwallet.  And this is type of data validation. At least I exclude using of !@$...
But your regexp will match all special symbols.

qa/rpc-tests/wallet-dump.py
riordant: HD mint keys are not added to the reserve pool, that is why they don't show as reserve keys. You are correct in that they probably shouldn't show `change` keys however. I could add a `hdmint=1` value for these?

qa/rpc-tests/wallet-dump.py
ultimaweapon: Yes that is what it should do because `importwallet` also splitting using space too (see here https://github.com/zcoinofficial/zcoin/blob/master/src/wallet/rpcdump.cpp#L461). If we follow the same format as `importwallet` that mean it will have more chance to catch bug when `dumpwallet` change it output format.

qa/rpc-tests/wallet-dump.py
ultimaweapon: I see. I think we can do it later. Already created an issue: https://github.com/zcoinofficial/zcoin/issues/602

qa/rpc-tests/wallet-dump.py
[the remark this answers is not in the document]
thebevrishot: I see

PR #587, opened 2019-08-06
Implement payload creation and parser for tokens minting

thebevrishot:
close https://github.com/zcoinofficial/zcoin/issues/583

### DIFF
--- a/src/exodus/tx.cpp
+++ b/src/exodus/tx.cpp
@@ -172,6 +172,9 @@ bool CMPTransaction::interpret_Transaction()
         case EXODUS_TYPE_CREATE_DENOMINATION:
             return interpret_CreateDenomination();
 
+        case EXODUS_TYPE_SIGMA_SIMPLE_MINT:
+            return interpret_SimpleMint();
+
         case EXODUS_MESSAGE_TYPE_DEACTIVATION:
             return interpret_Deactivation();
 
@@ -821,7 +824,7 @@ bool CMPTransaction::interpret_CreateDenomination()
 }
 
 /** Tx 1026 */
-bool CMPTransaction::interpret_CreateSimple()
+bool CMPTransaction::interpret_SimpleMint()
 {
     constexpr size_t exodusMintSize = 35;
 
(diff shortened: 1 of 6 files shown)


src/exodus/tx.h
```diff
@@ -216,6 +218,9 @@ class CMPTransaction
     unsigned int getIndexInBlock() const { return tx_idx; }
     uint32_t getDistributionProperty() const { return distribution_property; }
 
+    /** Sigma */
+    std::vector<std::pair<uint8_t, exodus::SigmaPublicKey>> getMints() const { return mints; }
```
ultimaweapon: We should return const reference instead.

src/exodus/tx.cpp
```diff
@@ -820,6 +820,44 @@ bool CMPTransaction::interpret_CreateDenomination()
...
+        SER_NETWORK, CLIENT_VERSION
+    );
+
+    for (auto &mint : mints) {
+        deserialized >> mint;
+    }
+
+    if ((!rpcOnly && exodus_debug_packets) || exodus_debug_packets_readonly) {
+        PrintToLog("\t        property: %d (%s)\n", property, strMPProperty(property));
+        PrintToLog("\t           mints: %d\n", mints.size());
```
ultimaweapon: I think we should also show all minted denomination too.

src/exodus/tx.h
```diff
@@ -315,6 +315,8 @@ class CMPTransaction
         if (block != other.block) return block > other.block;
         return tx_idx > other.tx_idx;
     }
+
+    boost::signals2::signal<void (CMPTransaction* const)> NotifyProcessedTransaction;
```
ultimaweapon: This is not right due to it is impossible for subscriber to subscribe all `CMPTransaction`. We should move this to `exodus` namespace and make it global.

src/exodus/tx.cpp
```diff
@@ -37,6 +37,11 @@ using boost::algorithm::token_compress_on;
 
 using namespace exodus;
 
+namespace exodus
+{
+boost::signals2::signal<void (CMPTransaction* const)> NotifyProcessedTransaction;
```
ultimaweapon: This is not right. What it mean is _A pointer to non-const CMPTransaction with argument as const_. It should be `CMPTransaction const *` instead.

src/exodus/tx.h
```diff
@@ -315,12 +315,14 @@ class CMPTransaction
...
-
-    boost::signals2::signal<void (CMPTransaction* const)> NotifyProcessedTransaction;
 };
 
 /** Parses a transaction and populates the CMPTransaction object. */
 int ParseTransaction(const CTransaction& tx, int nBlock, unsigned int idx, CMPTransaction& mptx, unsigned int nTime=0);
 
+namespace exodus
+{
+extern boost::signals2::signal<void (CMPTransaction* const)> NotifyProcessedTransaction;
```
ultimaweapon: Same here.

PR #588, opened 2019-08-06
Update Lookup Property tab to show sigma information

thebevrishot:
closes https://github.com/zcoinofficial/zcoin/issues/548

### DIFF
--- a/src/qt/lookupspdialog.cpp
+++ b/src/qt/lookupspdialog.cpp
@@ -424,10 +424,10 @@ void LookupSPDialog::matchingComboBoxChanged(int idx)
     updateDisplayedProperty();
 }
 
-class CustomWidgetItem : public QTableWidgetItem
+class NumericalCmpWidgetItem : public QTableWidgetItem
 {
 public:
-    CustomWidgetItem(const QString& text) : QTableWidgetItem(text)
+    NumericalCmpWidgetItem(const QString& text) : QTableWidgetItem(text)
     {
     }
 
@@ -441,8 +441,8 @@ void LookupSPDialog::addDenominationRow(uint8_t id, const std::string& value)
 {
     int workingRow = ui->denominationTable->rowCount();
     ui->denominationTable->insertRow(workingRow);
-    CustomWidgetItem *idCell = new CustomWidgetItem(QString::fromStdString(FormatIndivisibleMP(id)));
-    CustomWidgetItem *valuCell = new CustomWidgetItem(QString::fromStdString(value));
+    NumericalCmpWidgetItem *idCell = new NumericalCmpWidgetItem(QString::fromStdString(FormatIndivisibleMP(id)));
+    NumericalCmpWidgetItem *valuCell = new NumericalCmpWidgetItem(QString::fromStdString(value));
     idCell->setTextAlignment(Qt::AlignCenter);
     valuCell->setTextAlignment(Qt::AlignRight + Qt::AlignVCenter);
     ui->denominationTable->setItem(workingRow, 0, idCell);
--- a/src/qt/lookupspdialog.cpp
+++ b/src/qt/lookupspdialog.cpp
@@ -299,6 +299,9 @@ void LookupSPDialog::updateDisplayedProperty()
     {
        if (propertyId == 1) { tokenLabel = " EXODUS"; } else { tokenLabel = " TEXODUS"; }
     }
+
+    ui->sigmaStatusLabel->setText(QString::fromStdString(std::to_string(sp.sigmaStatus)));
+
     if (divisible) { strTotalTokens = FormatDivisibleMP(totalTokens); } else { strTotalTokens = FormatIndivisibleMP(totalTokens); }
     if (divisible) { strWalletTokens = FormatDivisibleMP(walletTokens); } else { strWalletTokens = FormatIndivisibleMP(walletTokens); }
     ui->totalTokensLabel->setText(QString::fromStdString(strTotalTokens + tokenLabel));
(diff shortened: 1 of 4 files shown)


thebevrishot:
qt-sigma-infomation

src/qt/lookupspdialog.cpp
```diff
@@ -59,6 +59,35 @@ LookupSPDialog::LookupSPDialog(QWidget *parent) :
...
+       ui->denominationTable->horizontalHeader()->setResizeMode(1, QHeaderView::Stretch);
+    #else
+       ui->denominationTable->horizontalHeader()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
+       ui->denominationTable->horizontalHeader()->setSectionResizeMode(1, QHeaderView::Stretch);
+    #endif
+    ui->denominationTable->setAlternatingRowColors(true);
+
+    // initial resizing
+    ui->denominationTable->resizeColumnToContents(0);
+    // borrowedColumnResizingFixer->stretchColumnWidth(0);
```
ultimaweapon: Why do we keep this?

src/qt/lookupspdialog.cpp
```diff
@@ -59,6 +59,35 @@ LookupSPDialog::LookupSPDialog(QWidget *parent) :
...
+    ui->denominationTable->verticalHeader()->setVisible(false);
+    ui->denominationTable->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    ui->denominationTable->setSelectionBehavior(QAbstractItemView::SelectRows);
+    ui->denominationTable->setSelectionMode(QAbstractItemView::SingleSelection);
+    ui->denominationTable->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOn);
+    ui->denominationTable->setTabKeyNavigation(false);
+    ui->denominationTable->setContextMenuPolicy(Qt::CustomContextMenu);
```
ultimaweapon: This required for custom context menu (right click menu). I don't think we need this.

src/qt/lookupspdialog.cpp
```diff
@@ -378,3 +423,28 @@ void LookupSPDialog::matchingComboBoxChanged(int idx)
...
+
+    bool operator<(const QTableWidgetItem& other) const
+    {
+        return text().toDouble() < other.text().toDouble();
+    }
+};
+
+void LookupSPDialog::addDenominationRow(uint8_t id, const std::string& value)
+{
+    int workingRow = ui->denominationTable->rowCount();
```
ultimaweapon: IMO rename to `index` will be more easy to understand.

src/qt/lookupspdialog.cpp
```diff
@@ -378,3 +423,28 @@ void LookupSPDialog::matchingComboBoxChanged(int idx)
...
+    ui->denominationTable->insertRow(workingRow);
+    NumericalCmpWidgetItem *idCell = new NumericalCmpWidgetItem(QString::fromStdString(FormatIndivisibleMP(id)));
+    NumericalCmpWidgetItem *valuCell = new NumericalCmpWidgetItem(QString::fromStdString(value));
+    idCell->setTextAlignment(Qt::AlignCenter);
+    valuCell->setTextAlignment(Qt::AlignRight + Qt::AlignVCenter);
```
ultimaweapon: This look weird. It should be `Qt::AlignRight | Qt::AlignVCenter` instead.

src/qt/lookupspdialog.cpp
```diff
@@ -378,3 +423,28 @@ void LookupSPDialog::matchingComboBoxChanged(int idx)
...
 }
+
+class NumericalCmpWidgetItem : public QTableWidgetItem
+{
+public:
+    NumericalCmpWidgetItem(const QString& text) : QTableWidgetItem(text)
+    {
+    }
+
+    bool operator<(const QTableWidgetItem& other) const
```
ultimaweapon: Better to apply `override` keyword here so it will know immediately it override base class method.

src/qt/lookupspdialog.cpp
```diff
@@ -378,3 +423,28 @@ void LookupSPDialog::matchingComboBoxChanged(int idx)
...
+    {
+        return text().toDouble() < other.text().toDouble();
+    }
+};
+
+void LookupSPDialog::addDenominationRow(uint8_t id, const std::string& value)
+{
+    int workingRow = ui->denominationTable->rowCount();
+    ui->denominationTable->insertRow(workingRow);
+    NumericalCmpWidgetItem *idCell = new NumericalCmpWidgetItem(QString::fromStdString(FormatIndivisibleMP(id)));
```
ultimaweapon: `FormatIndivisibleMP(id)` does not seem right.

src/qt/lookupspdialog.h
```diff
@@ -39,6 +41,9 @@ public Q_SLOTS:
     Ui::LookupSPDialog *ui;
     WalletModel *model;
 
+    GUIUtil::TableViewLastColumnResizingFixer *borrowedColumnResizingFixer;
+    void addDenominationRow(uint8_t id, const std::string& value);
```
ultimaweapon: We should put this one in a separated group instead of mixing with variable group.

src/qt/lookupspdialog.h
```diff
@@ -39,6 +41,9 @@ public Q_SLOTS:
     Ui::LookupSPDialog *ui;
     WalletModel *model;
 
+    GUIUtil::TableViewLastColumnResizingFixer *borrowedColumnResizingFixer;
```
ultimaweapon: Variable name hard to understand. IMO rename to `denominationTableLastColFixer` will be more better.

ultimaweapon: Not sure if we need this one due to I can't find where it is used.

src/qt/lookupspdialog.cpp
thebevrishot: removed

src/qt/lookupspdialog.cpp
thebevrishot: nice catch

PR #599, opened 2019-08-10
Sigma index fix

psolstice:
## PR intention
It's possible for the block that removed from the blockchain to be re-inserted back later. This PR fixes rare condition of not re-adding spends and mints into the index after such operation

## Code changes brief
In CSigmaState::RemoveBlock() clearing mints and spends from the block index should not be done


levonpetrosyan93:
But if we remove a block and connect another block anyway this to containers are being cleared in ConnectBlockSigma function.

riordant:
LGTM

PR #600, opened 2019-08-12
Change version in "commits since last version" badge

catchingknives:
## PR intention
Change version in "commits since last version" badge

## Code changes brief
Last PR was still based on 0.13.8.2. Now needs to be update with every release!


riordant:
As discussed, in order to avoid having to change the tag here for every release, it should point to `latest` rather than the explicit tag number. I have done this in the latest commit

PR #601, opened 2019-08-12
Dump wallet tests for Sigma and ZeroCoin

FreakyMaryk:
## PR intention
There were two tests added:
* check import and dump wallet works proper with sigma
* check import and dump wallet works proper with zerocoin


FreakyMaryk:
@riordant Could you pls take a look, cause now tests failing for some reason. I am making investigation, but it would be faster if you will take a look also.

qa/rpc-tests/wallet_dumpsigma.py
```diff
@@ -0,0 +1,84 @@
...
+        self.setup_clean_chain = False
+
+    def setup_nodes(self):
+        # This test requires mocktime
+        enable_mocktime()
+        return start_nodes(self.num_nodes, self.options.tmpdir)
+
+    def run_test(self):
+        getcontext().prec = 6
+        self.nodes[0].generate(401)
```
ultimaweapon: Test framework already generated 200 blocks for us when `setup_clean_chain` is `False` so we can reduce this number fo `201`.

qa/rpc-tests/wallet_dumpsigma.py
```diff
@@ -0,0 +1,84 @@
...
+        self.nodes[0].mint("{0:.2f}".format(3*denom_sum))
+        self.nodes[0].generate(6)
+
+        # spend, single confirmation
+        self.nodes[0].spendmany("", {'THAYjKnnCsN5xspnEcb1Ztvw4mSPBuwxzU': denom_sum})
+        self.nodes[0].generate(1)
+
+        #make unconfirmed mints and spends
+        #self.nodes[0].mint(denom_sum)
+        #self.nodes[0].spendmany("", {'THAYjKnnCsN5xspnEcb1Ztvw4mSPBuwxzU': denom_sum})
```
ultimaweapon: Why do we keep the above commented code?

qa/rpc-tests/wallet_dumpsigma.py
```diff
@@ -0,0 +1,84 @@
...
+        os.remove(self.options.tmpdir + "/node0/regtest/wallet.dat")
+        start_node(0, self.options.tmpdir)
+
+        self.nodes[0].importwallet(tmpdir + "/node0/wallet.unencrypted.dump")
+
+        exp_sigma_mints = self.nodes[0].listsigmamints()
+        exp_sigma_unspentmints = self.nodes[0].listunspentsigmamints(100)
+        exp_sigma_spendsigmas = self.nodes[0].listsigmaspends(100)
+
+        assert_equal(exp_sigma_unspentmints, exp_sigma_unspentmints)
```
ultimaweapon: One of operands is wrong.

qa/rpc-tests/wallet_dumpsigma.py
```diff
@@ -0,0 +1,84 @@
...
+        self.nodes[0].spendmany("", {'THAYjKnnCsN5xspnEcb1Ztvw4mSPBuwxzU': denom_sum})
+        self.nodes[0].generate(1)
+
+        #make unconfirmed mints and spends
+        #self.nodes[0].mint(denom_sum)
+        #self.nodes[0].spendmany("", {'THAYjKnnCsN5xspnEcb1Ztvw4mSPBuwxzU': denom_sum})
+
+        #get list of unspent mints and spends, mints
+        sigma_mints = self.nodes[0].listsigmamints()
+        sigma_unspentmints = self.nodes[0].listunspentsigmamints(100)
```
ultimaweapon: I think this one will give the empty `sigma_unspentmints` due to we want at least 100 confirmation but the mints we just generated had only 7 confirmations.

qa/rpc-tests/wallet_dumpzerocoin.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def setup_nodes(self):
+        # This test requires mocktime
+        enable_mocktime()
+        return start_nodes(self.num_nodes, self.options.tmpdir)
+
+    def run_test(self):
+        getcontext().prec = 6
+        self.nodes[0].generate(101)
+        self.sync_all()
```
ultimaweapon: We should not need the above 2 lines due to the same reason as before.

src/wallet/rpcdump.cpp
```diff
@@ -723,6 +748,30 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
...
+        file << strprintf("%s ", zerocoinEntry.denomination); // denomination
+        file << strprintf("%s ", zerocoinEntry.randomness.GetHex()); // randomness
+        file << strprintf("%s ", zerocoinEntry.serialNumber.GetHex()); // serialNumber
+        file << strprintf("%d ", zerocoinEntry.IsUsed); // IsUsed
+        file << strprintf("%s ", zerocoinEntry.nHeight); // nHeight
+        file << strprintf("%s ", zerocoinEntry.id); // id
+        if(!zerocoinEntry.ecdsaSecretKey.empty()){
+            uint256 ecdsaSecretKey;
+            std::copy(&zerocoinEntry.ecdsaSecretKey[0], &zerocoinEntry.ecdsaSecretKey[32], ecdsaSecretKey.begin());
+            file << strprintf("%s ", ecdsaSecretKey.GetHex()); // ecdsaSecretKey
```
ultimaweapon: There is `HexStr` function so we don't need to use this way.

src/wallet/rpcdump.cpp
```diff
@@ -723,6 +748,30 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
...
+
+    // Begin dump Zerocoins
+    list <CZerocoinEntry> listZerocoinEntries;
+    CWalletDB walletdb(pwalletMain->strWalletFile);
+    walletdb.ListPubCoin(listZerocoinEntries);
+
+    for (auto& zerocoinEntry : listZerocoinEntries) {
+        file << "zerocoin=1 ";
+        file << strprintf("%s ", zerocoinEntry.value.GetHex()); // value
+        file << strprintf("%s ", zerocoinEntry.denomination); // denomination
```
ultimaweapon: This does not seems right to use `%s` with denomination.

src/wallet/rpcdump.cpp
```diff
@@ -723,6 +748,30 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
...
+        file << strprintf("%s ", zerocoinEntry.randomness.GetHex()); // randomness
+        file << strprintf("%s ", zerocoinEntry.serialNumber.GetHex()); // serialNumber
+        file << strprintf("%d ", zerocoinEntry.IsUsed); // IsUsed
+        file << strprintf("%s ", zerocoinEntry.nHeight); // nHeight
```
ultimaweapon: Same here.

src/wallet/rpcdump.cpp
```diff
@@ -723,6 +748,30 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
...
+        file << strprintf("%s ", zerocoinEntry.id); // id
```
ultimaweapon: Same here.

src/wallet/rpcdump.cpp
```diff
@@ -462,67 +464,88 @@ UniValue importwallet(const UniValue& params, bool fHelp)
...
+            zerocoinEntry.denomination = stoi(vstr[2]);
+            zerocoinEntry.randomness.SetHex(vstr[3]);
+            zerocoinEntry.serialNumber.SetHex(vstr[4]);
+            zerocoinEntry.IsUsed = stoi(vstr[5]);
+            zerocoinEntry.nHeight = stoi(vstr[6]);
+            zerocoinEntry.id = stoi(vstr[7]);
+            if(vstr.size()>8){
+                uint256 ecdsaSecretKey;
+                ecdsaSecretKey.SetHex(vstr[8]);
+                zerocoinEntry.ecdsaSecretKey = std::vector<unsigned char>(ecdsaSecretKey.begin(),ecdsaSecretKey.end());
```
ultimaweapon: We need to update this once we change to use `HexStr`.

qa/rpc-tests/wallet_dumpsigma.py
riordant: Good point.

qa/rpc-tests/wallet_dumpsigma.py
riordant: No reason, it's an oversight. Will remove

qa/rpc-tests/wallet_dumpsigma.py
riordant: Correct. I have changed the number of confs to `1`.

qa/rpc-tests/wallet_dumpsigma.py
riordant: Correct

src/wallet/rpcdump.cpp
riordant: Can you clarify what you mean?

src/wallet/rpcdump.cpp
```diff
@@ -475,9 +475,7 @@ UniValue importwallet(const UniValue& params, bool fHelp)
             zerocoinEntry.nHeight = stoi(vstr[6]);
             zerocoinEntry.id = stoi(vstr[7]);
             if(vstr.size()>8){
-                uint256 ecdsaSecretKey;
-                ecdsaSecretKey.SetHex(vstr[8]);
-                zerocoinEntry.ecdsaSecretKey = std::vector<unsigned char>(ecdsaSecretKey.begin(),ecdsaSecretKey.end());
+                zerocoinEntry.ecdsaSecretKey = std::vector<unsigned char>(vstr[8].begin(),vstr[8].end());
```
ultimaweapon: This is not right. `vstr[8]` is a hex string, not binary. We need to convert it to binary first. We can use `ParseHex` function to convert hex string to binary.

PR #610, opened 2019-08-13
Showing fee when spending to yourself

levonpetrosyan93:


src/qt/transactionrecord.cpp
```diff
@@ -61,14 +68,14 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
-
+            
             if (mine) {
                 sub.involvesWatchAddress = mine & ISMINE_WATCH_ONLY;
                 if (ExtractDestination(txout.scriptPubKey, address) && IsMine(*wallet, address))
                 {
                     sub.type = TransactionRecord::SpendToSelf;
                     sub.address = CBitcoinAddress(address).ToString();
-                    sub.credit = txout.nValue;
+                    sub.credit = -(nTxFee / nOuts);
```
ultimaweapon: This will produce an incorrect value if we spend to our address and other address in the same transaction. My idea is we should follow normal transaction behavior. What it does for normal transaction is it will append fee to the first recipient. Take a look here for it logic https://github.com/zcoinofficial/zcoin/blob/master/src/qt/transactionrecord.cpp#L168

My idea for how to handle spend to yourself is:

- If all recipients are our then show only one entry with debit by fee.
- If there is at least one other recipient then we will add fee to the debit of that entry. For our addresses in this case just show as credit.

levonpetrosyan93: OK! Let's follow to your suggestions.

src/qt/transactionrecord.cpp
```diff
@@ -53,30 +53,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
-            isminetype mine = wallet->IsMine(txout);
+        }
 
+        if(isAllToMe){
             TransactionRecord sub(hash, nTime);
             CTxDestination address;
+            sub.involvesWatchAddress = ISMINE_WATCH_ONLY;
+            if (ExtractDestination(wtx.vout[0].scriptPubKey, address) && IsMine(*wallet, address)) {
+                sub.type = TransactionRecord::SpendToSelf;
+                sub.address = CBitcoinAddress(address).ToString();
```
ultimaweapon: I think we should not show only the first address. My idea is:

1. If there is one output then just show it address.
2. If there is more than one outputs. We have two choices here:
    1. Show all addresses separated by some symbol (e.g. `,`). For me this is the best choice.
    2. Just show something like `Multiple addresses`.

src/qt/transactionrecord.cpp
```diff
@@ -53,30 +53,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
+                sub.credit = -nTxFee;
```
ultimaweapon: I think this is not right. It should be `sub.debit = -nTxFee;` instead.

src/qt/transactionrecord.cpp
```diff
@@ -53,30 +53,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
+                TransactionRecord sub(hash, nTime);
+                CTxDestination address;
+
+                if (mine) {
+                    sub.involvesWatchAddress = mine & ISMINE_WATCH_ONLY;
+                    if (ExtractDestination(txout.scriptPubKey, address) && IsMine(*wallet, address)) {
+                        sub.type = TransactionRecord::SpendToSelf;
+                        sub.address = CBitcoinAddress(address).ToString();
+                        sub.credit = txout.nValue;
+                        parts.append(sub);
```
ultimaweapon: We forgot to set `sub.idx`.

src/qt/transactionrecord.cpp
```diff
@@ -53,30 +53,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
+                    ExtractDestination(txout.scriptPubKey, address);
+                    sub.type = TransactionRecord::SpendToAddress;
                     sub.address = CBitcoinAddress(address).ToString();
-                    sub.credit = txout.nValue;
+                    sub.debit = -txout.nValue;
+                    if (first) {
+                        sub.debit -= nTxFee;
+                        first = false;
+                    }
                     parts.append(sub);
```
ultimaweapon: Same here.

src/qt/transactionrecord.cpp
levonpetrosyan93: We were not setting it previously too.

src/qt/transactionrecord.cpp
levonpetrosyan93: Looks you are correct.

src/qt/transactionrecord.cpp
levonpetrosyan93: Agree

src/qt/transactionrecord.cpp
levonpetrosyan93: sub.credit = -nTxFee; this way in transaction list it is shown red, I think better to keep this;

src/qt/transactionrecord.cpp
```diff
@@ -53,30 +53,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
+                isAllToMe = false;
+                break;
             }
-            isminetype mine = wallet->IsMine(txout);
+        }
 
+        if(isAllToMe){
             TransactionRecord sub(hash, nTime);
             CTxDestination address;
+            sub.involvesWatchAddress = ISMINE_WATCH_ONLY;
```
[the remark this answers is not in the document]
levonpetrosyan93: fixed.

src/qt/transactionrecord.cpp
```diff
@@ -57,23 +57,35 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
...
+            } else if (!txout.scriptPubKey.IsSigmaMint()) {
+                CTxDestination address;
+                ExtractDestination(txout.scriptPubKey, address);
+                if (firstAddress) {
+                    addresses.append(CBitcoinAddress(address).ToString());
+                    firstAddress = false;
+                } else
+                    addresses.append(", " + CBitcoinAddress(address).ToString());
             }
+            if(mine == ISMINE_WATCH_ONLY)
```
ultimaweapon: This is not right. It should be `&` instead of `==`.

src/qt/transactionrecord.cpp
ultimaweapon: Yes. It is a bug from the previous code. Could you fix it with this PR? This PR already working on this part so we don't need another PR to fix it.

PR #616, opened 2019-08-15
Added remint parsing to getrawtransaction

a-bezrukov:
## PR intention
Adding remint transaction information to JSON output of getrawtransaction for explorers and the watchdog.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -32,6 +32,7 @@
 #include <boost/assign/list_of.hpp>
 #include <univalue.h>
 #include "sigma.h"
+#include "sigma/remint.h"
 
 using namespace std;
 
@@ -99,6 +100,19 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
 
             in.push_back(Pair("value", ValueFromAmount(spend->getIntDenomination())));
             in.push_back(Pair("valueSat", spend->getIntDenomination()));
+        } else if (txin.IsZerocoinRemint()){
+            std::shared_ptr<sigma::CoinRemintToV3>  remint;
+            try {
+                CDataStream serData(std::vector<unsigned char>(txin.scriptSig.begin()+1, txin.scriptSig.end()), SER_NETWORK, PROTOCOL_VERSION);
+                remint = std::make_shared<sigma::CoinRemintToV3>(serData);
+                in.push_back(Pair("publicCoinValue", remint->getPublicCoinValue().ToString(16)));
+            } catch (std::ios_base::failure &) {
+                throw JSONRPCError(RPC_DATABASE_ERROR, "An error occurred during processing the Sigma spend information");
+            }
+            fillStdFields(in, txin);
+
+            in.push_back(Pair("value", remint->getDenomination()));
+            in.push_back(Pair("valueSat", remint->getDenomination() * COIN));
         } else {
             in.push_back(Pair("txid", txin.prevout.hash.GetHex()));
             in.push_back(Pair("vout", (int64_t)txin.prevout.n));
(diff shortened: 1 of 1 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -99,6 +100,19 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
+            try {
+                CDataStream serData(std::vector<unsigned char>(txin.scriptSig.begin()+1, txin.scriptSig.end()), SER_NETWORK, PROTOCOL_VERSION);
+                remint = std::make_shared<sigma::CoinRemintToV3>(serData);
+                in.push_back(Pair("publicCoinValue", remint->getPublicCoinValue().ToString(16)));
+            } catch (std::ios_base::failure &) {
+                throw JSONRPCError(RPC_DATABASE_ERROR, "An error occurred during processing the Zerocoin to Sigma remint information");
+            }
+            fillStdFields(in, txin);
+
+            in.push_back(Pair("value", remint->getDenomination()));
```
ultimaweapon: We should format the value with `ValueFromAmount` so it will be the same format as the others.

src/rpc/rawtransaction.cpp
```diff
@@ -99,6 +100,19 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
 
             in.push_back(Pair("value", ValueFromAmount(spend->getIntDenomination())));
             in.push_back(Pair("valueSat", spend->getIntDenomination()));
+        } else if (txin.IsZerocoinRemint()){
+            std::shared_ptr<sigma::CoinRemintToV3>  remint;
```
ultimaweapon: Better to use `std::unique_ptr` instead due to it more lightweight since we don't need the functionalities of `std::shared_ptr`.

src/rpc/rawtransaction.cpp
a-bezrukov: Sorry, I won't as ValueFromAmount converts sats to coins. I am doing it in the opposite direction.

ultimaweapon: The problems with the current way is it will show without fractional while the others will show as fractional even it is a whole coin (e.g. 1 XZC on the others will show as `1.00000000` but this one will show just `1`).

src/rpc/rawtransaction.cpp
a-bezrukov: We don't have make_unique yet, so instead of `remint = std::make_shared<sigma::CoinRemintToV3>(serData);` I would need to have `remint = std::unique_ptr<sigma::CoinRemintToV3>(new sigma::CoinRemintToV3(serData));` which is ways less readable and more error prone. Also, the unique_ptr varian may work slower, as it has 2 memory allocations, instead of one, which make_shared has. More information is available at: https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/

src/rpc/rawtransaction.cpp
a-bezrukov: Aha, okey

src/rpc/rawtransaction.cpp
ultimaweapon: I see. You right. I totally forgot about `std::make_unique`. Agree with readability. Thanks.

BTW, actually `unique_ptr` will not allocated additional memory. What it does it just a container to do RAII.

src/rpc/rawtransaction.cpp
levonpetrosyan93: Done correct I think, remint->getDenomination() returns fractional value.

a-bezrukov: For me not printing the leading zeroes looks more correct as well, as it saves electrons both on producing the output and parsing it. But since we have this approach everywhere, I thought it is better to follow it, so I added...

PR #630, opened 2019-08-20
Prevent client to use the denomination with confirmations less than N blocks

thebevrishot:
closes https://github.com/zcoinofficial/zcoin/issues/619


thebevrishot:
wrong branch

PR #631, opened 2019-08-20
Prevent client to use denomination with confirmations less than N blocks

thebevrishot:
closes https://github.com/zcoinofficial/zcoin/issues/619


src/exodus/test/sp_tests.cpp
```diff
@@ -0,0 +1,102 @@
+#include "../sp.h"
+
+#include "test/test_bitcoin.h"
+#include "main.h"
+#include "miner.h"
+#include "wallet/wallet.h"
+
+#include <boost/test/unit_test.hpp>
+
+struct SpTestingSetup : public TestingSetup
```
ultimaweapon: There is already `TestChain100Setup` in `src/test/test_bitcoin.h` that provides block generation function. The problems with that class is it will generating 100 blocks immediately. My idea is we should move block generating function to `TestingSetup` and add one method to generate empty blocks there.

src/exodus/test/sp_tests.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    SpTestingSetup()
+        : TestingSetup(CBaseChainParams::REGTEST), spinfo(pathTemp / "MP_spinfo_test", false)
+    {
+    }
+
+    int GenerateEmptyBlock(size_t blocks)
```
ultimaweapon: If we follow my idea we don't need this method here.

src/exodus/rpctx.cpp
```diff
@@ -1570,6 +1575,18 @@ UniValue exodus_sendmint(const UniValue& params, bool fHelp)
...
             static_cast<size_t>(amount), static_cast<uint8_t>(denomId));
+
+        int confirmations;
+        try {
+            confirmations = _my_sps->getDenominationConfirmation(propertyId, denomId, minConfirms);
+        } catch (std::invalid_argument const &e) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, e.what());
+        }
+
+        if (confirmations != -1 && confirmations < minConfirms) {
```
ultimaweapon: We need to revise this condition if we change the behavior of `getDenominationConfirmation`.

src/exodus/sp.cpp
```diff
@@ -438,6 +439,60 @@ bool CMPSPInfo::getWatermark(uint256& watermark) const
...
+    std::string spPrevValueData;
+    if (pdb->Get(readoptions, spPrevKey, &spPrevValueData).IsNotFound()) {
+        return false;
+    }
+
+    CDataStream spPreValue(
+        spPrevValueData.data(),
+        spPrevValueData.data() + spPrevValueData.size(),
+        SER_DISK, CLIENT_VERSION
+    );
```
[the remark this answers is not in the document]
thebevrishot: `CDataStream` have no constructure for that way.

src/exodus/test/sp_tests.cpp
thebevrishot: hmm I'm ok to generate 100 blocks. it's not hurt performance much.

src/exodus/test/sp_tests.cpp
thebevrishot: still need this but simplify it  by calling exists function.

src/exodus/rpctx.cpp
thebevrishot: have changed, just throw if remaining block is not zero.

src/exodus/sp.cpp
```diff
@@ -438,6 +439,60 @@ bool CMPSPInfo::getWatermark(uint256& watermark) const
...
+bool CMPSPInfo::getPrevVersion(uint32_t propertyId, Entry &info) const
+{
+    CDataStream spPrevKeyData(SER_DISK, CLIENT_VERSION);
+    spPrevKeyData << 'b';
+    spPrevKeyData << info.update_block;
+    spPrevKeyData << propertyId;
+    leveldb::Slice spPrevKey(&spPrevKeyData[0], spPrevKeyData.size());
+
+    std::string spPrevValueData;
+    if (pdb->Get(readoptions, spPrevKey, &spPrevValueData).IsNotFound()) {
```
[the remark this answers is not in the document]
thebevrishot: nice catch. then I throw when found other error.

src/exodus/sp.cpp
```diff
@@ -438,6 +439,60 @@ bool CMPSPInfo::getWatermark(uint256& watermark) const
...
+        spPrevValueData.data(),
+        spPrevValueData.data() + spPrevValueData.size(),
+        SER_DISK, CLIENT_VERSION
+    );
+    spPreValue >> info;
+
+    return true;
+}
+
+int CMPSPInfo::getDenominationConfirmation(
```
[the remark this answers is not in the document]
thebevrishot: make sense

src/exodus/sp.cpp
```diff
@@ -438,6 +439,65 @@ bool CMPSPInfo::getWatermark(uint256& watermark) const
...
+    }
+
+    int targetBlock =
+        std::max(chainActive.Height() - target + 1, 0);
+    CBlockIndex *lastBlockHasDenomination = nullptr;
+
+    while (
+        denomination < info.denominations.size() &&
+        (lastBlockHasDenomination =
+            GetBlockIndex(info.update_block))->nHeight <= targetBlock) {
```
ultimaweapon: It should be `>` instead of `<=`.

src/exodus/sp.h
```diff
@@ -190,6 +190,11 @@ class CMPSPInfo : public CDBBase
     void setWatermark(const uint256& watermark);
     bool getWatermark(uint256& watermark) const;
 
+    bool getPrevVersion(uint32_t propertyId, Entry &info) const;
+
+    int getDenominationRemainingConfirmation(uint32_t propertyId, uint8_t denomination,
+        int target = INT32_MAX);
```
ultimaweapon: We should not specify default value for `target`.

src/exodus/test/sp_tests.cpp
```diff
@@ -0,0 +1,86 @@
...
+    BOOST_CHECK_EQUAL(
+        100,
+        db.getDenominationRemainingConfirmation(0, 0, 100)
+    );
+}
+
+BOOST_AUTO_TEST_CASE(exist_denomination)
+{
+
+    GenerateEmptyBlock(10);
```
ultimaweapon: After discussion we don't need this.

src/exodus/test/sp_tests.cpp
```diff
@@ -0,0 +1,86 @@
...
+    BOOST_CHECK_EQUAL(0, db.getDenominationRemainingConfirmation(property, 0, 1));
+
+    GenerateEmptyBlock(5);
+
+    // 6 confimations
+    BOOST_CHECK_EQUAL(0, db.getDenominationRemainingConfirmation(property, 0, 0));
+    BOOST_CHECK_EQUAL(1, db.getDenominationRemainingConfirmation(property, 0, 7));
+    BOOST_CHECK_EQUAL(0, db.getDenominationRemainingConfirmation(property, 0, 6));
+
+    // require 2 confirmations must return -1
```
ultimaweapon: The comment is not correct.

src/exodus/sp.cpp
thebevrishot: nice catch

thebevrishot:
@ultimaweapon resolved, please review.

Issue #559, opened 2019-07-30
Mnemonic Backup Word BIP39 support


reubenyap:
Zcoin has since 0.13 had HD key support.

With @riordant recent updates, Sigma updates are also now backed up using the same seed.

To build in BIP39 support direct into the QT wallet when making a new wallet.
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

Take a note on this though:
https://bitcoin.stackexchange.com/questions/87407/can-a-bip39-12-word-mnemonic-be-used-to-restore-a-wallet-through-bitcoin-cli

There are also various derivation paths. 
BIP32 BIP44 BIP49 BIP84 BIP141
so some research needs to be done on which has the most wide support.

A good summary is this.
https://iancoleman.io/bip39/?

reubenyap:
Solved by #641

PR #641, opened 2019-08-26
Implement BIP39 (Mnemonic code for generating deterministic keys)

levonpetrosyan93:
Fixes https://github.com/zcoinofficial/zcoin/issues/732
Implementation of #559 

https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

High level summary:
- 12 and 24 words supported
- passphrase supported but optional.
- By default, new wallets will be created with mnemonic. Users can choose to disable a mnemonic wallet with `usemnemonic=0`.
- In `zcoind` a 24 word mnemonic with no passphrase is created by default. This is also the default setting at the Qt screen
- Qt users are shown the mnemonic page on startup, if:
    - new wallet being created
    - `usemnemonic=1` (default. meaning the page is not shown if `usemnemonic=0` passed)
- `importwallet` and `importprivkey` block with a mnemonic enabled wallet. A suggestion is given to create a non-mnemonic wallet if import is need. The purpose of this is to have a single mnemonic backup that is not complicated by other imports. (The keys would still import into the wallet if this check was removed, but presumably whoever was to do this would be aware of the repercussions).

- conf setting interactions are handled

- This PR also removes the code that creates a new Master key following encryption, either for a mnemonic wallet or older BIP32/44 wallets. This feature was introduced by us when Zcoin launched (ie. it was not a feature of Bitcoin 0.13 or any other version ); it provides no extra security benefit and only complicates the wallet code. Warning messages have been updated to reflect this change.


src/wallet/wallet.cpp
```diff
@@ -1377,6 +1387,36 @@ CPubKey CWallet::GenerateNewHDMasterKey() {
...
+    else {
+        LogPrintf("CWallet::GenerateNewHDChain -- Generating new HD Chain\n");
+
+        std::string mnemonic = GetArg("-mnemonic", "");
+        std::string mnemonicPassphrase = GetArg("-mnemonicpassphrase", "");
+
+        SecureString secureMnemonic(mnemonic.begin(), mnemonic.end());
+        SecureString securePassphrase(mnemonicPassphrase.begin(), mnemonicPassphrase.end());
+
+        if (!newHdChain.SetMnemonic(secureMnemonic, securePassphrase, true))
```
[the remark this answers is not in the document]
levonpetrosyan93: Yes It is called, and in than case new mnemonic is being generated.

src/wallet/hdchain.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    SecureString mnemonicNew = mnemonic_;
+    if(setMasterKeyID) {
+        if(mnemonicNew.empty())
+            mnemonicNew = Mnemonic::mnemonic_generate(256);
+        if(!Mnemonic::mnemonic_check(mnemonicNew))
+            throw std::runtime_error(std::string(__func__) + ": mnemonic is invalid");
+        Mnemonic::mnemonic_to_seed(mnemonicNew, passPhrase_, seed);
+
+        masterKeyID = CKeyID(Hash160(seed.begin(), seed.end()));
```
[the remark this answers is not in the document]
levonpetrosyan93: Did a research on this, I think we need to somehow use 256bit for mnemonic and keep 160 for old versions, but need to so more investigation how make this compatible with HW wallets.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
levonpetrosyan93: Now when -mnemonic is not set, that means user did not provided memonic words for recovery, than new mnemonic is not generated by default. I mean in that case new wallet es being created. I we want to allow choose between legacy seed creation or mnemonic, than we can add another argument for that.

src/wallet/wallet.cpp
```diff
@@ -1377,6 +1387,36 @@ CPubKey CWallet::GenerateNewHDMasterKey() {
     return pubkey;
 }
 
+void CWallet::GenerateNewHDChain(){
+    CHDChain newHdChain;
+
+    std::string strSeed = GetArg("-hdseed", "not hex");
```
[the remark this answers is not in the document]
levonpetrosyan93: I looked several other places, they did same way. I think better to have this also to allow users recover hd wallets crated before mnemonic version.

src/wallet/rpcdump.cpp
```diff
@@ -702,8 +702,36 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
     file << "\n";
 
     // add the base58check encoded extended master if the wallet uses HD
-    CKeyID masterKeyID = pwalletMain->GetHDChain().masterKeyID;
-    if (!masterKeyID.IsNull())
+    CHDChain chain = pwalletMain->GetHDChain();
+    CKeyID masterKeyID = chain.masterKeyID;
+    if(!chain.IsNull())
```
riordant: @levonpetrosyan93 I think there should also be a condition check here for HDChain version >= `VERSION_WITH_BIP39`

src/wallet/wallet.cpp
```diff
@@ -698,11 +702,12 @@ bool CWallet::EncryptWallet(const SecureString &strWalletPassphrase) {
             pwalletdbEncryption = NULL;
         }
 
-        Lock();
-        Unlock(strWalletPassphrase, true);
-
         // if we are using HD, replace the HD master key (seed) with a new one
-        if (!hdChain.masterKeyID.IsNull()) {
+        if(!hdChain.IsNull()) {
```
riordant: Same here: `hdChain.nVersion >= CHDChain::VERSION_WITH_BIP39`

src/wallet/wallet.cpp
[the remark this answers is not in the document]
riordant: I agree with the current implementation, and don't think we should use legacy wallets with mnemonic wallet type anymore

src/wallet/rpcwallet.cpp
```diff
@@ -4037,6 +4037,7 @@ extern UniValue importprivkey(const UniValue& params, bool fHelp);
 extern UniValue importaddress(const UniValue& params, bool fHelp);
 extern UniValue importpubkey(const UniValue& params, bool fHelp);
 extern UniValue dumpwallet_zcoin(const UniValue& params, bool fHelp);
+extern UniValue dumpmnemonic(const UniValue& params, bool fHelp);
```
riordant: I also think an `importmnemonic` function is needed. Make sure to check for wallet version

levonpetrosyan93: I think better to allow  inputting mnemonic only during first startup.

src/wallet/rpcdump.cpp
levonpetrosyan93: Good catch. Thank you.

src/wallet/hdchain.cpp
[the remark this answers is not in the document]
riordant: @psolstice I believe this can be marked resolved?

src/wallet/wallet.cpp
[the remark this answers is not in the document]
riordant: @psolstice Also this? The idea is that new wallets will always generate a mnemonic

qashban4:
@riordant @ultimaweapon can you review so we can merge.

ultimaweapon:
Nope I can't. Don't have time to review this due to need to focus on Exodus.

qashban4:
Review to be complete by 2019-09-26 for further merge.

qashban4:
@levonpetrosyan93 to test it with HW wallet during W40
@psolstice to review and test on HW wallet during W40

qashban4:
@psolstice update your review and test result on HW wallet

qashban4:
@levonpetrosyan93 to finalize UI 
Once levon tests it with hw key and tells test procedure 
@psolstice needs less than a day to test and finalize review.
Aim to close it during middle of W42.

qashban4:
@psolstice lets finalize this week (W42) so we can merge it to master.

qashban4:
Instructions for testing mnemonic with hardware ui
1. Create a new wallet
2. Open zcoin_qt -> Dubug window -> Console
3. type "dumpmnemonic" command
4. copy mnemonic words
5. Open https://wallet.trezor.io
6. Wipe device if needed and reconnect it
7. Click on "Recover Wallet"
8. Input words (it will ask by random order)
9. Once wallet is recovered, create addresses and compare to  addresses in zcoin_qt

qashban4:
@riordant @psolstice lets review and conclude to merge to master Thanks.

src/qt/notifymnemonic.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+void NotifyMnemonic::notify()
+{
+    NotifyMnemonic notify;
+    SecureString mnemonic;
+    SecureString passphrase;
+    pwalletMain->GetHDChain().GetMnemonic(mnemonic, passphrase);
+    notify.ui->mnemonic->setText(mnemonic.c_str());
+    if(notify.exec())
```
riordant: This call is causing a failure (on Mac at least). 
When creating a new wallet, this function is called. It looks like the `exec` function doesn't exist.
Output:
```
************************
UNKNOWN EXCEPTION       
zcoin in Runaway exception  
```

GUI says this:
`A fatal error occurred. Zcoin can no longer continue safely and will quit.`

src/wallet/hdchain.h
```diff
@@ -0,0 +1,70 @@
+#ifndef ZCOIN_HDCHAIN_H
+#define ZCOIN_HDCHAIN_H
+
+#include "key.h"
+
+/* simple HD chain data model */
+class CHDChain
```
psolstice: You absolutely cannot move CHDChain to another file because it's bitcoin code. It will create a lot of problems when upgrading to newer bitcoin cores

src/wallet/hdchain.h
```diff
@@ -0,0 +1,70 @@
...
+            READWRITE(nExternalChainCounters);
+            nExternalChainCounters.resize(N_CHANGES);
+        }
+        if(this->nVersion >= VERSION_WITH_BIP39) {
+            READWRITE(mnemonic);
+            READWRITE(passPhrase);
+            READWRITE(seed);
+            READWRITE(fIsCrypted);
+            READWRITE(f12Words);
+        }
```
psolstice: Why do we save/restore mnemonic, passPhrase, seed etc? Do we really really store password to the wallet in plain text next to it?

src/wallet/hdchain.h
levonpetrosyan93: In Dash it is done the same way.

src/wallet/hdchain.h
levonpetrosyan93: we save it to be able to allow user to dump it any time.

psolstice: no! we might print it *once* during new wallet initialization. Sorry to overlook it earlier but there shouldn't be postfactum dumpmnemonic command at all. Look how it's done when setting up Trezor one for example.
The way you do it effectively ruins every wallet security. Only flags/counters/key ids can be saved

src/wallet/hdchain.h
psolstice: not a good example because they implemented HD stuff to 0.12 version of their bitcoin core, we're already at 0.13 and should not mess with that

src/wallet/hdchain.h
psolstice: I saw that's how dash is doing it but that's definitely not how it should be done

src/wallet/hdchain.h
levonpetrosyan93: ok, good, will move it back.

src/wallet/hdchain.h
psolstice: Got to the bottom of it. Dash get its own implementation of `CHDChain` and it's encrypted/decrypted when saved into the `wallet.db`. Your implementation is a mix between that of bitcoin and dash. And that means all of these fields are saved to the wallet.db **unencrypted**. Which means a huge security hole

levonpetrosyan93: It is being encrypted when you encrypt the wallet, it is being encrypted in our case also.

src/qt/notifymnemonic.cpp
levonpetrosyan93: fixed

src/wallet/hdchain.h
psolstice: nope, only keys are encrypted, everything you write to the wallet with `Write` is not

src/wallet/mnemoniccontainer.h
```diff
@@ -0,0 +1,47 @@
...
+    bool f12Words;
+
+    MnemonicConatiner() { SetNull(); }
+
+    ADD_SERIALIZE_METHODS;
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion)
+    {
+            READWRITE(mnemonic);
+            READWRITE(seed);
```
psolstice: Why do we need to save seed? The only purpose of `MnemonicContainer` is to save mnemonic for later recovery by user. Seed is saved within `wallet.dat` by bitcoin code, no need to duplicate it here.

src/wallet/wallet.cpp
```diff
@@ -1462,6 +1522,89 @@ bool CWallet::SetHDChain(const CHDChain &chain, bool memonly) {
...
+        if (!mnemonicConatiner.SetMnemonic(secureCryptedMnemonic, secureCryptedPassphrase, false))
+            return false;
+    }
+
+    mnemonicConatiner.SetCrypted(true);
+
+    return true;
+}
+
+bool CWallet::DecryptMnemonicConatiner(MnemonicConatiner& mnConatiner) const
```
psolstice: Typo. You need to replace Conatiner with Container all over the code

src/wallet/wallet.cpp
```diff
@@ -677,10 +689,10 @@ bool CWallet::EncryptWallet(const SecureString &strWalletPassphrase) {
     if (IsCrypted())
         return false;
 
-    CKeyingMaterial vMasterKey;
```
psolstice: I am strongly against renaming like that. It makes any core upgrade a lot harder. If there is a conflict with another variable introduced in this PR that variable needs renaming, not the original one

src/wallet/wallet.cpp
```diff
@@ -1462,6 +1522,89 @@ bool CWallet::SetHDChain(const CHDChain &chain, bool memonly) {
...
+    if (mnemonicConatiner.GetMnemonic(cryptedMnemonic)) {
+        SecureVector vectorMnemonic;
+
+        std::vector<unsigned char> CryptedMnemonic(cryptedMnemonic.begin(), cryptedMnemonic.end());
+        if (!CryptedMnemonic.empty() && !DecryptSecret(vMasterKey, CryptedMnemonic, id, vectorMnemonic))
+            return false;
+
+        SecureString mnemonic(vectorMnemonic.begin(), vectorMnemonic.end());
+        SecureString passphrase;
+        if (!mnConatiner.SetMnemonic(mnemonic, passphrase, false))
```
psolstice: Passing empty passphrase isn't looking good or obvious. There is no need to derive seed ad this point (the only goal of using `DecryptMnemonicContainer` is to display mnemonic to the user, password is something only the user knows about). Or at least do it without adding new local variable

src/wallet/wallet.cpp
```diff
@@ -1462,6 +1522,89 @@ bool CWallet::SetHDChain(const CHDChain &chain, bool memonly) {
...
+            return false;
+    }
+
+    mnConatiner.SetCrypted(false);
```
psolstice: Does it remain unencrypted in memory until client quits? Can we use the same time-based mechanism to remove it from memory when the wallet locks on timeout? Or after there is no need in reading mnemonic anymore?

levonpetrosyan93: It is being deleleted, when going out of scope.

src/wallet/wallet.cpp
levonpetrosyan93: when we pass false it only set's mnemonic, but may be better to create seperate function for that. Will do tat way.

src/wallet/wallet.cpp
levonpetrosyan93: thank you, will replace.

src/wallet/wallet.cpp
levonpetrosyan93: ok will try to find a better solution.

src/wallet/mnemoniccontainer.h
levonpetrosyan93: It only being kept here.

psolstice: Keeping secret data when unnecessary is generally not a good idea. And seed is essentially ultimate attacker's goal

levonpetrosyan93: I got your point but we are keeping seed here for deriving keys, please take a look on GenerateNewKey() and GetKeyFromKeypath() functions.

src/wallet/crypter.cpp
```diff
@@ -113,7 +113,7 @@ bool CCrypter::Decrypt(const std::vector<unsigned char>& vchCiphertext, CKeyingM
 }
 
 
-static bool EncryptSecret(const CKeyingMaterial& vMasterKey, const CKeyingMaterial &vchPlaintext, const uint256& nIV, std::vector<unsigned char> &vchCiphertext)
+bool CCryptoKeyStore::EncryptSecret(const CKeyingMaterial& vMasterKey, const CKeyingMaterial &vchPlaintext, const uint256& nIV, std::vector<unsigned char> &vchCiphertext)
```
psolstice: Sorry for not spotting this earlier. Modifying bitcoin code like that is not something we'd want to do. On the other hand adding another method to the `CCryptoKeyStore` is ok. This way you can even revert the change that makes `vMasterKey` protected instead of private

src/wallet/mnemoniccontainer.h
psolstice: I still don't get it. When initializing wallet with mnemonic we generate new seed and from that moment it is in the hands of bitcoin code. Why do we need to do our own way of generating new key? Why can't bitcoin code generate one for us?

src/wallet/crypter.cpp
levonpetrosyan93: Ok, will do that way. Thank you.

src/wallet/mnemoniccontainer.h
levonpetrosyan93: Mnemonic seed is being generated 64 byte, but Key we can set only 32, and using other hacks is generating addresses not compatible to hardware wallets.

src/wallet/wallet.cpp
```diff
@@ -752,7 +764,11 @@ bool CWallet::EncryptWallet(const SecureString &strWalletPassphrase) {
         Unlock(strWalletPassphrase, true);
 
         // if we are using HD, replace the HD master key (seed) with a new one
-        if (!hdChain.masterKeyID.IsNull()) {
+        if(!mnemonicContainer.IsNull() && hdChain.nVersion >= CHDChain::VERSION_WITH_BIP39) {
+            assert(EncryptMnemonicContainer(vMasterKey));
+            SetMinVersion(FEATURE_HD);
+            assert(SetMnemonicContainer(mnemonicContainer, false));
+        } else if (!hdChain.masterKeyID.IsNull()) {
```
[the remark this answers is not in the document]
riordant: Resolved following merge of my branch.

src/wallet/wallet.cpp
```diff
@@ -1427,6 +1449,44 @@ CPubKey CWallet::GenerateNewHDMasterKey() {
...
+    bool isHDSeedSet = strSeed != "not hex";
+
+    if(isHDSeedSet && IsHex(strSeed)) {
+        std::vector<unsigned char> seed = ParseHex(strSeed);
+        if (!mnContainer.SetSeed(SecureVector(seed.begin(), seed.end())))
+            throw std::runtime_error(std::string(__func__) + ": SetSeed failed");
+        newHdChain.masterKeyID = CKeyID(Hash160(seed.begin(), seed.end()));
+    }
+    else {
+        LogPrintf("CWallet::GenerateNewHDChain -- Generating new MnemonicContainer\n");
```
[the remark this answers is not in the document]
riordant: Resolved following merge of my branch.

PR #661, opened 2019-09-04
Fix segmentation fault when data is too large for class B

### LINKED ISSUE #656: Coredump when try to do many mint in a single transaction.
crash when mint 254 coins in a single transaction.

- [x] Implementation [2]
- [x] Review [0.5]

ultimaweapon:
Resolve #656.


src/exodus/encoding.h
```diff
@@ -1,21 +1,150 @@
...
+
+    std::array<unsigned char, 33> data = {}; // Zeroes all bytes.
+
+    // Write sequence and payload.
+    data[0] = 0x02; // Public key prefix.
+    data[1] = sequence;
+    std::copy(begin, end, data.begin() + 2);
+
+    // Obfuscation sequence and payload.
+    for (int i = 1; i < 32; i++) {
```
thebevrishot: use `PACKET_SIZE + 1` instead. We also need to rename it.

src/exodus/encoding.h
```diff
@@ -1,21 +1,150 @@
...
-#include <utility>
 #include <vector>
 
-/** Embedds a payload in obfuscated multisig outputs, and adds an Exodus marker output. */
-bool Exodus_Encode_ClassB(const std::string& senderAddress, const CPubKey& redeemingPubKey, const std::vector<unsigned char>& vchPayload, std::vector<std::pair<CScript, int64_t> >& vecOutputs);
+#include <stddef.h>
+
+namespace exodus {
+
+constexpr unsigned MAX_PAYLOAD_PER_KEY = 30;
```
thebevrishot: Per previous comment, better change this to `PACKET_SIZE - 1`

src/exodus/encoding.h
```diff
@@ -1,21 +1,150 @@
...
+        // Get number of keys required for payload.
+        // At most 3 keys per output is allowed.
+        int generatingKeys = (remaining > MAX_PAYLOAD_PER_KEY) ? 2 : 1;
+
+        // Put packet in the keys.
+        std::vector<CPubKey> keys = { redeemingKey }; // Always include the redeeming pubkey
+
+        for (int i = 0; i < generatingKeys; i++, sequence++) {
+            if (sequence >= (MAX_SHA256_OBFUSCATION_TIMES + 1)) {
+                throw std::runtime_error("packet too large");
```
thebevrishot: `std::invalid_argument` is more descriptive.

PR #664, opened 2019-09-05
Exodus HD mint

thebevrishot:
closes https://github.com/zcoinofficial/zcoin/issues/627


src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+#include "../../hdmint/mintpool.h"
+
+#include "hdmint.h"
+
+namespace exodus
+{
+
+class HDMintWallet;
+
+struct MintMeta
```
ultimaweapon: After take a look in the code I think we should remove this. The reasons are:

1. It need to have duplicated fields with `HDMint`.
2. It just act like the lookup table to `HDMint`.

Instead of having a single map we can do multiple-map. With this way it will more faster.

src/exodus/exodus.h
```diff
@@ -7,11 +7,11 @@ class CCoinsView;
...
-#include "exodus/log.h"
-#include "exodus/persistence.h"
-#include "exodus/tally.h"
-#include "exodus/sigma.h"
-#include "exodus/sigmadb.h"
+#include "log.h"
+#include "persistence.h"
+#include "tally.h"
+#include "sigma.h"
+#include "sigmadb.h"
```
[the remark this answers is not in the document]
thebevrishot: fixed by another PR.

src/exodus/sigmadb.h
```diff
@@ -7,6 +7,7 @@
 #include "sigma.h"
 
 #include <univalue.h>
+#include <uint256.h>
```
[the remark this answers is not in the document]
thebevrishot: fixed

src/exodus/sigmadb.cpp
```diff
@@ -294,15 +314,23 @@ std::pair<MintGroupId, MintGroupIndex> CMPMintList::RecordMint(
...
     auto keyData = CreateSpendSerialKey(propertyId, denomination, serialData);
-    auto status = pdb->Put(writeoptions, GetSlice(keyData), leveldb::Slice());
+    auto status = pdb->Put(writeoptions, GetSlice(keyData),
+        leveldb::Slice(std::string(spendTx.begin(), spendTx.end())));
+
     if (!status.ok()) {
         throw std::runtime_error("record serial fail");
     }
 
+    SpendAdded(propertyId, denomination, serial, spendTx);
```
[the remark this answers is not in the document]
thebevrishot: make sense

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+    uint256 pubCoinValueHash;
+};
+
+class HDMintTracker
+{
+private:
+    bool initialized;
+    std::string walletFile;
+    HDMintWallet *mintWallet;
+    std::map<uint256, MintMeta> mapSerialHashes;
```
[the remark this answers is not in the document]
thebevrishot: change to HDMint instead.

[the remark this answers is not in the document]
thebevrishot: class `HDMint` store only mint data don't store data to spend such as serial, randomness and private key. I'm not sure it's good idea if we add serial.

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+
+public:
+    HDMintTracker(std::string walletFile, HDMintWallet *mintWallet);
+    void Add(const HDMint& dMint, bool isNew = false);
+    bool HasPubcoinHash(const uint256& hashPubcoin) const;
+    bool HasSerialHash(const uint256& hashSerial) const;
+    bool IsEmpty() const { return mapSerialHashes.empty(); }
+    void Init();
```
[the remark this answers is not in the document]
thebevrishot: removed

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+
+    bool GetMetaFromSerial(const uint256& hashSerial, MintMeta& meta);
+    bool GetMetaFromPubcoin(const uint256& hashPubcoin, MintMeta& meta);
+    std::vector<uint256> GetSerialHashes();
```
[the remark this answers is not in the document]
thebevrishot: removed

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+    std::vector<MintMeta> GetMints(bool confirmedOnly, bool inactive) const;
```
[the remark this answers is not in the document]
thebevrishot: renamed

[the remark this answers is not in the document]
thebevrishot: renamed

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+
+    std::vector<SigmaMint> ListMints(bool unusedOnly = true, bool matureOnly = true);
+    std::vector<MintMeta> ListMetas(bool unusedOnly, bool matureOnly, bool load);
```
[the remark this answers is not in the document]
thebevrishot: removed

src/exodus/hdmint/tracker.h
```diff
@@ -0,0 +1,71 @@
...
+    void SetPubCoinValue(GroupElement const &other);
+    uint256 const &GetPubCoinValueHash() const;
+    bool IsUsed() { return !spendTx.IsNull(); }
+
+private:
+    GroupElement pubCoinValue;
+    uint256 pubCoinValueHash;
+};
+
+class HDMintTracker
```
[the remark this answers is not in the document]
thebevrishot: agree. but I think `MintTracker` is enough.

src/exodus/hdmint/tracker.h
thebevrishot: good idea. Then I change it to `boost::multi_index` instead.

src/exodus/walletmodels.h
```diff
@@ -83,31 +77,42 @@ class SigmaMintId
 class SigmaMint
 {
 public:
-    PropertyId property;
-    DenominationId denomination;
+    SigmaMintId id;
```
ultimaweapon: This will increase a chance to have error for key mismatch due to `SigmaMintId` contains public key. We should generate it only when we really need it.

src/exodus/walletmodels.h
```diff
@@ -83,31 +77,42 @@ class SigmaMintId
...
+
+    int32_t count;
```
ultimaweapon: IMO `count` is quite confusing. My idea is rename to `index` looks better.

ultimaweapon: Seem like it should be never negative. So we should change it type to `uint32_t` instead.

src/exodus/walletmodels.h
```diff
@@ -7,6 +7,7 @@
 
 #include "../serialize.h"
 #include "../uint256.h"
+#include "../pubkey.h"
```
ultimaweapon: We should ordered by ascending to prevent duplicated include. So it should go before `serialize.h`.

src/exodus/walletmodels.h
```diff
@@ -83,31 +77,42 @@ class SigmaMintId
...
+    uint256 spendTx;
     SigmaMintChainState chainState;
-    SigmaPrivateKey key;
-    uint256 spentTx;
 
 public:
     SigmaMint();
-    SigmaMint(PropertyId property, DenominationId denomination);
+    SigmaMint(
+        SigmaMintId const &id,
```
ultimaweapon: We need to remove this once we remove `id` from this class.

src/exodus/walletmodels.h
```diff
@@ -83,31 +77,42 @@ class SigmaMintId
...
+        int32_t count,
```
ultimaweapon: We need to update this once we update `count` field.

src/exodus/walletmodels.h
```diff
@@ -83,31 +77,42 @@ class SigmaMintId
...
+        const CKeyID& seedId,
+        const uint160& hashSerial);
```
ultimaweapon: Parameter name mismatched with field name.

src/exodus/walletmodels.h
```diff
@@ -148,18 +153,23 @@ struct hash<SigmaMintId>
     }
 };
 
+template<>
+struct hash<uint160> : hash<base_blob<160>>
```
ultimaweapon: This once should go to `src/uint256.h` instead.

src/exodus/walletmodels.cpp
```diff
@@ -62,31 +55,54 @@ bool SigmaMintId::operator!=(const SigmaMintId& other) const
...
-           key == other.key &&
-           spentTx == other.spentTx;
+    id = SigmaMintId();
+
+    count = 0;
+    seedId.SetNull();
+    serialId.SetNull();
+
+    spendTx.SetNull();
+    chainState = exodus::SigmaMintChainState();
```
ultimaweapon: We already in `exodus` namespace so no need to prefix it here.

ultimaweapon:
There is still unresolved comment so please fix it first.

src/exodus/test/walletmodels_tests.cpp
```diff
@@ -215,22 +205,58 @@ BOOST_AUTO_TEST_CASE(sigma_mint_unequality)
...
+    // Spend Tx
+    right = left;
+    right.spendTx = uint256();
+
+    BOOST_CHECK_NE(left, right);
+
+    // Block
     right = left;
-    right.chainState.Clear();
+    right.chainState.block = 501;
```
ultimaweapon: We should let chain state comparison to handle by itself so in the future it when there is more fields it will handle automatically.

src/exodus/test/walletmodels_tests.cpp
```diff
@@ -215,22 +205,58 @@ BOOST_AUTO_TEST_CASE(sigma_mint_unequality)
...
 
     BOOST_CHECK_NE(left, right);
 
-    // Key.
+    // Group
     right = left;
-    right.key.Generate();
+    right.chainState.group = 2;
```
ultimaweapon: Same here.

src/exodus/test/walletmodels_tests.cpp
```diff
@@ -215,22 +205,58 @@ BOOST_AUTO_TEST_CASE(sigma_mint_unequality)
...
+
+    BOOST_CHECK_NE(left, right);
+
+    // Index
+    right = left;
+    right.chainState.index = 51;
```
ultimaweapon: Same here.

src/exodus/walletmodels.h
```diff
@@ -128,13 +132,9 @@ struct hash<SigmaMintChainState>
...
-        size_t h = 0;
-
-        h ^= hash<int>()(state.block);
-        h ^= hash<SigmaMintGroup>()(state.group);
-        h ^= hash<SigmaMintIndex>()(state.index);
-
-        return h;
+        return hash<int>()(state.block)
+            ^ hash<SigmaMintGroup>()(state.group)
+            ^ hash<SigmaMintIndex>()(state.index);
```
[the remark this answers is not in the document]
thebevrishot: ok, then wait some comment from Stack Exchange

src/exodus/txprocessor.cpp
```diff
@@ -184,7 +185,7 @@ int TxProcessor::ProcessSimpleSpend(const CMPTransaction& tx)
     }
 
     assert(update_tally_map(tx.getReceiver(), property, amount, BALANCE));
-    sigmaDb->RecordSpendSerial(property, denomination, spend->serial, block);
+    sigmaDb->RecordSpendSerial(property, denomination, spend->serial, block, spendTx);
```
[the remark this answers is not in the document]
thebevrishot: NICE CATCH!

src/exodus/walletmodels.h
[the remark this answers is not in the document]
ultimaweapon: There is only one comments there, which tell it is personal preferences. So we can keep the new version.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+        ).
+        Finalize(countHash.data());
+
+    CHMAC_SHA512(countHash.data(), sizeof(countHash)).
+        Write(key.begin(), key.size()).
+        Finalize(result.data());
+
+    seed = uint512(result);
+
+    return GetSeedIndex(seedId);
```
ultimaweapon: We should cached the value of `GetSeedIndex` due to we call it multiple times in this function.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+}
+
+bool GetChildFromPath(string const &path, std::pair<uint32_t, bool> &child)
```
ultimaweapon: We should move this one to unnamed namespace.

ultimaweapon: IMO we should use regular expression instead. The benefits are:

- Easy to read what format we are dealing with.
- Less error prone.

ultimaweapon: We should throw instead if `path` is not in the expected format instead of return `bool`.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    child = {static_cast<uint32_t>(childPos), isHardened};
+
+    return true;
+}
+
+uint32_t SigmaWallet::GetSeedIndex(CKeyID const &seedId)
+{
+    if (!pwalletMain->mapKeyMetadata.count(seedId)) {
```
ultimaweapon: We need to lock `cs_wallet` before accessing `mapKeyMetadata`.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+        throw std::runtime_error("key not found");
+    }
+
+    auto const &meta = pwalletMain->mapKeyMetadata[seedId];
```
ultimaweapon: Better to use `find` instead. The problems with `[]` operator is it will create that entry automatically if it does not exists.

src/exodus/sigmawallet.cpp
ultimaweapon: If we follow my idea we don't need additional check here to see if `seedId` exists.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    if (child.second) {
+        throw std::runtime_error("hardened is not allowed");
+    }
+
+    return child.first;
+}
+
+bool SigmaWallet::SeedToPrivateKey(
```
ultimaweapon: IMO rename to `GeneratePrivateKey` looks better so it consistent with the other method.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    const uint512& seedZerocoin, exodus::SigmaPrivateKey& coin)
```
ultimaweapon: IMO just `seed` should be better so it consistent with the other place.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+        return false;
+    }
+
+    // Hash the public key in the group to obtain a serial number
+    auto serialNumber = priv.serialNumberFromSerializedPublicKey(OpenSSLContext::get_context(), &pubkey);
+
+    //hash randomness seed with Bottom 256 bits of seedZerocoin
+    Scalar randomness;
+    auto randomnessSeed = ArithToUint512(UintToArith512(seedZerocoin) >> 256).trim256();
+    randomness.memberFromSeed(randomnessSeed.begin());
```
ultimaweapon: Just `randomness.memberFromSeed(seedZerocoin.begin() + 32)` is enough. It is also easy to understand how bytes was manipulated too.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+        throw;
+    }
+}
+
+bool SigmaWallet::SetMintSeedSeen(
+    MintPoolEntry const &mintPoolEntry,
+    uint32_t propertyId,
+    uint8_t denomination,
+    exodus::SigmaMintChainState const &chainState,
+    uint256 const &spendTx)
```
ultimaweapon: When we recover mint that mean the mint should be always unspend. So we can remove this parameter and always set it to zero.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+        uint512 seed;
+        GenerateSeed(seedId, seed);
+
+        SigmaPrivateKey coin;
+        if (!SeedToPrivateKey(seed, coin)) {
+            return false;
+        }
+
+        serialId = GetSerialId(coin.serial);
+    } else {
```
ultimaweapon: The only cases that will invoke this method is when the mint not created yet but it already in the pool. That mean we don't need this case.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+}
+
+// Mint querying
+bool SigmaWallet::HasMint(SigmaMintId const &id) const
+{
+    CWalletDB walletdb(walletFile);
+    return walletdb.HasExodusHDMint(id);
+}
+
+bool SigmaWallet::HasSerial(secp_primitives::Scalar const &scalar) const
```
ultimaweapon: The method name is confusing. IMO rename to `HasMint` should be enough due to it already overload by parameter.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+{
+    CWalletDB walletdb(walletFile);
+    auto serialHash = GetSerialId(scalar);
```
ultimaweapon: `serialHash` is not consistent with the other. IMO rename it to just `id`.

src/exodus/sigmawallet.cpp
ultimaweapon: `scalar` is not indicated what data that need to pass in. Rename it to `serial` will be better.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    auto serialHash = GetSerialId(serial);
+    if (!walletdb.ReadExodusMintID(serialHash, id)) {
+        throw std::runtime_error("fail to read id");
+    }
+
+    return id;
+}
+
+size_t SigmaWallet::ListSigmaMints(
+    std::function<void(SigmaMint &)> const &f, bool unusedOnly, bool matureOnly) const
```
ultimaweapon: We don't need `unusedOnly` and `matureOnly` due to the caller can filter out from their callback.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+{
+    LOCK(pwalletMain->cs_wallet);
```
ultimaweapon: We don't need to lock here due to we did not access any wallet fields.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+}
+
+bool SigmaWallet::RemoveFromMintPool(SigmaPublicKey const &publicKey)
+{
+    auto &publicKeyIndex = mintPool.get<1>();
+    auto it = publicKeyIndex.find(publicKey);
+
+    if (it != publicKeyIndex.end()) {
+        publicKeyIndex.erase(it);
+        SaveMintPool();
```
ultimaweapon: We forgot to return.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    // publicKey is not in the pool
+    return false;
+}
+
+}; // exodus
```
ultimaweapon: Namespace closing brace don't need `;`.

src/exodus/sigmawallet.cpp
thebevrishot: this logic is also work with other situation which is running on big-endian machine and `>>` with a number which is not multiple of 32. Changing to `randomness.memberFromSeed(seedZerocoin.begin() + 32)` eliminate that functionality then I disagree to change.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+    return child.first;
+}
+
+bool SigmaWallet::SeedToPrivateKey(
+    const uint512& seedZerocoin, exodus::SigmaPrivateKey& coin)
+{
+    //convert state seed into a seed for the private key
+    uint256 nSeedPrivKey = seedZerocoin.trim256();
+    nSeedPrivKey = Hash(nSeedPrivKey.begin(), nSeedPrivKey.end());
```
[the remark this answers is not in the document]
thebevrishot: if change this one we need to change the below too. I have put reason why we should not change there.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+        }, false, false);
+
+    } catch (std::runtime_error const &e) {
+        LogPrintf("%s : fail to reset all mints chain state, %s\n", __func__, e.what());
+        throw;
+    }
+}
+
+bool SigmaWallet::SetMintSeedSeen(
```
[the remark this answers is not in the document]
thebevrishot: Good idea, I just added the function.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    MintPoolEntry const &mintPoolEntry,
```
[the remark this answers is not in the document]
thebevrishot: I dont edit this function just create new one `TryRecoverMint` which receive mint id and chain state.

src/exodus/sigmawallet.cpp
thebevrishot: I believe we should let this function be generic like this and make `TryRecoverMint` work that way instead.

src/exodus/sigmawallet.cpp
thebevrishot: make sense

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+            mintPool.erase(it);
+        }
+    }
+
+    if (updated) {
+        SaveMintPool();
+    }
+}
+
+size_t SigmaWallet::CountInMintPool(SigmaPublicKey const &pubKey)
```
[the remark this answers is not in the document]
thebevrishot: Use `IsMintInPool` is better i think

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    if (it == publicKeyIndex.end()) {
+        return false;
+    }
+
+    entry = *it;
+    return true;
+}
+
+// Generate coins to mint pool until amount of coins in mint pool touch the expected amount.
+size_t SigmaWallet::GenerateMintPool(size_t expectedCoins)
```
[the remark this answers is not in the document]
thebevrishot: agree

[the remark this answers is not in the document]
thebevrishot: Good idea to rename but I think `FillMintPool` is make more sense.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+    return generatedCoins;
+}
+
+void SigmaWallet::LoadMintPool()
+{
+    mintPool.clear();
+
+    CWalletDB walletdb(walletFile);
+
+    if (walletdb.HasExodusMintPool()) {
```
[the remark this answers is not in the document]
thebevrishot: We can. No any difference between check before and read but don't treat not found case as non exception.

src/exodus/sigmawallet.cpp
```diff
@@ -91,195 +72,193 @@ uint32_t SigmaWallet::CreateNextSeed(CKeyID &seedId, uint512& seed)
...
     }
 
-    if (childData.empty()) {
-        return false;
-    }
+    return child;
+}
 
-    auto childPos = std::stol(childData);
+secp_primitives::Scalar GetSerialFromPublicKey(
```
ultimaweapon: Our Exodus don't need ECDSA key to sign anything so we don't need to derive serial from ECDSA key.

src/exodus/sigmawallet.cpp
```diff
@@ -91,195 +72,193 @@ uint32_t SigmaWallet::CreateNextSeed(CKeyID &seedId, uint512& seed)
...
-        throw std::runtime_error("fail to parse HD key path");
-    }
-
-    if (child.second) {
-        throw std::runtime_error("hardened is not allowed");
+    uint32_t addressIndex;
+    try {
+        addressIndex = GetBIP44AddressIndex(it->second.hdKeypath);
+    } catch (std::runtime_error const &e) {
+        error("%s : fail to get child from, %s\n", __func__, e.what());
```
ultimaweapon: We should use `PrintToLog` instead due to `error` used to generate error code for Bitcoin but we don't that functionality here.

src/exodus/sigmawallet.cpp
```diff
@@ -91,195 +72,193 @@ uint32_t SigmaWallet::CreateNextSeed(CKeyID &seedId, uint512& seed)
...
     }
 
-    return child.first;
+    return addressIndex;
 }
 
-bool SigmaWallet::SeedToPrivateKey(
-    const uint512& seedZerocoin, exodus::SigmaPrivateKey& coin)
+bool SigmaWallet::GeneratePrivateKey(
+    const uint512& seed, exodus::SigmaPrivateKey& coin)
```
ultimaweapon: We don't need to prefix with `exodus` namespace due to we already in this namespace.

ultimaweapon: Better to rename `coin` to `key` instead so it match with the method name.

src/exodus/sigmawallet.cpp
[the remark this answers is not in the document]
ultimaweapon: After discussion, `uint` classes is just a byte array so we can go with my idea.

src/exodus/sigmawallet.cpp
ultimaweapon: After discussion the safest way is to treat it as a byte array so we can go with my idea.

src/exodus/wallet.h
```diff
@@ -34,6 +35,7 @@ class Wallet
 
         return output;
     }
+    void ResetState();
```
ultimaweapon: Right now this one will get invoked every time when we send tokens. My idea is we should remove this for now and handle it in #712 instead.

src/exodus/wallet.h
```diff
@@ -42,38 +44,35 @@ class Wallet
...
-    void ListSigmaMints(uint32_t propertyId, OutputIt it)
-    {
-        auto insertF = [propertyId, &it](SigmaMint& mint) {
-            if (mint.property == propertyId) {
-                *it++ = std::move(mint);
-            }
-        };
-
-        CWalletDB(walletFile).ListExodusMint<SigmaMintId, SigmaMint>(insertF);
+        auto mintWallet = this->mintWallet;
```
ultimaweapon: Why do we need to make a copy of `SigmaWallet` before invoking `ListMints`?

src/exodus/wallet.h
```diff
@@ -42,38 +44,35 @@ class Wallet
...
+        mintWallet.ListMints(it, false, false);
     }
 
     bool HasSigmaMint(const SigmaMintId& id);
     SigmaMint GetSigmaMint(const SigmaMintId& id);
+    boost::optional<SigmaMint> GetSpendableSigmaMint(
```
ultimaweapon: Why do we move this one to public?

src/exodus/wallet.cpp
```diff
@@ -17,7 +17,8 @@ namespace exodus {
 
 Wallet *wallet;
 
-Wallet::Wallet(const std::string& walletFile) : walletFile(walletFile)
+Wallet::Wallet(const std::string& walletFile)
+    : walletFile(walletFile), mintWallet()
```
ultimaweapon: We don't need to explicit invoke default constructor for `mintWallet`.

src/exodus/wallet.cpp
```diff
@@ -45,12 +54,16 @@ Wallet::~Wallet()
 
 SigmaMintId Wallet::CreateSigmaMint(PropertyId property, SigmaDenomination denomination)
 {
-    SigmaMint mint(property, denomination);
-    SigmaMintId id(mint, DefaultSigmaParams);
+    LOCK(pwalletMain->cs_wallet);
```
ultimaweapon: We don't need to lock here due to `mintWallet` already take care by them self.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
     if (!VerifySigmaSpend(mint->property, mint->denomination, mint->chainState.group, anonimitySet.size(), proof)) {
         throw WalletError(_("Failed to create spendable spend"));
     }
 
-    return SigmaSpend(SigmaMintId(*mint, DefaultSigmaParams), mint->chainState.group, anonimitySet.size(), proof);
+    return SigmaSpend(SigmaMintId(mint->property, mint->denomination, SigmaPublicKey(key, DefaultSigmaParams)),
+        mint->chainState.group, anonimitySet.size(), proof);
 }
 
 void Wallet::EraseSigmaMint(const SigmaMintId &id)
```
ultimaweapon: If should rename this method to `EraseUnconfirmedSigmaMint` too.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
+    mintWallet.PushFrontToMintPool(id);
 }
 
 bool Wallet::HasSigmaMint(const SigmaMintId& id)
 {
-    return CWalletDB(walletFile).HasExodusMint(id);
+    return mintWallet.HasMint(id);
+}
+
+bool Wallet::HasSigmaSpend(const secp_primitives::Scalar& serial, SigmaMint &mint)
```
ultimaweapon: Name is quite confusing. IMO rename to `HasSigmaMint` instead.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
+{
+    LOCK(pwalletMain->cs_wallet);
```
ultimaweapon: We don't need to lock here.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
     return mint;
 }
 
-void Wallet::SetSigmaMintUsedTransaction(SigmaMintId const &id, uint256 const &tx)
+boost::optional<SigmaMint>
+    Wallet::GetSpendableSigmaMint(PropertyId property, SigmaDenomination denomination)
 {
-    auto mint = GetSigmaMint(id);
+    // Get all spendable mints.
+    LOCK(pwalletMain->cs_wallet);
```
ultimaweapon: We don't need to lock here.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
+    std::vector<SigmaMint> spendables;
+    mintWallet.ListMints(std::back_inserter(spendables), true, true);
 
-    mint.spentTx = tx;
+    auto eraseFrom = std::remove_if(spendables.begin(), spendables.end(), [denomination](SigmaMint const &mint) -> bool {
+        return denomination != mint.denomination;
+    });
+    spendables.erase(eraseFrom, spendables.end());
```
ultimaweapon: We should use `boost::make_function_output_iterator` when invoke `ListMints` instead of erase it like this.

src/exodus/wallet.cpp
```diff
@@ -77,88 +90,132 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
+    mintWallet.UpdateMintSpendTx(id, tx);
+}
 
-    if (spendable.empty()) {
-        return boost::none;
+void Wallet::SetSigmaMintChainState(const SigmaMintId& id, const SigmaMintChainState& state)
+{
+    mintWallet.UpdateMintChainstate(id, state);
+}
+void Wallet::OnSpendAdded(
```
ultimaweapon: We should have one empty line before this method.

src/exodus/wallet.cpp
```diff
@@ -169,13 +226,24 @@ void Wallet::OnMintAdded(
     const SigmaPublicKey& pubKey,
     int block)
 {
+    LOCK(pwalletMain->cs_wallet);
```
ultimaweapon: We don't need to lock here anymore.

src/exodus/wallet.cpp
```diff
@@ -169,13 +226,24 @@ void Wallet::OnMintAdded(
...
+
     SigmaMintId id(property, denomination, pubKey);
+    auto isMintTracked = HasSigmaMint(id);
 
-    if (!HasSigmaMint(id)) {
-        return;
-    }
+    if (isMintTracked) {
```
ultimaweapon: `isMintTracked` is not quite correct due to we don't have tracker anymore. Just use `HasSigmaMint(id)` directly will be better.

src/exodus/sigmawallet.h
```diff
@@ -0,0 +1,156 @@
...
+    bool HasMint(SigmaMintId const &id) const;
+    bool HasMint(secp_primitives::Scalar const &serial) const;
+    SigmaMint GetMint(SigmaMintId const &id) const;
+    SigmaMint GetMint(secp_primitives::Scalar const &serial) const;
+    SigmaMintId GetMintId(secp_primitives::Scalar const &serial) const;
+
+    template<
+        class OutIt,
+        typename std::enable_if<is_iterator<OutIt>::value>::type* = nullptr
+    > OutIt ListMints(OutIt it, bool unusedOnly, bool matureOnly) const
```
ultimaweapon: IMO the two conditional parameters make code more hard to read from the caller site. Better to remove it and allow the caller site to filter by them self will looks better.

src/exodus/exodus.cpp
```diff
@@ -29,6 +29,7 @@
 #include "utilsbitcoin.h"
 #include "version.h"
 #ifdef ENABLE_WALLET
+#include "sigmawallet.h"
```
ultimaweapon: We should not need this anymore.

src/exodus/exodus.cpp
```diff
@@ -39,6 +40,7 @@
 #include "../coincontrol.h"
 #include "../coins.h"
 #include "../core_io.h"
+#include "../hdmint/hdmint.h"
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -73,7 +73,11 @@ class CHDChain
     static const int VERSION_BASIC = 1;
     static const int VERSION_WITH_BIP44 = 10;
     static const int CURRENT_VERSION = VERSION_WITH_BIP44;
+#ifdef ENABLE_EXODUS
+    static const int N_CHANGES = 4; // standard = 0/1, mint = 2, exodus = 3
```
ultimaweapon: We should make it always 4 without conditional; otherwise the data in wallet file will not consistent.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    }
+
+    template<class Key>
+    bool EraseExodusMintID(const Key& k)
+    {
+        return Erase(std::make_pair(std::string("exodus_mint_id"), k));
+    }
+
+    template<class K, class V>
+    bool ReadExodusHDMint(const K& k, V& v)
```
ultimaweapon: We don't need to identify it as HD due to we only have HD version.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
     {
-        return Exists(std::make_pair(std::string("exodus_sigma_mint"), k));
+        return Read(std::make_pair(std::string("exodus_hdmint"), k), v);
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    }
+
+    template<class K>
+    bool HasExodusHDMint(const K& k)
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    {
+        return Exists(std::make_pair(std::string("exodus_hdmint"), k));
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    }
+
+    template<class K, class V>
+    bool WriteExodusHDMint(const K &k, const V &v)
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    {
+        return Write(std::make_pair(std::string("exodus_hdmint"), k), v, true);
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    }
+
+    template<class K>
+    bool EraseExodusHDMint(const K& k)
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -286,22 +290,70 @@ class CWalletDB : public CDB
...
+    {
+        return Erase(std::make_pair(std::string("exodus_hdmint"), k));
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -312,44 +364,51 @@ class CWalletDB : public CDB
 
 
     template<typename K, typename V, typename InsertF>
-    void ListExodusMint(InsertF insertF)
+    void ListExodusHDMints(InsertF insertF)
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -312,44 +364,51 @@ class CWalletDB : public CDB
...
-        unsigned int fFlags = DB_SET_RANGE;
+        unsigned int flags = DB_SET_RANGE;
         while (true) {
+
             // Read next record
             CDataStream ssKey(SER_DISK, CLIENT_VERSION);
-            if (fFlags == DB_SET_RANGE)
-                ssKey << make_pair(std::string("exodus_sigma_mint"), K());
+            if (flags == DB_SET_RANGE) {
+                ssKey << std::make_pair(string("exodus_hdmint"), K());
```
ultimaweapon: Same here.

src/wallet/walletdb.h
```diff
@@ -312,44 +364,51 @@ class CWalletDB : public CDB
...
+            } else if (ret != 0) {
+                cursor->close();
+                throw runtime_error(std::string(__func__)+" : error scanning DB");
             }
+
             // Unserialize
-            std::string strType;
-            ssKey >> strType;
-            if (strType != "exodus_sigma_mint")
+            string type;
```
ultimaweapon: Namespace issue.

src/wallet/walletdb.h
```diff
@@ -312,44 +364,51 @@ class CWalletDB : public CDB
...
+            ssKey >> type;
+            if (type != "exodus_hdmint") {
```
ultimaweapon: Same as previous issue.

src/exodus/sigmawallet.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+bool MintPoolEntry::operator!=(MintPoolEntry const &another) const
+{
+    return !(*this == another);
+}
+
+SigmaWallet::SigmaWallet(const std::string& walletFile) : walletFile(walletFile)
+{
+    // Don't try to do anything else if the wallet is locked.
+    if (pwalletMain->IsLocked()) {
```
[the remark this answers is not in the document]
thebevrishot: Call after construct instead.

src/exodus/sigmawallet.h
```diff
@@ -59,49 +49,55 @@ class SigmaWallet
...
         >
-    > IndexedMintPool;
+    > MintPool;
 
     std::string walletFile;
-    IndexedMintPool mintPool;
+    MintPool mintPool;
     uint160 masterId;
 
+    static constexpr unsigned mintPoolCapacity = 20;
```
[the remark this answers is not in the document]
ultimaweapon: I mean it should all capitalized. e.g. `MINTPOOL_CAPACITY`. Sorry for not make it clear.

src/exodus/test/sigmawallet_tests.cpp
```diff
@@ -0,0 +1,328 @@
...
+{
+    // get sequence
+    std::vector<MintPoolEntry> mints;
+    sigmaWallet.GetMintPoolEntry(std::back_inserter(mints));
+
+    std::vector<uint32_t> mintPoolIndexs;
+    for (auto const &mint : mints) {
+        mintPoolIndexs.push_back(mint.index);
+    }
+    std::sort(mintPoolIndexs.begin(), mintPoolIndexs.end());
```
ultimaweapon: We should not sort it; otherwise we won't know the order is not right.

src/exodus/test/sigmawallet_tests.cpp
```diff
@@ -0,0 +1,328 @@
...
+    ));
+
+    // verify state after, mint wallet should be updated
+    std::vector<MintPoolEntry> mintPoolAfter;
+    sigmaWallet.GetMintPoolEntry(std::back_inserter(mintPoolAfter));
+
+    BOOST_CHECK(mintPool != mintPoolAfter);
+
+    // mintPool[1:] == mintPoolAfter[:size - 1]
+    BOOST_CHECK_EQUAL(true,
```
ultimaweapon: Better to check if mint pool also refilled.

src/exodus/test/sigmawallet_tests.cpp
```diff
@@ -0,0 +1,328 @@
...
+            mints.begin(), mints.end(),
+            std::vector<SigmaMint>{
+                unconfirmed.first, unspend.first, spend.first
+            }.begin(),
+            sigmaMintComparer
+        )
+    );
+}
+
+BOOST_AUTO_TEST_CASE(push_out_wallet_mint_back)
```
ultimaweapon: We need to update case name.

src/exodus/test/sigmawallet_tests.cpp
```diff
@@ -0,0 +1,328 @@
...
+    std::vector<MintPoolEntry> mintPool;
+    sigmaWallet.GetMintPoolEntry(std::back_inserter(mintPool));
+
+    SigmaMint mint;
+    SigmaPrivateKey privKey;
+    std::tie(mint, privKey) = sigmaWallet.GenerateMint(1, 0);
+
+    SigmaMintId id(1, 0, SigmaPublicKey(privKey, DefaultSigmaParams));
+
+    sigmaWallet.DeleteUnconfirmedMint(id);
```
ultimaweapon: We should check if mint exists before we delete and not exists after we delete too.

src/exodus/wallet.cpp
```diff
@@ -101,46 +104,53 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
...
-    LOCK(pwalletMain->cs_wallet);
     std::vector<SigmaMint> spendables;
-    mintWallet.ListMints(std::back_inserter(spendables), true, true);
 
-    auto eraseFrom = std::remove_if(spendables.begin(), spendables.end(), [denomination](SigmaMint const &mint) -> bool {
-        return denomination != mint.denomination;
-    });
-    spendables.erase(eraseFrom, spendables.end());
+    mintWallet.ListMints(boost::make_function_output_iterator(
+        [denomination, &spendables] (SigmaMint const &mint) {
```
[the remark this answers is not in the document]
thebevrishot: nice catch

src/exodus/sigmawallet.cpp
```diff
@@ -276,62 +229,61 @@ SigmaMint SigmaWallet::UpdateMint(const SigmaMintId &id, const std::function<voi
...
 
-    SigmaMintId id(propertyId, denomination, pubcoin);
+    MintPoolEntry entry;
+    if (!GetMintPoolEntry(id.pubKey, entry)) {
+        throw std::runtime_error("Fail to get mint from pool");
+    }
 
-    uint160 serialId;
+    // Regenerate the mint
+    auto const &pubcoin = id.pubKey;
```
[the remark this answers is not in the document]
thebevrishot: removed

src/exodus/sigmawallet.cpp
```diff
@@ -276,62 +229,61 @@ SigmaMint SigmaWallet::UpdateMint(const SigmaMintId &id, const std::function<voi
...
+    auto const &seedId = entry.seedId;
+    auto seedIndex = GetSeedIndex(seedId);
```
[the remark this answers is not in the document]
thebevrishot: removed

src/exodus/sigmawallet.cpp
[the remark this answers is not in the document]
thebevrishot: removed

PR #668, opened 2019-09-06
fix for help RPC call requiring passphrase

riordant:
Fixes https://github.com/zcoinofficial/zcoin/issues/672

## PR intention
The result of the `help` call with a locked & encrypted wallet is currently:
```
Error: Please enter the wallet passphrase with walletpassphrase first.
```
The issue is that the function `regeneratemintpool` does not have a `help` section, but requires wallet unlock. This PR adds that and makes the `help` function accessible from a locked wallet again.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2654,6 +2654,11 @@ UniValue fundrawtransaction(const UniValue& params, bool fHelp)
 
 UniValue regeneratemintpool(const UniValue &params, bool fHelp) {
 
+        if (fHelp || params.size() > 0)
+	        throw runtime_error(
+       	        	"regeneratemintpool\n"
+       		        "\nIf issues exist with the keys that map to mintpool entries in the DB, this function corrects them.\n"
+                );
     if (pwalletMain->IsLocked())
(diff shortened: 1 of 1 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -2654,6 +2654,11 @@ UniValue fundrawtransaction(const UniValue& params, bool fHelp)
 
 UniValue regeneratemintpool(const UniValue &params, bool fHelp) {
 
+        if (fHelp || params.size() > 0)
+	        throw runtime_error(
+       	        	"regeneratemintpool\n"
+       		        "\nIf issues exist with the keys that map to mintpool entries in the DB, this function corrects them.\n"
+                );
```
ultimaweapon: We should provides examples too so it will be the same as other Bitcoin RPCs.

riordant: I had considered that but no arguments are being passed. What do you suggest in this case?

ultimaweapon: We should show it without arguments like `getrawchangeaddress`.

riordant: Updated to include this.

PR #676, opened 2019-09-10
Prevent denomination creation if it not accepted by consensus

### LINKED ISSUE #665: Prevent user from create denomination if property is not enable sigma
This is on the client side. The consensus already prevent this. We should initialize integration tests for Exodus too.

- [x] Implementation [3]
- [x] Review [1]

ultimaweapon:
Resolve #665.

### DIFF
--- a/qa/rpc-tests/exodus_create_denomination.py
+++ b/qa/rpc-tests/exodus_create_denomination.py
@@ -8,7 +8,7 @@ class ExodusCreateDenominationTest(ExodusTestFramework):
         super().run_test()
 
         # create non-sigma token
-        self.nodes[0].exodus_sendissuancefixed(self.addrs[0], 1, 2, 0, '', '', 'Normal Token', '', '', '1000000')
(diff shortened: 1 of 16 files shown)


qa/rpc-tests/exodus_create_denomination.py
```diff
@@ -0,0 +1,121 @@
...
+        for i in range(255 - 2):
+            self.nodes[0].exodus_sendcreatedenomination(self.addrs[0], 4, str(i + 2))
+            self.nodes[0].generate(1) # we need to mine here otherwise the input chaining will be too long
+
+        assert_raises_message(
+            JSONRPCException,
+            'No more room for new denomination',
+            self.nodes[0].exodus_sendcreatedenomination, self.addrs[0], 4, '1000'
+        )
+
```
thebevrishot: please add a test case that try to add a divisible amount denomination to the indivisible asset.

PR #690, opened 2019-09-18
Fix BIP44 keypool issue with older encrypted wallets

riordant:
## PR intention
- Encrypted wallets that existed before the upgrade to HDMint/BIP44 continue to use the old keypool until it is exhausted. That means that these wallets will not use keys along the BIP44 path until all the current keys are used. The intention of this PR is to generate a new keypool, consisting of BIP44 keys, following wallet unlock. 

- This issue does not affect keys used for minting, new wallets generated, or older unencrypted wallets.

-While not a critical issue (the existing keys are still of course valid and retrievable using the master seed), Given the upgrade to BIP39, it is important that the same keys are generated between third party wallets as well as our own.

## Code changes brief
- `SetHDChain` now called following wallet unlock
- If previous wallet version detected in this function, upgrade wallet, and generate new keypool.
- Also removes a line of code that sets the same index as previous; the index should reset to 0, as this is a new keypath.

### DIFF
--- a/src/wallet/crypter.cpp
+++ b/src/wallet/crypter.cpp
@@ -205,9 +205,11 @@ bool CCryptoKeyStore::Unlock(const CKeyingMaterial& vMasterKeyIn, const bool& fF
         vMasterKey = vMasterKeyIn;
         fDecryptionThoroughlyChecked = true;
         if(!fFirstUnlock && zwalletMain){
-            uint160 hashSeedMaster = pwalletMain->GetHDChain().masterKeyID;
+            CHDChain hdChain = pwalletMain->GetHDChain();
+            uint160 hashSeedMaster = hdChain.masterKeyID;
             zwalletMain->SetupWallet(hashSeedMaster, false);
             zwalletMain->SyncWithChain();
+            pwalletMain->SetHDChain(hdChain, false); // Used to upgrade normal keys to BIP44
         }
     }
     NotifyStatusChanged(this);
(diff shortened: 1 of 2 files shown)


qashban4:
@ultimaweapon and @thebevrishot to review and conclude the same on 2019.09-24.

riordant:
> I don't think we need this. The reason is even with this PR it will still not use BIP44 to generate new keys due to the expression `hdChain.masterKeyID.IsNull()` in `CWallet::GenerateNewKey` still `true`.

I don't understand what you mean. When I say "older", I don't mean "non-HD" - I mean wallets that were generated before we introduced BIP44. So `hdChain.masterKeyID.IsNull()` will not be `true` in the case that it's a HD wallet. It's definitely needed

ultimaweapon:
I see. Did not know we have 2 versions of HD before. I will dismiss my request changes to let other who understand what going on review instead due to I don't have time to understand what is going on here.

Thanks for information.

src/wallet/crypter.cpp
```diff
@@ -205,9 +205,11 @@ bool CCryptoKeyStore::Unlock(const CKeyingMaterial& vMasterKeyIn, const bool& fF
         vMasterKey = vMasterKeyIn;
         fDecryptionThoroughlyChecked = true;
         if(!fFirstUnlock && zwalletMain){
-            uint160 hashSeedMaster = pwalletMain->GetHDChain().masterKeyID;
+            CHDChain hdChain = pwalletMain->GetHDChain();
```
thebevrishot: I believe this one is accidentally copying. It'll be better if we use `auto const &hdChain = ...` instead.

riordant: Agree.

riordant:
> I see. Did not know we have 2 versions of HD before. I will dismiss my request changes to let other who understand what going on review instead due to I don't have time to understand what is going on here.
> 
> Thanks for information.

Ok sure. Yes, specifically what's changing is the version of `CHDChain`. (With mnemonic PR, the version is upgrading again).

src/wallet/crypter.cpp
riordant: Updated, please re-review

PR #691, opened 2019-09-18
Refactor sigma primitives

### LINKED ISSUE #685: Refactor sigma primitives
Right now it implicitly use parameters from Zcoin. We should make it explicitly. Also `SigmaProof` class had some workaround that can be fixed now.

- [x] Implementation [3]
- [x] Review [1]

ultimaweapon:
Resolve #685.


src/exodus/sigmadb.h
```diff
@@ -89,6 +89,14 @@ class SigmaDatabase : public CDBBase
 
     uint16_t groupSize;
 
+public:
+    bool VerifySpend(
+        PropertyId property,
+        DenominationId denomination,
+        MintGroupId group,
+        size_t groupSize,
+        const SigmaProof& proof);
```
thebevrishot: please move this logic to elsewhere.

PR #698, opened 2019-09-21
Update Payment-Codes (#1)

HysMagus:
- Added a Payment Codes tab and screen to the existing Qt keychain wallet
- User's own Payment Code creates and validates itself upon first installation
- Masking of a recipient's Payment Code in preparation for inclusion into a Notification Transaction addressed to the recipient
- Some code cleanup, refactoring, reorganization for BIP-47's needs


HysMagus:
Created in Error

PR #704, opened 2019-09-24
Remove mint when sendmint fail

thebevrishot:
closes https://github.com/zcoinofficial/zcoin/issues/660

### DIFF
--- a/qa/rpc-tests/exodus_sendmint.py
+++ b/qa/rpc-tests/exodus_sendmint.py
@@ -12,7 +12,7 @@ class ExodusSendMintTest(ExodusTestFramework):
         self.nodes[0].generatetoaddress(100, self.addrs[0])
         self.nodes[0].generate(sigma_start_block - self.nodes[0].getblockcount())
 
-        assert(sigma_start_block == self.nodes[0].getblockcount())
+        assert_equal(sigma_start_block, self.nodes[0].getblockcount())
 
         # create non-sigma
         self.nodes[0].exodus_sendissuancefixed(
--- a/qa/rpc-tests/exodus_sendmint.py
+++ b/qa/rpc-tests/exodus_sendmint.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+from test_framework.authproxy import JSONRPCException
+from test_framework.test_framework import ExodusTestFramework
+from test_framework.util import assert_equal, assert_raises_message
+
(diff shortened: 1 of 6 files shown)


src/exodus/wallet.cpp
```diff
@@ -86,6 +86,11 @@ SigmaSpend Wallet::CreateSigmaSpend(PropertyId property, SigmaDenomination denom
     return SigmaSpend(SigmaMintId(*mint, DefaultSigmaParams), mint->chainState.group, anonimitySet.size(), proof);
 }
 
+bool Wallet::EraseSigmaMint(const SigmaMintId &id)
+{
+    return CWalletDB(walletFile).EraseExodusMint(id);
```
ultimaweapon: We should throw instead if it fail so it will more easy to use from the caller site.

src/exodus/rpctx.cpp
```diff
@@ -1635,6 +1638,13 @@ UniValue exodus_sendmint(const UniValue& params, bool fHelp)
 
     // check error and return the txid (or raw hex depending on autocommit)
     if (result != 0) {
+        for (auto const &id : ids) {
+            try {
+                wallet->EraseSigmaMint(id);
+            } catch (std::runtime_error const &e) {
+                LogPrintf("%s : Fail to erase sigma mints, %s", __func__, e.what());
```
ultimaweapon: Need to add line feed at the end of string.

qa/rpc-tests/exodus_sendmint.py
```diff
@@ -0,0 +1,105 @@
...
+class ExodusSendMintTest(ExodusTestFramework):
+    def run_test(self):
+        super().run_test()
+
+        sigma_start_block = 500
+
+        self.nodes[0].generatetoaddress(100, self.addrs[0])
+        self.nodes[0].generate(sigma_start_block - self.nodes[0].getblockcount())
+
+        assert(sigma_start_block == self.nodes[0].getblockcount())
```
ultimaweapon: Why do we use `assert` instead of `assert_equal`?

src/exodus/rpctx.cpp
thebevrishot: Nice catch

qa/rpc-tests/exodus_sendmint.py
thebevrishot: good idea to change to `assert_equal`.

PR #714, opened 2019-10-02
Fixed the unspent index for remints of 50 XZC

a-bezrukov:
## PR intention
Fixes an issue with calculating balance of remint transactions in case of 50XZC remints, which produced 2 outputs of 25 XZC Sigma mints. 

## Code changes brief
Now all remint outputs are summed up and used as the amount for the input.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -559,13 +559,16 @@ void CDbIndexHelper::ConnectTransaction(CTransaction const & tx, int height, int
         }
     }
 
-    no = 0;
     if(tx.IsZerocoinRemint()) {
-        for (std::vector<CTxIn>::const_iterator iter = tx.vin.begin(); iter != tx.vin.end(); ++iter) {
-            CTxIn const & input = *iter;
-            handleRemint(input, no, tx.GetHash(), height, txNumber, tx.vout[no].nValue, addressIndex, addressUnspentIndex, spentIndex);
-            ++no;
+        CAmount remintValue = 0;
+        for (std::vector<CTxOut>::const_iterator iter_out = tx.vout.begin(); iter_out != tx.vout.end(); ++iter_out) {
+            remintValue += iter_out->nValue;
         }
+        if (tx.vin.size() != 1) {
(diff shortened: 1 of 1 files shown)


src/txdb.cpp
```diff
@@ -559,13 +559,16 @@ void CDbIndexHelper::ConnectTransaction(CTransaction const & tx, int height, int
...
-            ++no;
+        CAmount remintValue = 0;
+        for (std::vector<CTxOut>::const_iterator iter_out = tx.vout.begin(); iter_out != tx.vout.end(); ++iter_out) {
+            remintValue += iter_out->nValue;
         }
+        if (tx.vin.size() != 1) {
+           error("A Zerocoin to Sigma remint tx shoud have just 1 input");
+           return;
+        }
+        handleRemint(tx.vin[0], 0, tx.GetHash(), height, txNumber, remintValue, addressIndex, addressUnspentIndex, spentIndex);
```
riordant: The second parameter `inputNo` to `handleRemint ` is no longer necessary and should be removed. This will also imply by the code that a remint can only have one input.

a-bezrukov: Good catch!

src/txdb.cpp
```diff
@@ -559,13 +559,16 @@ void CDbIndexHelper::ConnectTransaction(CTransaction const & tx, int height, int
...
     }
 
-    no = 0;
     if(tx.IsZerocoinRemint()) {
-        for (std::vector<CTxIn>::const_iterator iter = tx.vin.begin(); iter != tx.vin.end(); ++iter) {
-            CTxIn const & input = *iter;
-            handleRemint(input, no, tx.GetHash(), height, txNumber, tx.vout[no].nValue, addressIndex, addressUnspentIndex, spentIndex);
-            ++no;
+        CAmount remintValue = 0;
+        for (std::vector<CTxOut>::const_iterator iter_out = tx.vout.begin(); iter_out != tx.vout.end(); ++iter_out) {
```
psolstice: Sorry for nitpicking but if you're gonna do the changes to the code pls change to C++11 `for`. Iterators are way outdated

PR #725, opened 2019-10-07
Adjust activation block for Sigma feature

ultimaweapon:
We want to activate it on 21 October. This is the final decision after discussion with Poramin.


src/exodus/rules.cpp
```diff
@@ -130,10 +130,10 @@ CMainConsensusParams::CMainConsensusParams()
     FREEZENOTICE_FEATURE_BLOCK = 999999;
 
     // Sigma releated
-    SIGMA_FEATURE_BLOCK = 210000;
+    SIGMA_FEATURE_BLOCK = 208500; // 21 Oct 2018
```
thebevrishot: I believe the comment should be `21 Oct 2019`

thebevrishot: I try to comment in my way and I got another result
This is my solution
each block take ~300(297) seconds
If I start calculate from block 204244 which is mined at 1570515251
And the time we need to activate is 1571644800 (10/21/2019 @ 8:00am (UTC))
Then The different time is 1571644800 - 1570515251
Then I transform it to block amount by divide by 300(297) the result will be 3765(3803) blocks
Finally we should activate at block 204244 + 3765(3803) = 208009(208047)

ultimaweapon: Good idea.

PR #729, opened 2019-10-08
Implement private send using Sigma for Exodus

ultimaweapon:


qashban4:
@riordant Tadhg will do review 2019-10-09

riordant:
Starting review today 11/10/2019

src/main.cpp
```diff
@@ -4,6 +4,11 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "main.h"
+
+#if defined(HAVE_CONFIG_H)
```
riordant: This is defined in `main.h`, it compiles without it. Unless I'm missing something, it should be removed from here.

ultimaweapon: Yes it is included in `main.h` too. The problems is the one in `h` may removed in the future. If we don't have this in `cpp` it will cause all conditional on features to disable (e.g. `#ifdef ENABLE_WALLET` and `#ifdef ENABLE_EXODUS`). So the best way is to include what that file need even if it already included by other files so it will not broken when the other files removed it. It is similar to the problems you found on macOS for `std::array`.

riordant: Ok, agreed

PR #730, opened 2019-10-09
Fix issue with failing GetAddressGroupings with Private Spends

riordant:
Fixes https://github.com/zcoinofficial/zcoin/issues/723

## PR intention
`GetAddressGroupings()` in `wallet/wallet.cpp` is causing a segfault if the wallet contains Zerocoin or Sigma spends. The issue is it expects each input of the transaction it's looking at to be associated with an address, spend inputs are not.

## Code changes brief
Avoid the code block that deals with inputs entirely by checking if the overall tx is a spend. This is sufficient as spends cannot have non-spend inputs.


riordant:
@thebevrishot good catch. Updated

PR #735, opened 2019-10-11
Fix MacOS build issues on Exodus

riordant:
### a)
 Brief:
```
exodus/rpc.cpp:621:26: error: call to 'Pair' is ambiguous
    payloadObj.push_back(Pair("payloadsize", payload.size()));
exodus/rpc.cpp:1019:32: error: call to 'Pair' is ambiguous
        denomination.push_back(Pair("id", i));
```

Extended:
```
./univalue/include/univalue.h:180:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, uint64_t u64Val)
                                              ^
./univalue/include/univalue.h:187:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, int64_t i64Val)
                                              ^
./univalue/include/univalue.h:194:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, bool iVal)
                                              ^
./univalue/include/univalue.h:201:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, int iVal)
                                              ^
./univalue/include/univalue.h:208:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, double dVal)
                                              ^
./univalue/include/univalue.h:215:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(const char *cKey, const UniValue& uVal)
                                              ^
./univalue/include/univalue.h:221:47: note: candidate function
static inline std::pair<std::string,UniValue> Pair(std::string key, const UniValue& uVal)
```

On MacOS, it seems if the value argument to `Pair` is a number type, and is NOT one of the number value types above, it must be explicitly cast. Both values have been cast to `int64_t`.

### b)
Brief:
```
uint256.cpp:23:19: error: type 'const std::array<unsigned char, WIDTH>' does not provide a subscript operator
    memcpy(data, &vch[0], sizeof(data));
```

Extended:
```
uint256.cpp:23:19: error: type 'const std::array<unsigned char, WIDTH>' does not provide a subscript operator
    memcpy(data, &vch[0], sizeof(data));
                  ^~~ ~
uint256.cpp:101:26: note: in instantiation of member function 'base_blob<160>::base_blob' requested here
template base_blob<160>::base_blob(const std::array<unsigned char, 20>&);
                         ^
uint256.cpp:23:19: error: type 'const std::array<unsigned char, WIDTH>' does not provide a subscript operator
    memcpy(data, &vch[0], sizeof(data));
                  ^~~ ~
uint256.cpp:111:26: note: in instantiation of member function 'base_blob<256>::base_blob' requested here
template base_blob<256>::base_blob(const std::array<unsigned char, 32>&);
                         ^
uint256.cpp:23:19: error: type 'const std::array<unsigned char, WIDTH>' does not provide a subscript operator
    memcpy(data, &vch[0], sizeof(data));
                  ^~~ ~
uint256.cpp:121:26: note: in instantiation of member function 'base_blob<512>::base_blob' requested here
template base_blob<512>::base_blob(const std::array<unsigned char, 64>&);
```
The `std::array` type cannot be accessed via the subscript operator. Just passing it without this is sufficient. (Not sure why this is different on MacOS)


src/uint256.cpp
```diff
@@ -20,7 +20,7 @@ base_blob<BITS>::base_blob(const std::vector<unsigned char>& vch)
 template<unsigned int BITS>
 base_blob<BITS>::base_blob(const std::array<unsigned char, WIDTH>& vch)
 {
-    memcpy(data, &vch[0], sizeof(data));
+    memcpy(data, &vch, sizeof(data));
```
ultimaweapon: `std::array` does not overload operator `&`. That mean this one will return the address of `std::array` instead of the underlying array that stored the data. If the implementation put the underlying array as the first member and compiler does not put any implementation specific data before it then it will work but it is still dangerous. So better to use `vch.data()` instead.

riordant: The actual issue was a missing header (error message didn't indicate as such). reverted to original and added this header.

Issue #743, opened 2019-10-18
Reject Double mint into mempool


levonpetrosyan93:
Prevent adding mint into mempool when it contains mint with the same pubCoin value.

PR #744, opened 2019-10-20
Double mint fix

levonpetrosyan93:
Fix for #743

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1409,17 +1409,21 @@ bool AcceptToMemoryPoolWorker(
                 }
             }
         }
-        if(tx.IsSigmaMint()) {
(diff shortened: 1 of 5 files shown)


levonpetrosyan93:
> I think we need a unit test for it

Yes, Thank you, will add.

src/main.cpp
```diff
@@ -3172,6 +3180,15 @@ bool ConnectBlock(const CBlock &block, CValidationState &state, CBlockIndex *pin
                 sigmaState->RemoveSpendFromMempool(zcSpendSerial);
             }
        }
+        if(tx.IsSigmaMint()){
```
ultimaweapon: Same as previous issue.

PR #745, opened 2019-10-20
Moving Secp Wrappers to Sigma folder

levonpetrosyan93:


levonpetrosyan93:
> This is not the right way. The C++ wrapper need to access internal of secp256k1. It should compile into the same library as secp256k1. The example of the problems caused by this PR is it will use different compilation options for wrapper and secp256k1.

I think it is not depended from compilation options of secp256k1, It is not part of secp lib, it is our code so would be better keep it separate.

ultimaweapon:
One of my idea is refactor secp256k1 to expose it internal APIs that we need to public. Then our wrapper can use those APIs via public headers. The problems with this way is it will make secp256k1 upgrading harder.

PR #748, opened 2019-10-21
Add Zcoin icon to Readme

sctnightcore:
## PR intention
DeleteMe: Mandatory: What this PR is intended to do, what change it introduces, what issue it solves.

## Code changes brief
DeleteMe: Optional: What are the architectural, UX, or other kind of changes are made which are hard to induce from the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 <img src="https://zcoin.io/storage/2017/12/ci_logo_monochrome.svg" width="500">
-===============
 
 [![latest-release](https://img.shields.io/github/release/zcoinofficial/zcoin)](https://github.com/zcoinofficial/zcoin/releases)
 [![GitHub last-release](https://img.shields.io/github/release-date/zcoinofficial/zcoin)](https://github.com/zcoinofficial/zcoin/releases)
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Zcoin
+<img src="https://zcoin.io/storage/2017/12/ci_logo_monochrome.svg" width="500">
 ===============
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-Zcoin
+<img src="https://zcoin.io/storage/2017/12/ci_logo_monochrome.svg" width="500">
```
ultimaweapon: @reubenyap do we have a better URL than this one?

README.md
```diff
@@ -1,4 +1,4 @@
...
 ===============
```
ultimaweapon: This line should be removed too.

sctnightcore: =============== @ultimaweapon 
this ?

README.md
sctnightcore: I think size is ok for me .

README.md
sctnightcore: Fixed : 7d1be49

README.md
reubenyap: What is the issue with the URL?

ultimaweapon: @reubenyap the problems with this URL is it look like it is managed by CMS. So it is better to use the one that we uploaded to the web server by ourself.

riordant: @ultimaweapon where can we find that URL?

ultimaweapon: Do you mean the one that we uploaded by ourself? If yes, no I don't have.

ultimaweapon: @reubenyap can we upload this image to our website with a proper location?

reubenyap:
No longer needed since rebrand to Firo

PR #756, opened 2019-10-28
Verbose commenting for each function in HDMint

riordant:
## PR intention
- Added comments for each function in the `hdmint/` directory, following the `javadoc` specification
- Removed unused functions in `hdmint/tracker.cpp`
- Fixed a bug where if `SetMintSeedSeen()` returns `false` in `SyncWithChain()`, the loop doesn't terminate. This is solved by bringing `setChecked` outside of the parent `while` loop.

### DIFF
--- a/src/hdmint/hdmint.cpp
+++ b/src/hdmint/hdmint.cpp
@@ -5,11 +5,21 @@
 #include <tinyformat.h>
 #include "hdmint.h"
 
+/**
+ * CHDMint empty constructor
+ *
+ * @return CHDMint object
+ */
 CHDMint::CHDMint()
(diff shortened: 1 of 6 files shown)


src/hdmint/hdmint.cpp
```diff
@@ -5,11 +5,21 @@
 #include <tinyformat.h>
 #include "hdmint.h"
 
+/**
+ * CHDMint empty constructor
+ *
+ * @return CHDMint object
+ */
 CHDMint::CHDMint()
```
psolstice: Redundant comments like that won't do any good

PR #757, opened 2019-10-29
Enable cert

a-bezrukov:
## PR intention
This PR enables binary certification for windows. 

## Code changes brief
The certification server IP placed in the environment.conf file, so that only this Jenkins instance can access the environment.


ultimaweapon:
Seems like we need to update `contrib/gitian-descriptors/gitian-win-sign-all.yml` due to `contrib/gitian-descriptors/gitian-win.yml` is already moved to Ubuntu 18.04.

a-bezrukov:
Signing only the installer for now

PR #761, opened 2019-11-01
Fix for Non-HD wallet dump

riordant:
## PR intention
- Fixes a missing check for `masterKeyID` where the wallet is non-HD in `dumpwallet`
- Adds python test
- Show remint window once Zerocoins are imported, not ony after discovering a new block

Note this PR is failing due to needing https://github.com/zcoinofficial/zcoin/pull/758 so this should be merged first.


src/wallet/rpcdump.cpp
```diff
@@ -722,7 +726,8 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
         std::string strTime = EncodeDumpTime(it->first);
         std::string strAddr = CBitcoinAddress(keyid).ToString();
         CKey key;
-        pwalletMain->mapKeyMetadata[keyid].ParseComponents();
+        if(!masterKeyID.IsNull())
+            pwalletMain->mapKeyMetadata[keyid].ParseComponents();
```
psolstice: `ParseComponents()` function is crashing on invalid path (including empty path). This fixes non-HD wallet case but malformed path case remains. It needs to throw an exception instead

src/wallet/rpcdump.cpp
```diff
@@ -819,4 +826,4 @@ UniValue dumpwallet_zcoin(const UniValue& params, bool fHelp)
     dumpParams.push_back(params[0]);
 
     return dumpwallet(dumpParams, false);
-}
+}
```
psolstice: Also I found that there are bitcoin tests in 0.14 where `walletdb.ListPubCoin` fails due to the physical absence of wallet file. I fixed this in core14 branch by adding a check if the wallet file name is empty

riordant: Ok great. I will add your changes to this PR anyway

PR #769, opened 2019-11-13
Update Dockerfile for clientapi PR

riordant:
## PR intention
The clientapi PR will add the need for one more dependancy, `libminizip-dev`. The Jenkins file for CI pulls from the remote Docker image and does not use the local one. So this PR adds the dependancy to the main dockerfile so that CI can pass on `clientapi`.

An alternative solution is to get the Jenkins file to pull from the Dockerfile of the branch, though as it's a rare occurrence that it changes this way is probably ok.

(It should also be noted that the dependancy is only needed if the configure option `--enable-clientapi` is set).

### DIFF
--- a/contrib/docker/builder/Dockerfile
+++ b/contrib/docker/builder/Dockerfile
@@ -9,7 +9,7 @@ RUN apt-get update && apt-get install -y \
     autoconf automake bsdmainutils ccache cmake curl g++ g++-mingw-w64-x86-64 gcc gcc-mingw-w64-x86-64 git \
     libboost-all-dev libbz2-dev libcap-dev libdb4.8-dev libdb4.8++-dev libevent-dev libminiupnpc-dev libprotobuf-dev \
     libqrencode-dev libssl-dev libtool libzmq3-dev make pkg-config protobuf-compiler python-pip qtbase5-dev \
-    qttools5-dev-tools python3-zmq
+    qttools5-dev-tools python3-zmq minizip
 
 RUN pip install ez_setup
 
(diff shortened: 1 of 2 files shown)


contrib/docker/builder/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN apt-get update && apt-get install -y \
     autoconf automake bsdmainutils ccache cmake curl g++ g++-mingw-w64-x86-64 gcc gcc-mingw-w64-x86-64 git \
     libboost-all-dev libbz2-dev libcap-dev libdb4.8-dev libdb4.8++-dev libevent-dev libminiupnpc-dev libprotobuf-dev \
     libqrencode-dev libssl-dev libtool libzmq3-dev make pkg-config protobuf-compiler python-pip qtbase5-dev \
-    qttools5-dev-tools python3-zmq
+    qttools5-dev-tools python3-zmq libminizip-dev
```
thebevrishot: Ubuntu 14.04 have no `libminizip-dev`.

riordant: ok, I'm trying the Ubuntu 14 version. After I will squash the commits.

thebevrishot: Please dont squash after someone push a review that make it review more difficult.

riordant:
@thebevrishot It seems that minizip has been deprecated, from what I can tell it used to be in Trusty. As a result, we have two options:
- Update Gitian to Bionic now (will probably take some time)
- Update later, don't enable `clientapi` for CI right now
Thoughts? @ultimaweapon @psolstice

riordant:
Relevant commit: https://github.com/bitcoin/bitcoin/pull/13171

contrib/docker/builder/Dockerfile
riordant: Understand, but I also don't want to add a lot of small commits for the same feature. Anyhow, don't think we can push this

PR #776, opened 2019-12-01
Multi-core optimization to sigma creation/verification

psolstice:
## PR intention
Optimize sigma spend creation and verification by using multiple CPU cores if available

## Code changes brief
Added simple thread pool implementation. Changed `MultiExponent::get_multiple()` function to use multiple threads

### DIFF
--- a/src/secp256k1/include/MultiExponent.h
+++ b/src/secp256k1/include/MultiExponent.h
@@ -16,6 +16,8 @@ public:
     GroupElement get_multiple();
 
 private:
+    GroupElement get_multiple_single_thread(int start_point, int point_count);
+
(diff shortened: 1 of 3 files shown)


a-bezrukov:
It lacks auto testing, maybe one comparing 2-3 values received with ST mult with MT mult.

reubenyap:
After meeting today, decided not to merge this into 13.8.10 as @psolstice is still benchmarking various parameters and the performance gain due to Sigma still being relatively new is not significant at this point in time.

Will be merged in next release.

reubenyap:
KIV this as needs to be modified for Lelantus after refactor.

PR #780, opened 2019-12-03
Fixed the wallet loading

### LINKED ISSUE #778: Tx list is garbled when wallet loaded
## Actual behaviour
If my wallet has Sigma spends and I open my wallet when Sigma is disabled, I can see that the tx list is not what it was before. Tx times have changed, tx order has changed

## Desired behaviour
Tx list is not changed.

## Steps to reproduce
- regtest, mine 400, mint, spend and mint again
- mine up to 511, close wallet, open - the tx list will be garbled.

a-bezrukov:
Fixes #778

### DIFF
--- a/src/sigma.cpp
+++ b/src/sigma.cpp
@@ -185,12 +185,20 @@ bool CheckSigmaSpendTransaction(
         uint256 hashTx,
         bool isVerifyDB,
         int nHeight,
+        int nRealHeight,
         bool isCheckWallet,
         bool fStatefulSigmaCheck,
         CSigmaTxInfo *sigmaTxInfo) {
     bool hasSigmaSpendInputs = false, hasNonSigmaInputs = false;
     int vinIndex = -1;
     std::unordered_set<Scalar, sigma::CScalarHash> txSerials;
+
+    Consensus::Params const & params = ::Params().GetConsensus();
+
+    if(!isVerifyDB && !isCheckWallet) {
+        if(nRealHeight >= params.nDisableUnpaddedSigmaBlock && nRealHeight < params.nSigmaPaddingBlock)
+             return state.DoS(100, error("Sigma is disabled at this period."));
+    }
 
     for (const CTxIn &txin : tx.vin)
     {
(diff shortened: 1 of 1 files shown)


src/sigma.cpp
```diff
@@ -185,13 +185,21 @@ bool CheckSigmaSpendTransaction(
...
         bool isCheckWallet,
         bool fStatefulSigmaCheck,
         CSigmaTxInfo *sigmaTxInfo) {
     bool hasSigmaSpendInputs = false, hasNonSigmaInputs = false;
     int vinIndex = -1;
     std::unordered_set<Scalar, sigma::CScalarHash> txSerials;
 
+    Consensus::Params const & params = ::Params().GetConsensus();
+
+    if(!isVerifyDB && !isCheckWallet) {
```
riordant: Just noticed that `isVerifyDB` is always passed as false to this function. Is it necessary to keep it?

src/sigma.cpp
```diff
@@ -421,15 +428,14 @@ bool CheckSigmaTransaction(
         realHeight = chainActive.Height();
     }
 
-    if(tx.IsSigmaSpend() && realHeight >= consensus.nDisableUnpaddedSigmaBlock && realHeight < consensus.nSigmaPaddingBlock)
-        return state.DoS(100, error("Sigma is bisabled at this period."));
+    bool const allowSigma = (realHeight >= consensus.nSigmaStartBlock);
```
riordant: Shouldn't we just return immediately if Sigma is not allowed? (I understand that this code already existed)

src/sigma.cpp
levonpetrosyan93: it is code cleanup, better do it with another PR I think.

a-bezrukov: It was my initial implementation but I was asked to make the change smaller.

src/sigma.cpp
a-bezrukov: We may be checking the wallet transactions or the block DB transactions. These checks should be done on Sigma disabled block heights.

PR #781, opened 2019-12-03
Fix wallet crash when the wallet initiates reindex

### LINKED ISSUE #779: Wallet crashes when blockchain DB load fails
## Actual behaviour
If blockchain DB load error occures, the wallet crashes, and reindexing is initiated by the wallet.

a-bezrukov:
Fixes #779

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1668,7 +1668,6 @@ bool AppInit2(boost::thread_group &threadGroup, CScheduler &scheduler) {
                 delete pcoinsdbview;
                 delete pcoinscatcher;
                 delete pblocktree;
-                sigma::CSigmaState::GetState()->Reset();
 
                 pblocktree = new CBlockTreeDB(nBlockTreeDBCache, false, fReindex);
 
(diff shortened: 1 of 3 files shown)


src/init.cpp
```diff
@@ -1668,6 +1668,7 @@ bool AppInit2(boost::thread_group &threadGroup, CScheduler &scheduler) {
                 delete pcoinsdbview;
                 delete pcoinscatcher;
                 delete pblocktree;
+                sigma::CSigmaState::GetState()->Reset();
```
levonpetrosyan93: Wold it be more logical to move this to UnloadBlockIndex() function?

a-bezrukov: There is no logical evidence that moving this into UnloadBlockIndex will make the program any better as UnloadBlockIndex is called exactly once 4 lines above. 
It maybe more aesthetic to put it there. That's why I am doing so.

levonpetrosyan93: Saying logical, I mean all such calls are at UnloadBlockIndex.

PR #786, opened 2019-12-10
Fixed tor library picking wrong openssl library on MacOS

psolstice:
## PR intention
Latest `brew` version deprecated OpenSSL 1.0.* but `tor` was unable to pick up the new version

## Code changes brief
Forced every dependency (`tor` and `secp256k1`) to use the same version that our code is using by adding `--with-openssl-dir` option


psolstice:
Check on all platforms is required (including gitian builds) before merging!

Issue #760, opened 2019-10-30
dumpwallet command does not dump Zerocoin data


italicbold:
I have been waiting for https://github.com/zcoinofficial/zcoin/issues/541 to be in latest version of Zcoin client. However when I try dump my old wallet to get my zcoins into a new hd wallet the command crashes zcoin client.

## Client
Zcoin Core version v0.13.8.5-dc33dff (64-bit)

## Actual behaviour
Zcoin client crashes, empty file produced by dumpwallet. Nothing appended to debug.log

## Desired behaviour
Zcoin client does not crash and dumps zcoins with wallet

## Steps to reproduce
walletpassphrase "xxx" 120
dumpwallet "d:\wallet.txt" 
dumpwallet "d:\wallet.txt" xxx

italicbold:
Folowing steps from https://zcoin.io/how-to-upgrade-your-legacy-zcoin-wallet-to-a-hd-wallet/

italicbold:
Works fine in Zcoin v0.13.8.4-7a38885, but does not dump the zero coins.

riordant:
Hi, from the steps you've shown, you're on Windows correct?

italicbold:
Yes thats correct, Windows 10 version 1903.

italicbold:
Just tested, same behavior on Ubuntu 16.04.6 LTS

riordant:
Ive discovered the issue. Will push a fix soon

riordant:
https://github.com/zcoinofficial/zcoin/pull/761 is the fix, we will do an interim release with it. Thanks

italicbold:
thanks for the prompt turnaround!

italicbold:
Hi, dumpwallet no longer crashes, however when I try and import tha dumped wallet the client displays an error and gets stuck on 2% pemenantly:

![Capture](https://user-images.githubusercontent.com/22705442/70298441-55fb0b00-17fa-11ea-9c92-b2b3c16a9407.JPG)

Console outputs **error: stoi (code -1)**

# Wallet dump created by Zcoin v0.13.8.9-beb6440
# * Created on 2019-12-05T17:12:13Z
# * Best block at time of backup was 220908 

Zcoin v0.13.8.9-beb6440 used for dump and import on windows 10.

Not sure if this issue is related to incorrect dump or import wallet, let me know if you want me to raise a new issue thanks.

riordant:
[...]

PR #787, opened 2019-12-10
Fix issue with old Zerocoin import

riordant:
## PR intention
Fixes an issue with importing Zerocoins from before the ECDSA key was introduced.

## Code changes brief
On import, the string containing Zerocoin data is split into `n` elements. The last of these elements is a `#`, which is a delimiter for the next line. This extra element was not taken into account when attempting to import the ECDSA key and `IsUsedForRemint` boolean - old Zerocoins would try to import this values, leading to a bad access.


reubenyap:
Solves #760

PR #797, opened 2019-12-17
Ban zcoin nodes with version < 0.13.8.8

justanwar:
## PR intention
Ban versions of Zcoin client 0.13.8.5 and below

## Code changes brief
"Parse standard client version string in form of /Satosh:0.x.y.z/ and drop connection with every legacy node" - As per #742

### DIFF
--- a/src/version.h
+++ b/src/version.h
@@ -49,6 +49,6 @@ static const int SHORT_IDS_BLOCKS_VERSION = 90013;
 static const int INVALID_CB_NO_BAN_VERSION = 90013;
 
 //! minimum version of official client to connect to
-static const int MIN_ZCOIN_CLIENT_VERSION = 130805; // 0.13.8.5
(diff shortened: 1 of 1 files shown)


src/version.h
```diff
@@ -49,6 +49,6 @@ static const int SHORT_IDS_BLOCKS_VERSION = 90013;
 static const int INVALID_CB_NO_BAN_VERSION = 90013;
 
 //! minimum version of official client to connect to
-static const int MIN_ZCOIN_CLIENT_VERSION = 130802; // 0.13.8.2
+static const int MIN_ZCOIN_CLIENT_VERSION = 130805; // 0.13.8.5
```
psolstice: I think we need to set it to the first version that use sigma padding (including private releases!).

justanwar: I'm trying to locate which private version sigma padding was released in, if that fails it looks like we have to go with v0.13.8.8

PR #798, opened 2019-12-17
Add backend for new GUI (clientapi)

riordant:
## PR intention
Add `clientapi`, the backend component for the new rich GUI.

## Code changes brief
- main additions: `clientapi/` and `zmqserver/` modules (isolated from main code)
- adding clientapi preprocessor directive (configure option) `--enable-clientapi`, and conf setting `clientapi={0/1}`. Unless these are explicitly enabled, the previous two modules are disabled.
- require `libminizip` and `libzmq` dependancies under the directive (CI includes these dependancies and builds with the directive)
- add `minizip` package to `depends` packages
- `clientapi` architecture and documentation under docs and `clientapi/` directory
- add tests
- Extend `validationinterface` to include new functions

#### Znodes:
- add getters and setters for each entry of the Znode class, and a function to determine if this Znode belongs to the wallet
- add `nRank` entry for Znode class (in-memory, not written to DB)
- add `GetBlockchainSynced` function: internally calls `IsBlockchainSynced` as usual, but allows us to monitor sync changes
- Lock Znode collateral in `AddToWallet` (so if we are syncing/reindexing, it will lock as it is found on-chain)
#### other:
- Add the ability to create a Sigma spend transaction without signing, for the purposes of getting fee before wallet unlock. This is done is a similar way to how it's done for normal send.
- Fix a "subtract fee from amount" issue


src/bitcoind.cpp
```diff
@@ -118,6 +122,15 @@ bool AppInit(int argc, char* argv[])
             fprintf(stderr, "Error: %s\n", e.what());
             return false;
         }
+    bool fApi = false;
```
akshaynexus: This boolean is not used anywhere

akshaynexus: nvm my bad

reubenyap:
Decision to keep both repos separate.

Issue #762, opened 2019-11-01
Sigma "Available amount to mint" should exclude locked spends.


azilber:
## Actual behaviour
When XZC is locked for znode use, the Sigma screen does not take the locked spends into account.

## Desired behaviour
Should deduct the znode amount(s) from the available amount to Sigma mint.

## Steps to reproduce
*Step 1* Setup Znode
*Step 2* Enter Sigma mint

catchingknives:
@azilber You mean under `Available amount to mint`?

reubenyap:
Hi @levonpetrosyan93 can you look into this? I agree that this is misleading and should be an easy fix. Could you work on this after mnemonics?

reubenyap:
Solved by #799

PR #799, opened 2019-12-17
Exclude locked coins from Available to mint amount

levonpetrosyan93:
Implementation of #762


levonpetrosyan93:
Thank you Tadhg, will fix first, about sencond, I think it is not due to my changes, but will try to fix it also.

PR #803, opened 2019-12-20
Add test coverage to CI

riordant:
Partial fix for https://github.com/zcoinofficial/zcoin/issues/681 (yet to be added to https://codecov.io )

Will fail as it needs the two dependancies added to the Dockerfile.

As we are upgrading to Bionic for building, we should probably add the new Dockerfile along with the changes from that PR at the same time. TBD

## PR intention
Add test coverage to our CI, modifying the builder Dockerfile and Jenkins.

## Code changes brief
- `Jenkinsfile`: `--enable-lcov` in `./configure`, `make cov` instead of `make check` (performs `make check` then creates test coverage report)
- `Dockerfile`: Add necessary deps (`lcov` and `default-jre`)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,15 +15,15 @@ pipeline {
                 sh 'mkdir -p dist'
                 sh 'tar -C dist --strip-components=1 -xzf zcoin-*.tar.gz'
                 dir('dist') {
-                    sh './configure --enable-elysium --enable-tests'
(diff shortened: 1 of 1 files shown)


riordant:
requires https://github.com/zcoinofficial/zcoin/pull/860 to pass.

ultimaweapon:
@reubenyap could you remove me from reviewers? I can't do it by my self.

PR #806, opened 2019-12-23
Upgrade to Bionic

a-bezrukov:
## PR intention
Set Ubuntu 18.04 "Bionic" as the build engine for the Gitian builder

## Code changes brief
Most of the update is straight-forward as it was ported from the upstream repository.


configure.ac
```diff
@@ -55,7 +55,7 @@ case $host in
   ;;
 esac
 dnl Require C++11 compiler (no GNU extensions)
-AX_CXX_COMPILE_STDCXX([11], [noext], [mandatory])
+AX_CXX_COMPILE_STDCXX([11], [noext], [mandatory], [nodefault])
```
psolstice: For this change to work you need to update the respective m4 file for that function
https://github.com/bitcoin/bitcoin/blob/master/build-aux/m4/ax_cxx_compile_stdcxx.m4
Probable the entire `build-aux` directory is worth updating

src/tor/configure.ac
```diff
@@ -584,7 +586,6 @@ AC_CHECK_FUNCS(
 	backtrace \
 	backtrace_symbols_fd \
 	eventfd \
-	explicit_bzero \
```
psolstice: Source for this change? Will it be in the tor upstream someday?

configure.ac
a-bezrukov: I'd better remove this nodefault. Updating the whole dir would end up in merge conflicts later

src/tor/configure.ac
a-bezrukov: Please read the comment few lines up for the source of change.

a-bezrukov: > explicit_bzero was removed as it breaks compat with GLIBC 2.21

Explicit_bzero was popping up in the export lists and it was bound to glibc 2.25, which is too new for many distros

psolstice: I saw that. The question is did you take it from somewhere and is there a chance it'll appear in official tor sources? Because it's likely to be overwritten by subsequent tor upgrades

a-bezrukov: I removed it myself. I think the likelihood it'll appear in the parent repo is quite low as they definitely don't need it be compatible with glibc 2.21

reubenyap:
To merge only after 13.8.10 release.

contrib/gitian-build.sh
```diff
@@ -1,453 +1,93 @@
-#!/usr/bin/env bash
-# Copyright (c) 2016 The Bitcoin Core developers
-# Distributed under the MIT software license, see the accompanying
-# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#!/bin/bash
```
psolstice: Where does this come from? We are replacing bitcoin version with something entirely else. Why?

a-bezrukov: This is the script we are using now for releases.

a-bezrukov: Reverted as it's better to leave it as is.

psolstice: Ok. I think that if you want to add a script you can always add a new file without re-purposing the existing one. Also bitcoin changed shell script here to python in latest versions

PR #817, opened 2020-01-28
Fix exodus sigma transactions representation on QT

thebevrishot:
## PR intention
Fix the problem which transaction tables show invalid type and amount as shown in
https://github.com/zcoinofficial/zcoin/issues/634

## Code changes brief
Add specific handle code for each exodus sigma transaction types
- Override amount when detect exodus sigma spend
- Override amount when detect exodus sigma mint
- Override amount when detect exodus denomination creation to show `N/A`

Add missing type mapping in transaction type function `shrinkTxType`

closes https://github.com/zcoinofficial/zcoin/issues/634


thebevrishot:
Exodus_mintspend

src/qt/txhistorydialog.cpp
```diff
@@ -330,6 +354,12 @@ int TXHistoryDialog::PopulateHistoryMap()
             s_stolistdb->getRecipients(txHash, "", &receiveArray, &tmpAmount, &stoFee);
             displayAmount = FormatShortMP(mp_obj.getProperty(), tmpAmount) + getTokenLabel(mp_obj.getProperty());
         }
+
+        // override - hide display amount for denomination creation transactions
+        if (type == EXODUS_TYPE_CREATE_DENOMINATION) {
```
ultimaweapon: I think we should merge this with line 344 so it more maintainable.

src/qt/txhistorydialog.cpp
```diff
@@ -301,6 +301,27 @@ int TXHistoryDialog::PopulateHistoryMap()
         uint64_t amountNew = 0;
         htxo.valid = getValidMPTX(txHash, &tmpBlock, &type, &amountNew);
         if (htxo.valid && type == EXODUS_TYPE_TRADE_OFFER && amountNew > 0) amount = amountNew; // override for when amount for sale has been auto-adjusted
+
+        if (htxo.valid && type == EXODUS_TYPE_SIMPLE_SPEND) { // override amount for spend
+            auto denomination = mp_obj.getDenomination();
+            std::array<uint8_t, 1> denoms = {denomination};
+            amount = SumDenominationsValue(mp_obj.getProperty(), denoms.begin(), denoms.end());
```
ultimaweapon: Can we make `CMPTransaction::getAmount` to supports spend instead?

src/qt/txhistorydialog.cpp
```diff
@@ -301,6 +301,27 @@ int TXHistoryDialog::PopulateHistoryMap()
...
+            std::vector<uint8_t> denoms;
+            std::transform(
+                mints.begin(),
+                mints.end(),
+                std::back_inserter(denoms),
+                [](std::pair<uint8_t, exodus::SigmaPublicKey> mint) -> uint8_t {
+                    return mint.first;
+                });
+
+            amount = SumDenominationsValue(mp_obj.getProperty(), denoms.begin(), denoms.end());
```
ultimaweapon: Same here.

src/exodus/tx.h
```diff
@@ -269,6 +269,24 @@ class CMPTransaction
...
+        std::vector<uint8_t> denoms;
+        std::transform(
+            mints.begin(),
+            mints.end(),
+            std::back_inserter(denoms),
+            [] (std::pair<uint8_t, exodus::SigmaPublicKey> mint) -> uint8_t {
+                return mint.first;
+            });
+
+        return SumDenominationsValue(getProperty(), denoms.begin(), denoms.end());
```
ultimaweapon: We can eliminate the temporary `denoms` by using `boost::make_transform_iterator`.

src/exodus/tx.h
```diff
@@ -269,6 +269,24 @@ class CMPTransaction
...
+    }
+
+    CAmount getSpendAmount() const {
+        std::vector<uint8_t> denoms = {getDenomination()};
```
ultimaweapon: Better to use `std::array` to improve performance with 10X.

src/qt/txhistorydialog.cpp
```diff
@@ -238,6 +238,11 @@ int TXHistoryDialog::PopulateHistoryMap()
                 pending.type == EXODUS_TYPE_METADEX_CANCEL_ECOSYSTEM || pending.type == EXODUS_TYPE_SEND_ALL) {
                 htxo.amount = "N/A";
             }
+
+            if (pending.type == EXODUS_TYPE_SIMPLE_SPEND) {
+                htxo.amount = FormatShortMP(pending.prop, pending.amount) + getTokenLabel(pending.prop);
```
ultimaweapon: Do we still need this once we fix the previous issue?

thebevrishot: This solves the problem that the amount of exodus spend transactions are shown as negative when the transaction is pending.

src/qt/txhistorydialog.cpp
```diff
@@ -352,6 +352,11 @@ int TXHistoryDialog::PopulateHistoryMap()
             displayAmount = FormatShortMP(mp_obj.getProperty(), tmpAmount) + getTokenLabel(mp_obj.getProperty());
         }
 
+        // Override - sigma spend address
+        if (type == EXODUS_TYPE_SIMPLE_SPEND) {
+            htxo.address = "Spend";
+        }
```
ultimaweapon: Better to merge this with line 323 so it will be the only place that modifies the address.

src/qt/txhistorydialog.cpp
ultimaweapon: After a discussion, we need to fix this based on who is the receiver.

src/qt/txhistorydialog.cpp
```diff
@@ -324,6 +326,13 @@ int TXHistoryDialog::PopulateHistoryMap()
         if (!IsMyAddress(mp_obj.getSender())) htxo.address = mp_obj.getReceiver();
         if (htxo.fundsMoved && IsMyAddress(mp_obj.getSender())) displayAmount = "-" + displayAmount;
 
+        if (type == EXODUS_TYPE_SIMPLE_SPEND) {
+            htxo.address = "Spend";
+            if (htxo.fundsMoved && !IsMyAddress(mp_obj.getReceiver())) {
+                displayAmount = "-" + displayAmount;
+            }
+        }
```
ultimaweapon: Better to merge this with the above statements.

thebevrishot: The problem is displayAmount is define and initialize at line 320

PR #824, opened 2020-02-21
Rename exodus to elysium

thebevrishot:
## PR intention
Rename exodus to elysium

## Code changes brief
- Rename directory src/exodus to src/elysium
- Rename exodus in qt
- Rename exodus rpcs
- Update rpc tests


src/elysium/elysium.cpp
```diff
@@ -2597,8 +2597,8 @@ void CMPTxList::LoadActivations(int blockHeight)
     CheckLiveActivations(blockHeight);
 
     // This alert never expires as long as custom activations are used
-    if (mapArgs.count("-exodusactivationallowsender") || mapArgs.count("-exodusactivationignoresender")) {
-        AddAlert("exodus", ALERT_CLIENT_VERSION_EXPIRY, std::numeric_limits<uint32_t>::max(),
+    if (mapArgs.count("-elysiumactivationallowsender") || mapArgs.count("-elysiumactivationignoresender")) {
+        AddAlert("elysium", ALERT_CLIENT_VERSION_EXPIRY, std::numeric_limits<uint32_t>::max(),
```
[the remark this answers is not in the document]
thebevrishot: It should be safe to change because the function `AddAlerts` writes nothing permanently. And I already change any related "exodus" string value.

src/elysium/createpayload.cpp
```diff
@@ -634,5 +635,37 @@ std::vector<unsigned char> CreatePayload_SimpleSpend(
...
+    elysium::swapByteOrder(groupSize);
+
+    PUSH_BACK_BYTES(payload, messageVer);
+    PUSH_BACK_BYTES(payload, messageType);
+    PUSH_BACK_BYTES(payload, propertyId);
+    PUSH_BACK_BYTES(payload, denomination);
+    PUSH_BACK_BYTES(payload, group);
+    PUSH_BACK_BYTES(payload, groupSize);
+
+    payload.insert(payload.end(), pubkey.begin(), pubkey.end());
```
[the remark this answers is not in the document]
thebevrishot: After personal discussion we can't share the logic.

src/elysium/notifications.cpp
```diff
@@ -67,22 +67,22 @@ void AddAlert(const std::string& sender, uint16_t alertType, uint32_t alertExpir
     newAlert.alert_message = alertMessage;
 
     // very basic sanity checks for broadcast alerts to catch malformed packets
-    if (sender != "exodus" && (alertType < ALERT_BLOCK_EXPIRY || alertType > ALERT_CLIENT_VERSION_EXPIRY)) {
+    if (sender != "elysium" && (alertType < ALERT_BLOCK_EXPIRY || alertType > ALERT_CLIENT_VERSION_EXPIRY)) {
```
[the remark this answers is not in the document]
thebevrishot: I already change the function `AddAlert` calling points. I believe this line is safe to change. But it would be good if you could do double-check.

src/wallet/walletdb.h
```diff
@@ -298,119 +298,201 @@ class CWalletDB : public CDB
     bool WriteHDChain(const CHDChain& chain);
     bool WriteMnemonic(const MnemonicContainer& mnContainer);
 
-#ifdef ENABLE_EXODUS
+#ifdef ENABLE_ELYSIUM
 
+private:
+    template<typename K, typename V, typename InsertF>
+    void ListElysiumMints(string const &type, InsertF insertF)
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/rpc/client.cpp
```diff
@@ -143,136 +143,140 @@ static const CRPCConvertParam vRPCConvertParams[] =
...
+	{ "elysium_createpayload_revoke", 0 },
+	{ "elysium_createpayload_changeissuer", 0 },
+	{ "elysium_createpayload_trade", 0 },
+	{ "elysium_createpayload_trade", 2 },
+	{ "elysium_createpayload_canceltradesbyprice", 0 },
+	{ "elysium_createpayload_canceltradesbyprice", 2 },
+	{ "elysium_createpayload_canceltradesbypair", 0 },
+	{ "elysium_createpayload_canceltradesbypair", 1 },
+	{ "elysium_createpayload_cancelalltrades", 0 },
+	{ "elysium_createpayload_createdenomination", 0},
```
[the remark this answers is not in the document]
thebevrishot: Removed

src/qt/walletview.cpp
```diff
@@ -387,27 +387,28 @@ void WalletView::gotoBitcoinHistoryTab()
...
-    gotoExodusHistoryTab();
-    exodusTransactionsView->focusTransaction(txid);
+    gotoElysiumHistoryTab();
+    elysiumTransactionsView->focusTransaction(txid);
 }
 #endif
 
 void WalletView::focusBitcoinHistoryTab(const QModelIndex &idx)
 {
+    std::cout << "focusBitcoinHistoryTab" << std::endl;
```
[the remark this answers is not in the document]
thebevrishot: Nice catch!

src/qt/lookupspdialog.cpp
```diff
@@ -323,7 +323,7 @@ void LookupSPDialog::updateDisplayedProperty()
     }
     else
     {
-       if (propertyId == 1) { tokenLabel = " EXODUS"; } else { tokenLabel = " TEXODUS"; }
+       if (propertyId == 1) { tokenLabel = " Elysium"; } else { tokenLabel = " TElysium"; }
```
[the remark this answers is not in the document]
thebevrishot: Nice catch again!!!!

src/elysium/elysium.cpp
```diff
@@ -1990,15 +1990,15 @@ int exodus_init()
             boost::filesystem::path tradePath = GetDataDir() / "MP_tradelist";
             boost::filesystem::path spPath = GetDataDir() / "MP_spinfo";
             boost::filesystem::path stoPath = GetDataDir() / "MP_stolist";
-            boost::filesystem::path exodusTXDBPath = GetDataDir() / "Exodus_TXDB";
-            boost::filesystem::path feesPath = GetDataDir() / "EXODUS_feecache";
-            boost::filesystem::path feeHistoryPath = GetDataDir() / "EXODUS_feehistory";
+            boost::filesystem::path elysiumTXDBPath = GetDataDir() / "Exodus_TXDB";
+            boost::filesystem::path feesPath = GetDataDir() / "Exodus_feecache";
```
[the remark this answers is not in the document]
thebevrishot: I try to change to back to Elysium but in the wrong case. Thanks for nice catch again.

src/elysium/elysium.cpp
```diff
@@ -2199,9 +2199,9 @@ bool exodus_handler_tx(const CTransaction& tx, int nBlock, unsigned int idx, con
  * @param nDataSize The length of the payload
  * @return True, if Class C is enabled and the payload is small enough
  */
-bool exodus::UseEncodingClassC(size_t nDataSize)
+bool elysium::UseEncodingClassC(size_t nDataSize)
 {
-    size_t nTotalSize = nDataSize + magic.size(); // Marker "exodus"
+    size_t nTotalSize = nDataSize + magic.size(); // Marker "elysium"
```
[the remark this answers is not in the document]
thebevrishot: Correct

src/elysium/createpayload.h
```diff
@@ -1,6 +1,7 @@
-#ifndef ZCOIN_EXODUS_CREATEPAYLOAD_H
-#define ZCOIN_EXODUS_CREATEPAYLOAD_H
+#ifndef ZCOIN_ELYSIUM_CREATEPAYLOAD_H
+#define ZCOIN_ELYSIUM_CREATEPAYLOAD_H
 
+#include "coinsigner.h"
```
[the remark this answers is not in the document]
thebevrishot: Correct

src/elysium/tx.h
```diff
@@ -135,11 +135,14 @@ class CMPTransaction
...
-    std::vector<std::pair<uint8_t, exodus::SigmaPublicKey>> mints;
+    std::vector<std::pair<uint8_t, elysium::SigmaPublicKey>> mints;
     uint8_t denomination;
     uint32_t group;
     uint16_t groupSize;
-    std::unique_ptr<exodus::SigmaProof> spend;
+    std::unique_ptr<elysium::SigmaProof> spend;
+
+    std::array<uint8_t, 33> sigmaECDSAPubkey;
+    std::array<uint8_t, 64> sigmaECDSASignature;
```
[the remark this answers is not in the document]
thebevrishot: I believe `ecdsaSignature` is more proper. the name `signature` is too generic.

src/elysium/tx.h
```diff
@@ -135,11 +135,14 @@ class CMPTransaction
...
     SigmaStatus sigmaStatus;
-    std::vector<std::pair<uint8_t, exodus::SigmaPublicKey>> mints;
+    std::vector<std::pair<uint8_t, elysium::SigmaPublicKey>> mints;
     uint8_t denomination;
     uint32_t group;
     uint16_t groupSize;
-    std::unique_ptr<exodus::SigmaProof> spend;
+    std::unique_ptr<elysium::SigmaProof> spend;
+
+    std::array<uint8_t, 33> sigmaECDSAPubkey;
```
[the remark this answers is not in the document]
thebevrishot: same reason with `signature`

src/elysium/tx.h
```diff
@@ -240,6 +243,7 @@ class CMPTransaction
     uint64_t getFeePaid() const { return tx_fee_paid; }
     const std::string& getSender() const { return sender; }
     const std::string& getReceiver() const { return receiver; }
+    boost::optional<CAmount> getReferenceAmount() const { return referenceAmount; }
```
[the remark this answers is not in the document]
thebevrishot: make sense

src/elysium/tx.cpp
```diff
@@ -856,9 +856,21 @@ bool CMPTransaction::interpret_SimpleSpend()
     );
 
     spend.reset(new SigmaProof(DefaultSigmaParams));
-
     try {
-        serialized >> *spend;
+        if (version == MP_TX_PKT_V1) {
```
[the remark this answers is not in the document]
thebevrishot: Sorry, not sure what you mean.

src/elysium/coinsigner.h
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2020 The Zcoin Core Developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ZCOIN_ELYSIUM_COINSIGNER_H
+#define ZCOIN_ELYSIUM_COINSIGNER_H
+
+#include "hash.h"
```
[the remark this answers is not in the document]
thebevrishot: agree

src/elysium/wallet.h
```diff
@@ -47,25 +57,44 @@ class Wallet
     void ClearAllChainState();
 
     SigmaSpend CreateSigmaSpend(PropertyId property, SigmaDenomination denomination, bool fPadding);
+    SigmaSpend CreateLegacySigmaSpend(PropertyId property, SigmaDenomination denomination, bool fPadding);
```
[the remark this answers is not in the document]
thebevrishot: Make sense

thebevrishot:
@ultimaweapon the unused static variable warning is suppressed. Please review.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -268,26 +268,24 @@ def compare_mints(self, expected, actual):
...
             return []
 
-        return self.nodes[0].generate(required_block - current_block)
-
-    def create_default_property(self, name, sigma = True, amount = None, address = None):
-        if address is None:
-            address = self.addrs[0]
+        return node.generate(required_block - current_block)
 
+    def create_default_property(self, name, node, address, sigma = True, amount = None):
```
ultimaweapon: Same as the previous issue.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -268,26 +268,24 @@ def compare_mints(self, expected, actual):
...
         sigma_status = 1 if sigma else 0
 
         if amount is None:
-            self.nodes[0].elysium_sendissuancemanaged(address, 1, 1, 0, '', '', name, '', '', sigma_status)
+            node.elysium_sendissuancemanaged(address, 1, 1, 0, '', '', name, '', '', sigma_status)
```
ultimaweapon: Same here.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -268,26 +268,24 @@ def compare_mints(self, expected, actual):
...
         else:
-            self.nodes[0].elysium_sendissuancefixed(address, 1, 1, 0, '', '', name, '', '', amount, sigma_status)
+            node.elysium_sendissuancefixed(address, 1, 1, 0, '', '', name, '', '', amount, sigma_status)
```
ultimaweapon: Same here.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -268,26 +268,24 @@ def compare_mints(self, expected, actual):
...
 
-        self.nodes[0].generate(1)
+        node.generate(1)
```
ultimaweapon: Same here.

src/elysium/tx.h
```diff
@@ -14,6 +15,8 @@ class CTransaction;
 #include "../uint256.h"
 #include "../utilstrencodings.h"
 
+#include <secp256k1/include/Scalar.h>
```
ultimaweapon: The right way to include this one is just `#include <Scalar.h>` due to it is an external library.

src/elysium/tx.cpp
```diff
@@ -856,19 +856,25 @@ bool CMPTransaction::interpret_SimpleSpend()
...
 
     spend.reset(new SigmaProof(DefaultSigmaParams));
+    serial.reset(new secp_primitives::Scalar());
     try {
         if (version == MP_TX_PKT_V1) {
-            serialized.read(reinterpret_cast<char*>(sigmaECDSAPubkey.data()), sizeof(sigmaECDSAPubkey));
-            serialized >> spend->proof;
-            serialized.read(reinterpret_cast<char*>(sigmaECDSASignature.data()), sizeof(sigmaECDSASignature));
+            std::array<uint8_t, 33> pubkeyBuffer;
+            serialized.read(reinterpret_cast<char*>(pubkeyBuffer.data()), sizeof(pubkeyBuffer));
```
ultimaweapon: `sizeof` on a class is dangerous.

src/sigma/sigmaplus_proof.h
```diff
@@ -51,6 +51,20 @@ class SigmaPlusProof {
...
+            m == other.m &&
+            B_ == other.B_ &&
+            r1Proof_ == other.r1Proof_ &&
+            Gk_ == other.Gk_ &&
+            z_ == other.z_;
+    }
+
+    bool operator!=(const SigmaPlusProof& other) const {
+        return !(*this == other);
+    }
```
ultimaweapon: IMO this one should come after the constructors and before the class-specific methods.

src/sigma/r1_proof.h
```diff
@@ -56,6 +56,20 @@ class R1Proof {
...
+            C_ == other.C_ &&
+            D_ == other.D_ &&
+            f_ == other.f_ &&
+            ZA_ == other.ZA_ &&
+            ZC_ == other.ZC_;
+    }
+
+    bool operator!=(R1Proof const &other) const {
+        return !(*this == other);
+    }
```
ultimaweapon: Same as the previous issue.

src/elysium/createpayload.cpp
```diff
@@ -629,17 +630,23 @@ std::vector<unsigned char> CreatePayload_SimpleSpend(
...
 
+// V1
 std::vector<unsigned char> CreatePayload_SimpleSpend(
     uint32_t propertyId, uint8_t denomination, uint32_t group,
     uint16_t groupSize, elysium::SigmaProof const &proof,
-    std::array<uint8_t, 64> const &signature, std::array<uint8_t, 33> const &pubkey)
+    Signature const &signature, CPubKey const &pubkey)
 {
+    if (pubkey.size() != CPubKey::COMPRESSED_PUBLIC_KEY_SIZE) {
+        throw std::runtime_error("Publickey size is invalid");
```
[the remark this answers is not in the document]
thebevrishot: Make sense

src/elysium/tx.h
thebevrishot: Removed

src/elysium/wallet.h
```diff
@@ -46,26 +56,45 @@ class Wallet
...
     boost::optional<SigmaMint> GetSpendableSigmaMint(
-        PropertyId property, SigmaDenomination denomination);
+        PropertyId property, SigmaDenomination denomination, CoinVersion version);
     void SetSigmaMintChainState(const SigmaMintId &id, const SigmaMintChainState &state);
 
+    SigmaWallet& GetMintWallet(CoinVersion version);
+    SigmaWallet& GetMintWallet(SigmaMintId const &id);
+    SigmaWallet& GetMintWallet(secp_primitives::Scalar &serial);
+
+    bool HasSigmaMint(const SigmaMintId& id, CoinVersion &version);
```
ultimaweapon: This one is quite hard to use. My idea is removing the last argument and make it returns `boost::optional<CoinVersion>` instead.

src/elysium/wallet.h
```diff
@@ -46,26 +56,45 @@ class Wallet
...
+    bool HasSigmaMint(const secp_primitives::Scalar &scalar, CoinVersion &version);
```
ultimaweapon: Same here.

src/elysium/wallet.cpp
```diff
@@ -172,26 +196,82 @@ boost::optional<SigmaMint> Wallet::GetSpendableSigmaMint(PropertyId property, Si
...
 
+SigmaWallet& Wallet::GetMintWallet(CoinVersion version)
+{
+    switch (version) {
+    case CoinVersion::SigmaV0:
+        return mintWalletV0;
+    case CoinVersion::SigmaV1:
+        return mintWalletV1;
+    default:
+        throw new std::runtime_error("Coin version is not found.");
```
ultimaweapon: This one should be `std::invalid_argument` instead.

src/elysium/wallet.cpp
```diff
@@ -172,26 +196,82 @@ boost::optional<SigmaMint> Wallet::GetSpendableSigmaMint(PropertyId property, Si
...
+}
+
+SigmaWallet& Wallet::GetMintWallet(SigmaMintId const &id)
+{
+    CoinVersion version;
+    if (HasSigmaMint(id, version)) {
+        return GetMintWallet(version);
+    }
+
+    throw std::runtime_error("Sigma Mint Id is not found.");
```
ultimaweapon: Same here.

src/elysium/wallet.h
```diff
@@ -46,26 +56,45 @@ class Wallet
...
 
 protected:
     boost::optional<SigmaMint> GetSpendableSigmaMint(
-        PropertyId property, SigmaDenomination denomination);
+        PropertyId property, SigmaDenomination denomination, CoinVersion version);
     void SetSigmaMintChainState(const SigmaMintId &id, const SigmaMintChainState &state);
 
+    SigmaWallet& GetMintWallet(CoinVersion version);
+    SigmaWallet& GetMintWallet(SigmaMintId const &id);
+    SigmaWallet& GetMintWallet(secp_primitives::Scalar &serial);
```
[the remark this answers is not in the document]
thebevrishot: Removed

src/elysium/wallet.h
thebevrishot: Good idea

src/elysium/wallet.h
```diff
@@ -17,10 +19,18 @@
...
 
-namespace exodus {
+namespace elysium {
 
 class Wallet
 {
+protected:
+    enum class CoinVersion : uint8_t
+    {
+        Unknown,
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/elysium/sigmawallet.cpp
```diff
@@ -503,4 +472,17 @@ bool SigmaWallet::RemoveFromMintPool(SigmaPublicKey const &publicKey)
     return false;
 }
 
-} // exodus
+// Helper
+std::unique_ptr<CWalletDB> SigmaWallet::EnsureDBConnection(CWalletDB* &db) const
```
ultimaweapon: This is quite dangerous due to if the caller does not declare a variable to hold the lifetime it will be broken.

src/elysium/sigmawallet.h
```diff
@@ -78,31 +80,49 @@ class SigmaWallet
...
     // Generator
-private:
+protected:
     uint32_t GenerateNewSeed(CKeyID &seedId, uint512 &seed);
     uint32_t GenerateSeed(CKeyID const &seedId, uint512 &seed);
     uint32_t GetSeedIndex(CKeyID const &seedId);
 
 protected:
-    SigmaPrivateKey GeneratePrivateKey(uint512 const &seed);
+    virtual uint32_t ChangeIndex() = 0;
```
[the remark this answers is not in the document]
thebevrishot: Agree

src/elysium/sigmawallet.h
```diff
@@ -78,31 +80,49 @@ class SigmaWallet
...
+
+    virtual bool WriteElysiumMintId(uint160 const &hash, SigmaMintId const &mintId, CWalletDB *db = nullptr) = 0;
+    virtual bool ReadElysiumMintId(uint160 const &hash, SigmaMintId &mintId, CWalletDB *db = nullptr) const = 0;
+    virtual bool EraseElysiumMintId(uint160 const &hash, CWalletDB *db = nullptr) = 0;
+    virtual bool HasElysiumMintId(uint160 const &hash, CWalletDB *db = nullptr) const = 0;
+
+    virtual bool WriteElysiumMintPool(std::vector<MintPoolEntry> const &mints, CWalletDB *db = nullptr) = 0;
+    virtual bool ReadElysiumMintPool(std::vector<MintPoolEntry> &mints, CWalletDB *db = nullptr) = 0;
+
+    virtual void ListElysiumMints(std::function<void(SigmaMintId&, SigmaMint&)>, CWalletDB *db = nullptr) = 0;
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/elysium/sigmawallet.cpp
thebevrishot: Nice catch

src/elysium/sigmawallet.h
```diff
@@ -67,14 +67,55 @@ class SigmaWallet
...
+            bool mustDelete;
+
+        public:
+            DBDeleter(bool mustDelete);
+
+        public:
+            void operator()(CWalletDB* db);
+        };
+
+        std::unique_ptr<CWalletDB, DBDeleter> EnsureDBConnection(CWalletDB *db) const;
```
ultimaweapon: After a discussion we should create a new class to wrap a `CWalletDB` instance to be able to select a user privided or the default one.

src/elysium/sigmawallet.h
```diff
@@ -67,14 +67,55 @@ class SigmaWallet
         >
     > MintPool;
 
+protected:
+    class WalletDB {
```
ultimaweapon: IMO just `Database` looks better and less confuse too.

src/elysium/sigmawallet.cpp
```diff
@@ -45,8 +45,9 @@ bool SigmaWallet::MintPoolEntry::operator!=(MintPoolEntry const &another) const
     return !(*this == another);
 }
 
-SigmaWallet::SigmaWallet() : walletFile(pwalletMain->strWalletFile)
+SigmaWallet::SigmaWallet(WalletDB *db) : walletFile(pwalletMain->strWalletFile)
 {
+    walletDB.reset(db);
```
ultimaweapon: Better to invoke it constructor instead of `reset`.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+    std::array<uint8_t, 32> ecdsaKeyOut;
+    return GeneratePrivateKey(seed, ecdsaKeyOut);
+}
+
+SigmaPrivateKey SigmaWalletV1::GeneratePrivateKey(
+    uint512 const &seed, ECDSAPrivateKey &ecdsaKeyOut)
+{
+    // first 32 bytes as seed of ecdsa key and serial
+    std::array<uint8_t, 32> tmp;
```
ultimaweapon: Better to use `ECDSAPrivateKey` so it will be matched with `ecdsaKeyOut`.

ultimaweapon: Thi one use only inside the loop so better to move it there.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+    std::copy(seed.begin(), seed.begin() + 32, ecdsaKeyOut.begin());
+
+    // hash until get valid private key
+    secp256k1_pubkey pubkey;
+    do {
+        std::copy(ecdsaKeyOut.begin(), ecdsaKeyOut.end(), tmp.begin());
+        CSHA256().Write(tmp.begin(), tmp.size()).Finalize(ecdsaKeyOut.begin());
```
ultimaweapon: Better to use `tmp.data()` instead of `tmp.begin()` so it matched with the purpose.

ultimaweapon: Same for `ecdsaKeyOut`.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+namespace elysium {
+
+SigmaWalletV1::SigmaWalletV1()
+    : SigmaWallet(new SigmaWalletV1::Database())
+{
+}
+
+bool SigmaWalletV1::GeneratePublicKey(ECDSAPrivateKey const &priv, secp256k1_pubkey &out)
+{
+    return secp256k1_ec_pubkey_create(
```
ultimaweapon: Better to use `secp256k1_ec_pubkey_create() == 1` to explicitly convert to `bool`.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+    // first 32 bytes as seed of ecdsa key and serial
+    std::array<uint8_t, 32> tmp;
+    std::copy(seed.begin(), seed.begin() + 32, ecdsaKeyOut.begin());
+
+    // hash until get valid private key
+    secp256k1_pubkey pubkey;
+    do {
+        std::copy(ecdsaKeyOut.begin(), ecdsaKeyOut.end(), tmp.begin());
+        CSHA256().Write(tmp.begin(), tmp.size()).Finalize(ecdsaKeyOut.begin());
+    } while(!GeneratePublicKey(ecdsaKeyOut, pubkey));
```
ultimaweapon: Space error.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    secp_primitives::Scalar serial;
+    GenerateSerial(pubkey, serial);
+
+    // last 32 bytes as seed of randomness
+    std::array<uint8_t, 32> randomnessSeed;
+    std::copy(seed.begin() + 32, seed.end(), randomnessSeed.data());
+    secp_primitives::Scalar randomness;
+    randomness.memberFromSeed(randomnessSeed.begin());
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+bool SigmaWalletV1::GeneratePublicKey(ECDSAPrivateKey const &priv, secp256k1_pubkey &out)
+{
+    return secp256k1_ec_pubkey_create(
+        OpenSSLContext::get_context(),
+        &out,
+        priv.data());
+}
+
+void SigmaWalletV1::GenerateSerial(secp256k1_pubkey const &pubkey, secp_primitives::Scalar &serial)
```
ultimaweapon: Better to return a `secp_primitives::Scalar` directly so it is more easy to use.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+    std::array<uint8_t, 33> compressedPub;
+
+    size_t outSize = sizeof(compressedPub);
+    secp256k1_ec_pubkey_serialize(
+        OpenSSLContext::get_context(),
```
ultimaweapon: We need to update this once we created a dedicated context.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+        compressedPub.begin(),
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+        OpenSSLContext::get_context(),
+        &out,
+        priv.data());
+}
+
+void SigmaWalletV1::GenerateSerial(secp256k1_pubkey const &pubkey, secp_primitives::Scalar &serial)
+{
+    std::array<uint8_t, 33> compressedPub;
+
+    size_t outSize = sizeof(compressedPub);
```
ultimaweapon: `sizeof` on the class is dangerous.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+        &outSize,
+        &pubkey,
+        SECP256K1_EC_COMPRESSED);
+
+    if (outSize != 33) {
+        throw std::runtime_error("Compressed public key size is invalid.");
+    }
+
+    std::array<uint8_t, CSHA256::OUTPUT_SIZE> hash;
+    CSHA256().Write(compressedPub.begin(), sizeof(compressedPub)).Finalize(hash.data());
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    serial.memberFromSeed(hash.begin());
```
ultimaweapon: Same here.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+}
+
+uint32_t SigmaWalletV1::BIP44ChangeIndex() const
+{
+    return BIP44_ELYSIUM_MINT_INDEX_V1;
+}
+
+SigmaPrivateKey SigmaWalletV1::GeneratePrivateKey(uint512 const &seed)
+{
+    std::array<uint8_t, 32> ecdsaKeyOut;
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+    randomness.memberFromSeed(randomnessSeed.begin());
+
+    return SigmaPrivateKey(serial, randomness);
+}
+
+CoinSigner SigmaWalletV1::GetSigner(SigmaMintId const &id)
+{
+    auto mint = GetMint(id);
+    uint512 seed;
+    std::array<uint8_t, 32> ecdsaKey;
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.h
```diff
@@ -0,0 +1,60 @@
...
+
+#ifndef ZCOIN_ELYSIUM_SIGMAWALLETV1_H
+#define ZCOIN_ELYSIUM_SIGMAWALLETV1_H
+
+#include "coinsigner.h"
+#include "sigmawallet.h"
+
+namespace elysium {
+
+typedef std::array<uint8_t, 32> ECDSAPrivateKey;
```
ultimaweapon: Better to move inside `SigmaWalletV1`.

src/elysium/sigmawalletv1.h
```diff
@@ -0,0 +1,60 @@
...
+    SigmaWalletV1();
+
+protected:
+    bool GeneratePublicKey(ECDSAPrivateKey const &priv, secp256k1_pubkey &out);
+    void GenerateSerial(secp256k1_pubkey const &pubkey, secp_primitives::Scalar &serial);
+
+protected:
+    uint32_t BIP44ChangeIndex() const;
+    SigmaPrivateKey GeneratePrivateKey(uint512 const &seed);
+    SigmaPrivateKey GeneratePrivateKey(uint512 const &seed, ECDSAPrivateKey &ecdsaKeyOut);
```
ultimaweapon: Same as the previous issue.

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2020 The Zcoin Core Developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "sigmawalletv1.h"
+
+#include "../libzerocoin/Zerocoin.h"
+#include "../sigma/openssl_context.h"
```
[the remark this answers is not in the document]
thebevrishot: created

src/elysium/sigmawalletv1.cpp
```diff
@@ -0,0 +1,173 @@
...
+}
+
+SigmaPrivateKey SigmaWalletV1::GeneratePrivateKey(uint512 const &seed)
+{
+    std::array<uint8_t, 32> ecdsaKeyOut;
+    return GeneratePrivateKey(seed, ecdsaKeyOut);
+}
+
+SigmaPrivateKey SigmaWalletV1::GeneratePrivateKey(
+    uint512 const &seed, ECDSAPrivateKey &ecdsaKeyOut)
```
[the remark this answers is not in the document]
thebevrishot: `signatureKey` is quite weird for me. just `key` would be better.

src/elysium/sigmawalletv1.cpp
thebevrishot: removed

src/elysium/sigmawalletv1.cpp
thebevrishot: make sense

src/elysium/sigmawalletv1.cpp
thebevrishot: good idea

src/elysium/coinsigner.h
```diff
@@ -0,0 +1,33 @@
...
+#include "ecdsa_context.h"
+#include "signature.h"
+#include "key.h"
+#include "pubkey.h"
+
+#include <array>
+
+namespace elysium {
+
+class CoinSigner
```
ultimaweapon: Seems like this class is tried to do signing abstraction but it is still tied with `CKey` and `CPubKey`, which is belong to ECDSA with secp256k1. So we have two choices here:

1. Make it right.
2. Remove this abstraction and use `CKey` directly.

For me the second choice is better for now due to it is less likely to that we will have more signing algorithm in the future.

src/elysium/wallet.cpp
```diff
@@ -237,9 +313,12 @@ void Wallet::OnMintAdded(
...
         // 2. try to recover new mint
-        if (mintWallet.TryRecoverMint(
-            id, SigmaMintChainState(block, group, idx)
-        )) {
+        SigmaMintChainState chainState(block, group, idx);
+        if (mintWalletV0.TryRecoverMint(id, chainState)) {
+            LogPrintf("%s : Found new legacy mint when try to recover\n", __func__);
+        }
+
+        if (mintWalletV1.TryRecoverMint(id, chainState)) {
```
ultimaweapon: The mint should not belongs to more than one wallet. So better to chain this condition with the above condition.

src/elysium/wallet.cpp
```diff
@@ -172,26 +194,78 @@ boost::optional<SigmaMint> Wallet::GetSpendableSigmaMint(PropertyId property, Si
 
 SigmaPrivateKey Wallet::GetKey(const SigmaMint &mint)
 {
-    return mintWallet.GeneratePrivateKey(mint.seedId);
+    // Try all mint wallets
+    try {
+        return mintWalletV1.GeneratePrivateKey(mint.seedId);
```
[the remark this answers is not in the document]
thebevrishot: It should throw when the seed id is not found in file `src/elysium/sigmawallet.cpp` line 88 - 91

src/elysium/wallet.cpp
thebevrishot: Make sense

src/elysium/wallet.cpp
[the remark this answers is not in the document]
ultimaweapon: I don't think it will throw because all seeds come from the main wallet.

src/elysium/signature.cpp
```diff
@@ -0,0 +1,81 @@
...
+            &(this->signature),
+            signature)) {
+            valid = true;
+        }
+    } else {
+        throw std::invalid_argument("Signature encoding type is not supported");
+    }
+}
+
+std::vector<unsigned char> Signature::GetCompact() const
```
ultimaweapon: Can we use `std::array` instead?

src/elysium/signature.cpp
```diff
@@ -0,0 +1,81 @@
...
+        context.Get(),
+        result.data(),
+        &signature)) {
+        throw std::runtime_error("Serialized size is in valid");
+    }
+
+    return result;
+}
+
+std::vector<unsigned char> Signature::GetDER() const
```
ultimaweapon: Same here.

src/elysium/signature.h
```diff
@@ -0,0 +1,73 @@
...
+        s.write(reinterpret_cast<char*>(buffer.data()), buffer.size());
+    }
+
+    /** unserialize compact
+     */
+    template<typename Stream>
+    void Unserialize(Stream& s, int nType, int nVersion)
+    {
+        std::array<uint8_t, SIGNATURE_COMPACT_SERIALIZED_SIZE> buffer;
+        s.read(reinterpret_cast<char*>(buffer.begin()), sizeof(buffer));
```
ultimaweapon: Same as the previous issue.

src/elysium/signature.h
```diff
@@ -0,0 +1,73 @@
...
+        if (1 != secp256k1_ecdsa_signature_parse_compact(
+            context.Get(),
+            &signature,
+            reinterpret_cast<const unsigned char*>(buffer.begin()))) {
```
ultimaweapon: Same here.

src/elysium/signature.cpp
```diff
@@ -0,0 +1,81 @@
...
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "signature.h"
+
+#include <stdexcept>
+
+namespace elysium {
+
+Signature::Signature() : valid(false), context(ECDSAContext::CreateSignContext())
```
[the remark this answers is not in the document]
thebevrishot: We already set valid flag as false to identify signature be not initialized yet.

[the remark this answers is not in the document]
thebevrishot: In the secp256k1 library that this class passes the context through don't use information from it but I use the signing mode following `CKey` class.

src/elysium/signature.cpp
thebevrishot: We can, but users of this class such as `CPubKey` accept vectors of unsigned char directly. Then I think using vector is a better idea.

src/elysium/signature.cpp
thebevrishot: Because we don't use an array with GetDER then I think we should not use with this too.

src/elysium/ecdsa_signature.h
```diff
@@ -54,20 +49,25 @@ class Signature
     template<typename Stream>
     void Unserialize(Stream& s, int nType, int nVersion)
     {
-        std::array<uint8_t, SIGNATURE_COMPACT_SERIALIZED_SIZE> buffer;
-        s.read(reinterpret_cast<char*>(buffer.begin()), sizeof(buffer));
+        std::array<uint8_t, COMPACT_SIZE> buffer;
+        s.read(reinterpret_cast<char*>(buffer.data()), sizeof(buffer));
```
ultimaweapon: `sizeof` on a class is dangerous.

src/elysium/signature.cpp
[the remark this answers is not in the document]
ultimaweapon: But this class is just a signature so it should follow `CPubKey` instead, is not it?

src/elysium/signaturebuilder.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // serialize payload
+    CKeyID keyId;
+    if (!receiver.GetKeyID(keyId)) {
+        throw std::runtime_error("Fail to get address key id.");
+    }
+
+    hasher.write(reinterpret_cast<char*>(keyId.begin()), keyId.size());
+
+    // reference amount
+    elysium::swapByteOrder(referenceAmount);
```
ultimaweapon: Why do we need to switch endianness?

src/elysium/signaturebuilder.cpp
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2020 The Zcoin Core Developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "convert.h"
```
ultimaweapon: We should not need this if we don't switch endianness.

src/elysium/signaturebuilder.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "sigmaprimitives.h"
```
ultimaweapon: Same here if we don't sign a proof.

src/elysium/signaturebuilder.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "signaturebuilder.h"
+
+#include "../base58.h"
+
+#include <secp256k1/include/Scalar.h>
```
ultimaweapon: Seems like we don't need this.

src/elysium/signaturebuilder.h
```diff
@@ -0,0 +1,47 @@
...
+#include "../key.h"
+
+#include <secp256k1/include/Scalar.h>
+
+namespace elysium {
+
+class SignatureBuilder
+{
+public:
+    virtual ECDSASignature Sign(CKey &signer) = 0;
```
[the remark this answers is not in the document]
thebevrishot: The class hold `CHashWriter` which would be mutate when get result from it.

src/elysium/signaturebuilder.h
```diff
@@ -0,0 +1,47 @@
...
+    virtual bool Verify(ECDSASignature const &signature) = 0;
```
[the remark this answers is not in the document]
thebevrishot: Sorry, which one?

src/elysium/signaturebuilder.h
```diff
@@ -0,0 +1,47 @@
...
+{
+protected:
+    CHashWriter hasher;
+    CPubKey publicKey;
+
+public:
+    SigmaV1SignatureBuilder(
+        CBitcoinAddress const &receiver,
+        int64_t referenceAmount,
+        SigmaProof const &proof,
```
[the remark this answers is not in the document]
thebevrishot: After the personal discussion, we would follow sigma.

src/elysium/signaturebuilder.h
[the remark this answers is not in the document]
thebevrishot: Sorry, Just found your comment that could resolve the problem. Awesome!!

src/elysium/signaturebuilder.cpp
thebevrishot: The endianness would be the same with datas that we store to payload.

src/elysium/signature.cpp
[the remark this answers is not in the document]
thebevrishot: Actually, the context should be pass depending on user purposes. Then we should let callers pass it when this class needs. Except serialize function we would serialize only if we sign and deserialize when we need to verify.

src/elysium/sigmawallet.cpp
```diff
@@ -129,13 +134,11 @@ uint32_t SigmaWallet::GetSeedIndex(CKeyID const &seedId)
     // parse last index
     uint32_t addressIndex;
```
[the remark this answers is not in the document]
thebevrishot: Nice catch

src/elysium/txprocessor.cpp
```diff
@@ -154,18 +155,43 @@ int TxProcessor::ProcessSimpleSpend(const CMPTransaction& tx)
...
     auto denomination = tx.getDenomination();
     auto group = tx.getGroup();
     auto groupSize = tx.getGroupSize();
 
     bool const fPadding = block >= ::Params().GetConsensus().nSigmaPaddingBlock;
 
     assert(spend);
+    assert(serial);
+
+    if (!serial->isMember()) {
```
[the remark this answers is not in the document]
thebevrishot: Nice catch again!!!

src/elysium/signaturebuilder.h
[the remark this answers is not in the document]
thebevrishot: Agree, I forget to update when changing to key instead of the signer.

src/elysium/signaturebuilder.h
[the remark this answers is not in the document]
thebevrishot: Sound good.

src/elysium/signaturebuilder.h
```diff
@@ -0,0 +1,47 @@
...
+{
+public:
+    virtual ECDSASignature Sign(CKey &signer) = 0;
+    virtual bool Verify(ECDSASignature const &signature) = 0;
+};
+
+class SigmaV1SignatureBuilder : SignatureBuilder
+{
+protected:
+    CHashWriter hasher;
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/elysium/signaturebuilder.cpp
```diff
@@ -0,0 +1,60 @@
...
+SigmaV1SignatureBuilder::SigmaV1SignatureBuilder(
+    CBitcoinAddress const &receiver,
+    int64_t referenceAmount,
+    SigmaProof const &proof,
+    CPubKey const &publicKey)
+    : hasher(CHashWriter(SER_GETHASH, PROTOCOL_VERSION)), publicKey(publicKey)
+{
+    // serialize payload
+    CKeyID keyId;
+    if (!receiver.GetKeyID(keyId)) {
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/elysium/rpctx.cpp
```diff
@@ -1709,25 +1710,45 @@ UniValue exodus_sendspend(const UniValue& params, bool fHelp)
...
-        auto spend = wallet->CreateSigmaSpend(propertyId, denomination, fPadding);
-        mint = spend.mint;
+        auto spend = wallet->CreateSigmaSpendV1(propertyId, denomination, fPadding);
+        auto key = wallet->GetSigmaSignatureKey(spend.mint);
+        auto pubkey = key.GetPubKey();
 
+        SigmaV1SignatureBuilder sigBuilder(address, referenceAmount, spend.proof);
+        auto signature = sigBuilder.Sign(key);
+
+        SigmaV1SignatureBuilder sigVerifier(address, referenceAmount, spend.proof);
```
ultimaweapon: Why we don't use `sigBuilder` instead?

thebevrishot: I name it following the role.

thebevrishot: Sorry, seem like I misunderstood. Fixed!

src/elysium/ecdsa_signature.cpp
```diff
@@ -17,32 +17,39 @@ ECDSASignature::ECDSASignature(secp256k1_ecdsa_signature const &sig)
 {
 }
 
-ECDSASignature::ECDSASignature(ECDSAContext const &context, unsigned char const *signature, size_t len)
-    : valid(false)
+ECDSASignature ECDSASignature::Parse(ECDSAContext const &context, unsigned char const *signature, size_t len)
 {
+    ECDSASignature sig;
```
ultimaweapon: Why not use `secp256k1_ecdsa_signature` instead? With this way we don't need to update `valid` flag by ourself.

src/elysium/ecdsa_signature.cpp
```diff
@@ -17,32 +17,39 @@ ECDSASignature::ECDSASignature(secp256k1_ecdsa_signature const &sig)
...
     if (len >= 70 && len <= 72) {
         if (1 == secp256k1_ecdsa_signature_parse_der(
             context.Get(),
-            &(this->signature),
+            &(sig.signature),
             signature,
             len))
         {
-            valid = true;
+            sig.valid = true;
```
ultimaweapon: I think we should throw instead. It looks weird that this method can return an invalid signature.

src/elysium/ecdsa_signature.cpp
```diff
@@ -17,32 +17,39 @@ ECDSASignature::ECDSASignature(secp256k1_ecdsa_signature const &sig)
...
         }
     } else if (len == COMPACT_SIZE) {
         if (1 == secp256k1_ecdsa_signature_parse_compact(
             context.Get(),
-            &(this->signature),
+            &(sig.signature),
             signature)) {
-            valid = true;
+            sig.valid = true;
```
ultimaweapon: Same here.

src/elysium/test/sigmawalletv1_tests.cpp
```diff
@@ -0,0 +1,422 @@
...
+    });
+
+    BOOST_CHECK_EQUAL(0, counter);
+}
+
+BOOST_AUTO_TEST_CASE(listelysiummints_withsomemints)
+{
+    std::vector<std::pair<SigmaMintId, SigmaMint>> mints;
+    for (size_t i = 0; i < 10; i++)
+    {
```
ultimaweapon: Brace error.

src/elysium/test/sigmawalletv0_tests.cpp
```diff
@@ -0,0 +1,363 @@
...
+    std::vector<MintPoolEntry> mintPool;
+
+    PopulateMintEntries(3, 0, 10, std::back_inserter(mintPool));
+
+    wallet->WriteMintPool(mintPool);
+
+    std::vector<MintPoolEntry> data;
+    BOOST_CHECK_EQUAL(true, wallet->ReadMintPool(data));
+    BOOST_CHECK(mintPool == data);
+    BOOST_CHECK(std::is_permutation(mintPool.begin(), mintPool.end(), data.begin()));
```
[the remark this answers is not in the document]
thebevrishot: To check the equality of two collections ignoring the order.

src/elysium/test/sigmawalletv0_tests.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // Proxy
+    using SigmaWalletV0::GeneratePrivateKey;
+    SigmaPrivateKey GeneratePrivateKey(uint512 const &seed)
+    {
+        return SigmaWalletV0::GeneratePrivateKey(seed);
+    }
+
+    bool WriteMint(SigmaMintId const &id, SigmaMint const &mint)
+    {
+        return database->WriteMint(id, mint);
```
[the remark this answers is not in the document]
thebevrishot: Those methods are just wrappers of the database. I think it would be better to test it directly then.

src/elysium/test/sigmawalletv1_tests.cpp
```diff
@@ -0,0 +1,422 @@
...
+    std::vector<MintPoolEntry> mintPool;
+
+    PopulateMintEntries(3, 0, 10, std::back_inserter(mintPool));
+
+    wallet->WriteMintPool(mintPool);
+
+    std::vector<MintPoolEntry> data;
+    BOOST_CHECK_EQUAL(true, wallet->ReadMintPool(data));
+    BOOST_CHECK(mintPool == data);
+    BOOST_CHECK(std::is_permutation(mintPool.begin(), mintPool.end(), data.begin()));
```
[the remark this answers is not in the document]
thebevrishot: same here

src/elysium/test/sigmawalletv0_tests.cpp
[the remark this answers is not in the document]
ultimaweapon: Its already checked by the above line. But I already approved so no needs to remove it right now. Just remove it if there is a request change from the other reviewers.

src/elysium/txprocessor.cpp
```diff
@@ -154,18 +155,53 @@ int TxProcessor::ProcessSimpleSpend(const CMPTransaction& tx)
...
     auto group = tx.getGroup();
     auto groupSize = tx.getGroupSize();
 
     bool const fPadding = block >= ::Params().GetConsensus().nSigmaPaddingBlock;
 
     assert(spend);
+    assert(serial);
+
+    if (IsFeatureActivated(FEATURE_SIGMA_SPENDV1, block)) {
+        if (!serial->isMember()) {
```
levonpetrosyan93, line 169: We need to add `|| serial->isZero()`  here

levonpetrosyan93:
We can call `!pub.IsValid()` here
https://github.com/thebevrishot/zcoin/blob/5d0c31afa53e14e107cf79dbe9381899f23b6b27/src/elysium/sigmadb.cpp#L547

levonpetrosyan93:
We need to call `g.isInfinity()` here
https://github.com/thebevrishot/zcoin/blob/5d0c31afa53e14e107cf79dbe9381899f23b6b27/src/elysium/sigmaprimitives.cpp#L29

levonpetrosyan93:
We need to check` isZero()` here for `serial` and `randomness`
https://github.com/thebevrishot/zcoin/blob/5d0c31afa53e14e107cf79dbe9381899f23b6b27/src/elysium/sigmaprimitives.cpp#L70

levonpetrosyan93:
also this one, check `.. && !commitment.isInfinity()`
https://github.com/thebevrishot/zcoin/blob/5d0c31afa53e14e107cf79dbe9381899f23b6b27/src/elysium/sigmaprimitives.cpp#L104

src/elysium/test/sigmawallet_tests.cpp
```diff
@@ -102,6 +109,98 @@ class TestSigmaWallet : public SigmaWallet
...
+            auto local = Connection(db);
+            return local->ReadElysiumMintPoolV0(mints);
+        }
+
+        void ListMints(std::function<void(SigmaMintId&, SigmaMint&)> const &inserter, CWalletDB *db = nullptr)
+        {
+            auto local = Connection(db);
+            local->ListElysiumMintsV0<SigmaMintId, SigmaMint>(inserter);
+        }
+    };
```
levonpetrosyan93, line 203: why are we copying above logic here?

thebevrishot: The purpose of this is testing the abstract parent class of wallet v0 and v1, you can think that logic above is mocking functions to test the main functionality. Actually, the better way to do is defining separated functions instead of relied on DB functions of v0 but we have a lot of things to do if we need it. Anyway, I think we could keep it like this and add a todo comment to remove DB dependency when we have to change the DB functions of v0.

thebevrishot: The purpose of this is to test the abstract parent class of wallet v0 and v1, then we need to define all undefined functions of the class. Most logics are the same with v0 but the reason that I don't test through v0 because I need v0 to be able to change without worrying about the parent class and the tests.

src/elysium/txprocessor.cpp
thebevrishot: @levonpetrosyan93 After discussing with @ultimaweapon we should remove this check because we should let the protocol layer deal with this instead of rejecting in consensus layer. How do you think?

thebevrishot: To be more clearer we would let function `VerifySigmaSpend` handle it.

levonpetrosyan93: but looks  have no such checks at protocol layer,

ultimaweapon: Why we do not make it have that check at the cryptographic level?

thebevrishot: @ultimaweapon if you mean lib sigma, the interface doesn't accept serial we use it before passing to the library(function `Verify` in file src/elysium/sigmaprimitives.h). Anyway, after research and discussion with @levonpetrosyan93 we agree to delete the check here.

levonpetrosyan93: At Sigma we have check for serial at CoinSpend when it is checked to be the hash of pubkey, I think it is enough, but at elysium code I could not find such check.

thebevrishot: In Elysium we generate serial from pubkey here https://github.com/thebevrishot/zcoin/blob/rename-exodus/src/elysium/tx.cpp#L875

src/elysium/rules.cpp
```diff
@@ -131,6 +132,7 @@ CMainConsensusParams::CMainConsensusParams()
 
     // Sigma releated
     SIGMA_FEATURE_BLOCK = 212000; // 4 Nov 2019
+    SIGMA_SPENDV1_FEATURE_BLOCK = 258177; // 13 April 2020
```
ultimaweapon, line 135: We need to update this due to a block number is already passed this value.

src/elysium/sigmaprimitives.cpp
```diff
@@ -101,7 +102,7 @@ bool SigmaPublicKey::operator!=(const SigmaPublicKey& other) const
 
 bool SigmaPublicKey::IsValid() const
 {
-    return commitment.isMember();
+    return commitment.isMember() && !commitment.isInfinity();
```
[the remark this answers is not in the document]
levonpetrosyan93: @ultimaweapon to be sure that there is no such commitment, please run a reindex with this code. that will also cover the case when both serial and randomness are 0's.

src/elysium/sigmaprimitives.cpp
```diff
@@ -67,7 +67,8 @@ bool SigmaPrivateKey::operator!=(const SigmaPrivateKey& other) const
 
 bool SigmaPrivateKey::IsValid() const
 {
-    return serial.isMember() && randomness.isMember();
+    return serial.isMember() && randomness.isMember() &&
+        !serial.isZero() && !randomness.isZero();
```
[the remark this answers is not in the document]
levonpetrosyan93: actually it is not possible to generate 0 Scalars with previous client without editing code.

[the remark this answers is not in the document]
levonpetrosyan93: also you can put a condition with HF block SIGMA_SPENDV1_FEATURE_BLOCK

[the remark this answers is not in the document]
levonpetrosyan93: actually it will be more secure, @thebevrishot please put such condition

src/elysium/sigmaprimitives.cpp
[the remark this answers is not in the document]
ultimaweapon: Reindex won't help due to we don't know if there is going to be a new one before everyone is upgraded. So the only way is to activate this new check on a certain block instead.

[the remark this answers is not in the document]
levonpetrosyan93: ok, let's do that way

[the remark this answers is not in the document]
thebevrishot: This check would effect anon set. I think we should do following.
1) Reject non-member and inf after v1 activated.
2) Reject non-member old coins from anon group but don't refund for non-member mint transaction because that would effect elysium state. If there is some non-member coin in the set users would not be able to spend any coin then they need to reindex to fix anonimity group state.

src/elysium/sigmaprimitives.cpp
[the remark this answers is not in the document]
thebevrishot: Actually, this class is used in wallet level then I would try to handle it following wallet version by applying zero checking only in the new one.

src/elysium/rules.cpp
thebevrishot: You're right. Per the meeting, I would put it to 4 May 2020.

src/elysium/sigmaprimitives.cpp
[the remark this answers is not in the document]
thebevrishot: After check whole logic there are no non-member mints are not be added to anon set. Then reindexing is not required.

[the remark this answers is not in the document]
levonpetrosyan93: So what we do, we just reject inf commitments after HF block, it is more safe, and exactly there is no need for reindex.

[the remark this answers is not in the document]
ultimaweapon: The non-member check is already there from the beginning. So the only cases we need to handle is the infinity point checking.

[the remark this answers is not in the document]
thebevrishot: @ultimaweapon updated, please check.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
...
+    BOOST_CHECK(sr.IsMember());
+    BOOST_CHECK(sr.IsValid());
+
+    SigmaPrivateKey sinf(s, inf);
+    BOOST_CHECK(sinf.IsMember());
+    BOOST_CHECK(!sinf.IsValid());
+
+    SigmaPrivateKey infr(s, inf);
+    BOOST_CHECK(infr.IsMember());
+    BOOST_CHECK(!infr.IsValid());
```
ultimaweapon, line 60: This test is exactly the same as the above.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
     BOOST_CHECK_NE(key.randomness, randomness);
 }
 
+BOOST_AUTO_TEST_CASE(private_key_inf)
```
ultimaweapon, line 45: IMO better to use a long name so it is easier to understand (e.g. `private_key_infinity`).

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
...
+{
+    Scalar inf;
+    Scalar s(1), r(2);
+
+    SigmaPrivateKey sr(s, r);
```
ultimaweapon, line 50: What is the meaning of `sr`? Better to rename it so it is easier to understand.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
...
+    BOOST_CHECK(sr.IsMember());
+    BOOST_CHECK(sr.IsValid());
+
+    SigmaPrivateKey sinf(s, inf);
```
ultimaweapon, line 54: Same here.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
...
+    BOOST_CHECK(sinf.IsMember());
+    BOOST_CHECK(!sinf.IsValid());
+
+    SigmaPrivateKey infr(s, inf);
```
ultimaweapon, line 58: Same here.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
     BOOST_CHECK_NE(key.randomness, randomness);
 }
 
+BOOST_AUTO_TEST_CASE(private_key_inf)
+{
+    Scalar inf;
```
ultimaweapon, line 47: Namespace error.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -42,6 +42,24 @@ BOOST_AUTO_TEST_CASE(private_key)
...
+    Scalar s(1), r(2);
```
ultimaweapon, line 48: Same here.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -75,6 +93,21 @@ BOOST_AUTO_TEST_CASE(public_key)
     BOOST_CHECK_EQUAL(pub.commitment, commit);
 }
 
+BOOST_AUTO_TEST_CASE(public_key_inf)
```
ultimaweapon, line 96: Same as the previous issue.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -75,6 +93,21 @@ BOOST_AUTO_TEST_CASE(public_key)
...
+    SigmaPrivateKey k;
+    k.Generate();
+
+    SigmaPublicKey infp, pub(k, params);
+
+    BOOST_CHECK(infp.IsMember());
+    BOOST_CHECK(!infp.IsValid());
+
+    BOOST_CHECK(pub.IsMember());
+    BOOST_CHECK(pub.IsValid());
```
ultimaweapon, line 108: This test should not be here. This case should test the infinity point only.

src/elysium/test/sigmaprimitives_tests.cpp
```diff
@@ -75,6 +93,21 @@ BOOST_AUTO_TEST_CASE(public_key)
     BOOST_CHECK_EQUAL(pub.commitment, commit);
 }
 
+BOOST_AUTO_TEST_CASE(public_key_inf)
+{
+    auto& params = DefaultSigmaParams;
+    SigmaPrivateKey k;
+    k.Generate();
+
+    SigmaPublicKey infp, pub(k, params);
```
ultimaweapon, line 102: Better to rename this to just `pub` once we followed my previous comment.

src/elysium/sigmawalletv1.cpp
```diff
@@ -64,21 +64,27 @@ SigmaPrivateKey SigmaWalletV1::GeneratePrivateKey(
...
+        } while (!GetPublicKey(signatureKey, pubkey));
+
+        auto serial = GenerateSerial(pubkey);
 
-    auto serial = GenerateSerial(pubkey);
+        // last 32 bytes as seed of randomness
+        std::array<uint8_t, 32> randomnessSeed;
+        std::copy(seed.begin() + 32, seed.end(), randomnessSeed.begin());
+        secp_primitives::Scalar randomness;
+        randomness.memberFromSeed(randomnessSeed.data());
```
ultimaweapon, line 82: Better to move this out of this loop due to it does not need any data from the loop.

src/elysium/txprocessor.cpp
```diff
@@ -75,7 +75,11 @@ int TxProcessor::ProcessSimpleMint(const CMPTransaction& tx)
         auto denom = mint.first;
         auto& pubkey = mint.second;
 
-        if (!pubkey.IsValid()) {
+        auto isValid = IsFeatureActivated(FEATURE_SIGMA_SPENDV1, block)
```
ultimaweapon, line 78: Better to cache the result of `IsFeatureActivated` outside this loop.

PR #829, opened 2020-03-18
Revise CI on Actions

ultimaweapon:
This PR will change the behavior of CI on GitHub Actions as the following description:

- Enable it to run on PR that targets the `master` branch.
- Build from source distribution tarball instead of the source in the repository to prevent building error when creating a release.
- Create artifacts for the output binaries.


.github/workflows/ci-master.yml
```diff
@@ -3,63 +3,184 @@ on:
...
+      working-directory: ${{ env.SOURCE_ARTIFACT }}
+    - name: Prepare Files for Artifact
+      run: |
+        mkdir -p $ARTIFACT_DIR
+        mv $SOURCE_ARTIFACT/src/{zcoin-cli,zcoin-tx,zcoind,qt/zcoin-qt} $ARTIFACT_DIR
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v1
+      with:
+        name: linux-binaries
+        path: ${{ env.ARTIFACT_DIR }}
```
thebevrishot, line 87: I believe we should archive using tarball before upload to preserve permissions.

.github/workflows/ci-master.yml
```diff
@@ -3,63 +3,184 @@ on:
...
+    - name: Prepare Files for Artifact
+      run: |
+        mkdir -p $ARTIFACT_DIR
+        mv $SOURCE_ARTIFACT/src/{zcoin-cli.exe,zcoin-tx.exe,zcoind.exe,qt/zcoin-qt.exe} $ARTIFACT_DIR
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v1
+      with:
+        name: windows-binaries
+        path: ${{ env.ARTIFACT_DIR }}
+  build-mac:
```
thebevrishot, line 150: It should be macos if we refer to the OS.

.github/workflows/ci-master.yml
```diff
@@ -3,63 +3,184 @@ on:
...
+      working-directory: ${{ env.SOURCE_ARTIFACT }}
+    - name: Prepare Files for Artifact
+      run: |
+        mkdir -p $ARTIFACT_DIR
+        mv $SOURCE_ARTIFACT/src/{zcoin-cli,zcoin-tx,zcoind,qt/zcoin-qt} $ARTIFACT_DIR
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v1
+      with:
+        name: mac-binaries
+        path: ${{ env.ARTIFACT_DIR }}
```
thebevrishot, line 186: Might be we need to do with macos to preserve permission same as above.

.github/workflows/ci-master.yml
ultimaweapon: I think just `mac` is enough. It not ambiguous and shorter to type.

.github/workflows/ci-master.yml
ultimaweapon: Yes, I agreed. But it is also ugly that it needs to extract files twice. We should try to find another solution.

.github/workflows/ci-master.yml
thebevrishot: Hmm, Ok makes sense.

.github/workflows/ci-master.yml
thebevrishot: After investigating, we don't have simple ways to do. And the main purpose of binaries is used for development then I think we should let it be like this for now and create an issue instead.

ultimaweapon: https://github.com/zcoinofficial/zcoin/issues/833

PR #842, opened 2020-05-21
Fix elysium tests fail

thebevrishot:
## PR intention
Fix qa tests fail because of the situation.
1) nodes in network broadcast invalid spork messages and are banned.
2) the network is splitted.
3) when new blocks or txs are produced and some tests have to wait for mempool and blocks of all nodes to be synced then it would timeout.

## Code changes brief
- ~~Prevent invalid spork message to be sent. (`src/spork.cpp`)~~
- Update the Elysium test framework. (`qa/rpc-tests/test_framework/test_framework.py`)


src/spork.cpp
```diff
@@ -84,7 +84,10 @@ bool CSporkManager::UpdateSpork(int nSporkID, int64_t nValue)
 {
     CSporkMessage spork = CSporkMessage(nSporkID, nValue, GetTime());
 
-    spork.Sign(strMasterPrivKey);
+    if (!spork.Sign(strMasterPrivKey)) {
```
psolstice, line 87: Transition to evo depends on some particular spork to be set at block nDIP0003EnforcementHeight. This change basically disables LLMQ.

The plan is to remove all the spork code after mainnet transitioned to evo we need it to stay here. Are you sure the ban is because of incorrect sporks?

thebevrishot: yes, it reaches this line (https://github.com/zcoinofficial/zcoin/blob/core14/src/spork.cpp#L53). My change just prevents messages which contain incorrect signatures to be broadcasted.

psolstice: I changed spork handling. LLMQ doesn't need spork anymore. And I changed `spork.cpp`. Please update your PR

src/spork.cpp
```diff
@@ -84,7 +84,10 @@ bool CSporkManager::UpdateSpork(int nSporkID, int64_t nValue)
...
+        LogPrintf("CSporkManager::UpdateSpork -- fail to generate valid signature\n");
```
akshaynexus, line 88: why not use `return error("CSporkManager::UpdateSpork -- fail to generate valid signature");` here?

src/elysium/ecdsa_signature.h
```diff
@@ -10,6 +10,8 @@
 #include <secp256k1.h>
 
 #include <array>
+#include <cstdint>
+#include <stdexcept>
```
akshaynexus, line 14: any reason for adding this in?

src/spork.cpp
thebevrishot: updated.

src/spork.cpp
thebevrishot: you're right. but I already remove the change. thanks anyway.

src/elysium/ecdsa_signature.h
thebevrishot: nice catch, I forgot to remove `cstdint`.  but `stdexcept` is included because we use `std::logic_error` and `std::runtime_error` in this file. please check the update.

PR #843, opened 2020-05-25
Fixing overview page for non HD wallets

levonpetrosyan93:
For non HD wallets at overview page in place of `Anonymized Balances` it shows garbage, as for that case Sigma is not available and balance for it is not set.
PR checks if it is not HD wallet just disables that labels on qt page.


psolstice (approved):
LGTM

PR #845, opened 2020-05-29
Upgrade to bitcoin core 0.14 and dash deterministic masternodes

psolstice:
## PR intention
This PR brings following changes to the zcoin:
- Bitcoin core upgraded to 0.14 (some PRs from 0.15 are also included)
- Deterministic masternodes are merged from recent dash code

## Code changes brief
- Merges with bitcoin and partly dash were performed. Repository now contains references to original commits
- Cleanups are done to ensure minimal difference with parent repositories. This ensures less problematic upgrade to newer versions of bitcoin
- Legacy dash code coexists with modern one. The former can be removed later when transition to evo znode system is complete
- Bugfixes and refactors


src/validation.cpp
```diff
@@ -29,10 +39,5412 @@
...
+            LogPrintf("Updating spend state from Mempool..");
+            zwalletMain->GetTracker().UpdateSpendStateFromMempool(zcSpendSerialsV3);
+        }
+#endif
+    }
+    if(markZcoinSpendTransactionSerial)
+        sigmaState->AddMintsToMempool(zcMintPubcoinsV3);
+#ifdef ENABLE_WALLET
+    if(tx.IsSigmaMint()){
+        BOOST_FOREACH(const CTxOut &txout, tx.vout)
```
levonpetrosyan93, line 1338: This is not critical thing, but looks we already added pubcoins upper, no need to do this again.

PR #846, opened 2020-05-31
Core14 changes

riordant:
- make CHDMintWallet a member of CWallet (to better handle multiple wallet architecture in core 14)

- Mac test optimisations: 
     - make walletDB object a part of `CHDMintWallet`, pass as reference where possible
     - minimuse file I/O

- some HDMint code cleanup

### DIFF
--- a/src/hdmint/wallet.cpp
+++ b/src/hdmint/wallet.cpp
@@ -175,6 +175,10 @@ void CHDMintWallet::GenerateMintPool(CWalletDB& walletdb, int32_t nIndex)
         LogPrintf("%s : hashSeedMaster=%s hashPubcoin=%s seedId=%d count=%d\n", __func__, hashSeedMaster.GetHex(), hashPubcoin.GetHex(), seedId.GetHex(), nLastCount);
     }
 
+    // write hdchain back to database
+    if (!walletdb.WriteHDChain(pwalletMain->GetHDChain()))
+        throw std::runtime_error(std::string(__func__) + ": Writing HD chain model failed");
+
     // Update local + DB entries for count last generated
     nCountNextGenerate = nLastCount;
     walletdb.WriteMintSeedCount(nCountNextGenerate);
@@ -433,7 +437,7 @@ bool CHDMintWallet::SetMintSeedSeen(CWalletDB& walletdb, std::pair<uint256,MintP
 
     LogPrintf("%s: Adding mint to tracker.. \n", __func__);
     // Add to tracker which also adds to database
-    tracker.Add(dMint, true);
+    tracker.Add(walletdb, dMint, true);
 
     return true;
 }
@@ -526,7 +530,7 @@ bool CHDMintWallet::CreateMintSeed(CWalletDB& walletdb, uint512& mintSeed, const
         int32_t chainIndex = pwalletMain->GetHDChain().nExternalChainCounters[BIP44_MINT_INDEX];
         if(nCount==chainIndex){
             // If chainIndex is the same as n (ie. we are generating next available key), generate a new key.
-            pubKey = pwalletMain->GenerateNewKey(BIP44_MINT_INDEX);
+            pubKey = pwalletMain->GenerateNewKey(BIP44_MINT_INDEX, nWriteChain);
         }
         else if(nCount<chainIndex){
             // if it's less than the current chain index, we are regenerating the mintpool. get the key at n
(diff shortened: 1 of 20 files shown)


src/wallet/wallet.h
```diff
@@ -751,6 +750,8 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     MasterKeyMap mapMasterKeys;
     unsigned int nMasterKeyMaxID;
 
+    CHDMintWallet* zwallet;
```
ultimaweapon, line 753: This will cause memory leak on some places due to it re-assign this without freeing the previos value. Better to to use `std::unique_ptr` instead.

riordant: The value will not be reassigned within the same instance ever, but a `unique_ptr` is better anyway. Updated, please take a look

riordant:
@levonpetrosyan93 @ultimaweapon rebased, some issues so had to make a new commit. please re-review.

src/wallet/wallet.cpp
```diff
@@ -102,7 +102,7 @@ struct CompareByAmount
 
 static void EnsureMintWalletAvailable()
 {
-    if (!zwalletMain) {
+    if (!pwalletMain->zwallet) {
```
psolstice, line 105: I think we need to ensure `pwalletMain` is not null here

src/wallet/wallet.cpp
```diff
@@ -175,7 +175,7 @@ CPubKey CWallet::GetKeyFromKeypath(uint32_t nChange, uint32_t nChild) {
     return pubkey;
 }
 
-CPubKey CWallet::GenerateNewKey(uint32_t nChange)
+CPubKey CWallet::GenerateNewKey(uint32_t nChange, bool nWriteChain)
```
psolstice, line 178: Why do you start boolean variables with `n`? In bitcoin it's `f` to denote variable as a "flag"

src/wallet/rpcwallet.cpp
```diff
@@ -84,7 +84,7 @@ bool EnsureWalletIsAvailable(CWallet * const pwallet, bool avoidException)
 
 void EnsureSigmaWalletIsAvailable()
 {
-    if (!zwalletMain) {
+    if (!pwalletMain->zwallet) {
```
psolstice, line 87: Need to check `pwalletMain` for null

src/hdmint/wallet.cpp
```diff
@@ -521,7 +520,7 @@ CKeyID CHDMintWallet::GetMintSeedID(int32_t nCount){
  * @param seedId (optional) seedId of the key to use for mint generation.
  * @return sucess
  */
-bool CHDMintWallet::CreateMintSeed(uint512& mintSeed, const int32_t& nCount, CKeyID& seedId)
+bool CHDMintWallet::CreateMintSeed(CWalletDB& walletdb, uint512& mintSeed, const int32_t& nCount, CKeyID& seedId, bool nWriteChain)
```
psolstice, line 523: `nWriteChain` -> `fWriteChain`

src/wallet/sigmaspendbuilder.cpp
```diff
@@ -168,13 +168,14 @@ CAmount SigmaSpendBuilder::GetChanges(std::vector<CTxOut>& outputs, CAmount amou
     auto params = sigma::Params::get_default();
 
     CHDMint hdMint;
+    CWalletDB walletdb(pwalletMain->strWalletFile);
```
psolstice, line 171: This is incorrect in multiple wallets scenario. You need to pass `CWalletDB` object reference to `GetChanges`

psolstice: Or you can obtain wallet file name from `CWallet` object and use it to create `CWalletDB`

src/wallet/rpcwallet.cpp
```diff
@@ -84,7 +84,7 @@ bool EnsureWalletIsAvailable(CWallet * const pwallet, bool avoidException)
 
 void EnsureSigmaWalletIsAvailable()
 {
-    if (!pwalletMain->zwallet) {
+    if (!pwalletMain && !pwalletMain->zwallet) {
```
psolstice, line 87: ||, not &&

src/wallet/wallet.cpp
```diff
@@ -102,7 +102,7 @@ struct CompareByAmount
 
 static void EnsureMintWalletAvailable()
 {
-    if (!pwalletMain->zwallet) {
+    if (!pwalletMain && !pwalletMain->zwallet) {
```
psolstice, line 105: && -> ||

PR #850, opened 2020-06-01
Fixes for Znode warning

riordant:
Separating this from https://github.com/zcoinofficial/zcoin/pull/846 for easier review

GUI changes to fix Znode warning. Now shows warning at the overview screen rather than as a popup, given that the warning logic is not executed until znode sync is complete.

### DIFF
--- a/src/qt/notifyznodewarning.cpp
+++ b/src/qt/notifyznodewarning.cpp
@@ -31,7 +31,7 @@ void NotifyZnodeWarning::notify()
         (int)hoursToEnforcement);
 
     SetMiscWarning(strWarning);
-    AlertNotify(strWarning);
+    uiInterface.NotifyAlertChanged();
 }
 
 bool NotifyZnodeWarning::shouldShow()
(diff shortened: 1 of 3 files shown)


src/qt/notifyznodewarning.cpp
```diff
@@ -2,47 +2,46 @@
...
-    msg.setText(messageWarning);
-    msg.setIcon(QMessageBox::Warning);
-    msg.exec();
-    nConsidered = true;
+    std::string strWarning = strprintf(_("WARNING: Legacy znodes detected. You should migrate to the new Znode layout before it becomes enforced (approximately %i days and %i hours). For details on how to migrate, go to https://zcoin.io/znode-migration"),
+        (int)daysToEnforcement,
+        (int)hoursToEnforcement);
+
+    SetMiscWarning(strWarning);
+    AlertNotify(strWarning);
```
psolstice, line 34: This particular call runs a program specified with `-alertnotify` argument and passes an argument of `strWarning` to it. I doubt we need it.

You should just call
```
uiInterface.NotifyAlertChanged();
```

riordant: Done

PR #851, opened 2020-06-02
Fix elysium reindexing tests

thebevrishot:
## PR intention
Fix the problem that block is created and add to the chain before reindex is done on reindexing test by adding deley.

## Code changes brief
- Add deley after sync block in test


thebevrishot:
@riordant not sure why can't reply the comment directly

On the line above delay, we already wait until it reaches destination block using `sync_blocks`. but even the block number is reached, the flag might be unset yet.

PR #856, opened 2020-06-07
Mobile api

levonpetrosyan93:
Merging this will make maintaining our ElectrumX more easy.


src/rpc/misc.cpp
```diff
@@ -949,6 +949,172 @@ UniValue getaddressbalance(const JSONRPCRequest& request)
 
 }
 
+UniValue getanonymityset(const JSONRPCRequest& request)
```
psolstice, line 952: Access to sigma state should be protected by `cs_main`. Applies to all RPC functions introduced in this PR

a-bezrukov: Thanks, fixed

psolstice:
Please squash before merging

PR #860, opened 2020-06-12
fix test coverage generation

riordant:
## PR intention
when compiling with `--enable-lcov` and running `make cov`, a test coverage report is supposed to be generated in the root of the project. This fails in systems where the default python version is python 2, as the test script for `rpc-tests.py` is run with `python` - `rpc-tests.py` was changed some time ago to use python 3 syntax. So simply change this to use `$(PYTHON)`, where it will select one of the Python 3 versions available on the system.


riordant:
@a-bezrukov please re-review thanks

PR #868, opened 2020-06-29
Removed block syncing in elysium reindex test

thebevrishot:
## PR intention
Fix test fails while trying to sync block by removing syncing logic because this test focuses on node state validation after reindexing.

## Code changes brief
- Update elysium_sigma_reindex.py to wait for node state to reach the destination block instead of sync with the whole network.


ultimaweapon:
What is the difference between the old method and the new method?

thebevrishot:
@ultimaweapon I added one-sec delay in the first place because I found new blocks are generated while the reindex flag is set which should be the reason why new blocks are not broadcasted and make syncing block logic fail. But as I said in the PR description this test focus on the local state then I remove syncing logic out in this PR.

PR #878, opened 2020-07-13
Update rpcevo.cpp

justanwar:
Fix evo help text regarding ipAndPort.


src/rpc/rpcevo.cpp
```diff
@@ -55,7 +55,7 @@ std::string GetHelpString(int nParamNum, std::string strParamName)
         },
         {"ipAndPort",
             "%d. \"ipAndPort\"                (string, required) IP and port in the form \"IP:PORT\".\n"
-            "                              Must be unique on the network. Can be set to 0, which will require a ProUpServTx afterwards.\n"
+            "                              Must be unique on the network. Can be set to "", which will require a ProUpServTx afterwards.\n"
```
psolstice, line 58: Should be `\"\"`

justanwar: Updated

PR #892, opened 2020-08-13
Qt: 'Reindex Sigma' option fix

riordant:
## PR intention
parameter interaction happens before Qt loads settings, and so only `zapwalletmints` is loaded when `Reindex Sigma` is set, the other conf settings (`reindex`, `zapwallettxes=1`) are not.

## Code changes brief
load these conf settings in `optionsmodel.cpp`


reubenyap:
As Sigma is no longer used this needs to be fixed for Lelantus.

PR #907, opened 2020-09-10
Fix for deadlock when creating a transaction using Qt GUI with lots of inputs

psolstice:
## PR intention
This is a fix for potential deadlock if some operation in Qt GUI takes at least few seconds to run and evo znode list is refreshed in background while the operation is still running

## Code changes brief
Additional locks are acquired to make sure no deadlock happens


levonpetrosyan93 (approved):
LGTM

thebevrishot (approved):

PR #911, opened 2020-09-24
Lelantus fixes

levonpetrosyan93:


src/hdmint/tracker.cpp
```diff
@@ -228,8 +228,11 @@ bool CHDMintTracker::HasPubcoinHash(const uint256& hashPubcoin) const
     }
 
     for (auto const & it : mapLelantusSerialHashes) {
+        CWalletDB walletdb(strWalletFile);
```
psolstice, line 231: this is really bad for performance. You need to use *one* `CWalletDB` class instance and pass it as an argument to different functions. Here it's inside the loop

psolstice (changes requested):

psolstice (approved):
for testing purposes only. Further optimisation needed

PR #915, opened 2020-09-30
Merge bip47 clientapi

HysMagus:
# PR for BIP47 on Zcoin
**This PR is not ready for merging as Lelantus needs to be integrated.**


src/wallet/crypter.h
```diff
@@ -116,6 +116,7 @@ class CCryptoKeyStore : public CBasicKeyStore
 {
 private:
     CryptedKeyMap mapCryptedKeys;
+    KeyMap mapBip47Keys;
```
riordant, line 119: I don't think this separate `KeyMap` is needed. you should find some way of using `mapCryptedKey` for what you need, from what I can see in the code there isn't a reason why we need another object.

riordant (changes requested):

reubenyap:
Closing this as @a-bezrukov is doing a significant refactor of this.

PR #916, opened 2020-10-01
Update privacy by default feature

thebevrishot:
## PR intention
Update privacy by default feature by notifying users to mint all instead of show dialog and block usage.

## Code changes brief
- Create auto mint notification design file and create a class to use it
- Update and refactor Lelantus model and automint model
- Update logic to detect new funds to avoiding show notification when detected znode funds
- Update walletview class to show notification when found signals from automint model

If detect mintable funds on startup or found incoming funds, notification should be shown
atm-notification

If users click anonymize `Anonymize Funds` dialog should be shown
1) if wallet is not encrypted
atm-uen

or 2) if wallet is encrypted. (require wallet passphase to do the operation)
atm-enc

and progress message would be shown
atm-progress-unlocking

atm-progress-anonymizing

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -102,6 +102,7 @@ QT_FORMS_UI = \
   qt/forms/sigmacoincontroldialog.ui \
   qt/forms/askpassphrasedialog.ui \
   qt/forms/automintdialog.ui \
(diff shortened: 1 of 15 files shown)


a-bezrukov (approved):

psolstice (approved):
LGTM

PR #920, opened 2020-10-15
Fixing crashes during invalidateblock

levonpetrosyan93:


src/batchproof_container.cpp
```diff
@@ -65,6 +65,53 @@ void BatchProofContainer::add(lelantus::JoinSplit* joinSplit,
...
+            }
+        }
+    }
+}
+void BatchProofContainer::removeLelantus(std::unordered_map<Scalar, int> spentSerials) {
+    for(auto& spendSerial : spentSerials) {
+        std::pair<uint32_t, bool> key = std::make_pair(spendSerial.second, false);
+        if (lelantusSigmaProofs.count(key) > 0) {
+            auto &vProofs = lelantusSigmaProofs[key];
+            for (auto dataItr = vProofs.begin(); dataItr < vProofs.end(); dataItr++) {
```
psolstice, line 105: can't use `<` here

psolstice (changes requested):

src/batchproof_container.cpp
psolstice: er you can (found out it's a vector) but shouldn't. Looks alien

psolstice (approved):

PR #928, opened 2020-11-16
Lelantus fixes

levonpetrosyan93:
1. Adding function to estimate fee,
2. Fixing issue of missing jmints at coincntrol

### DIFF
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -926,9 +926,9 @@ public:
 
     // Returns a list of unspent and verified coins, I.E. coins which are ready
     // to be spent.
-    std::list<CSigmaEntry> GetAvailableCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false) const;
+    std::list<CSigmaEntry> GetAvailableCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false, bool forEstimation = false) const;
 
-    std::list<CLelantusEntry> GetAvailableLelantusCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false) const;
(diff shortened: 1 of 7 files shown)


src/wallet/wallet.h
```diff
@@ -926,9 +926,9 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
 
     // Returns a list of unspent and verified coins, I.E. coins which are ready
     // to be spent.
-    std::list<CSigmaEntry> GetAvailableCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false) const;
+    std::list<CSigmaEntry> GetAvailableCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false, bool forEstimation = false) const;
 
-    std::list<CLelantusEntry> GetAvailableLelantusCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false) const;
+    std::list<CLelantusEntry> GetAvailableLelantusCoins(const CCoinControl *coinControl = NULL, bool includeUnsafe = false, bool forEstimation = false) const;
```
psolstice, line 931: Please consider adding another function for getting an estimates (`GetAvailableLelantusCoinValues` ?). It gets all the coins with their values (no need in secret data).
It would be much better than propagating `forEstimate` flag down the API chain

psolstice (changes requested):

src/wallet/wallet.h
levonpetrosyan93: I did not go with that way as much part of the code will be copied.

psolstice (approved):
Can do it now but it needs refactor

PR #930, opened 2020-11-22
Firo rebrand

psolstice:
## PR intention
Changed name to `firo`

## Code changes brief
Lots of automatic word replacements. Graphics changed to reflect rebranding. Also adds a piece of code to manage migration to new directory and file names


.github/FUNDING.yml
```diff
@@ -1 +1 @@
-custom: ["https://zcs.zcoin.io", "https://opencollective.com/zcoin"]
+custom: ["https://zcs.firo.org", "https://opencollective.com/firo"]
```
nopslide, line 1: Not sure we have this url

justanwar: We do. Can leave it for later since it does not affect the daemon.

levonpetrosyan93 (approved):

a-bezrukov (approved):

PR #931, opened 2020-11-23
Lelantus sigma optimization

levonpetrosyan93:

### DIFF
--- a/src/liblelantus/sigmaextended_verifier.cpp
+++ b/src/liblelantus/sigmaextended_verifier.cpp
@@ -50,11 +50,7 @@ bool  SigmaExtendedVerifier::verify(
     std::vector<Scalar> f_i_;
     f_i_.resize(N);
 
-    ptr = f_i_.data();
-    start_ptr = ptr;
-    end_ptr = ptr + N;
-    Scalar f_i(uint64_t(1));
-    compute_fis(f_i, m, f_);
+    compute_fis(m, f_, f_i_);
 
     /*
          * Optimization for getting power for last 'commits' array element is done similarly to the one used in creating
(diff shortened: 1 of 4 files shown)


src/liblelantus/sigmaextended_verifier.cpp
```diff
@@ -88,10 +84,9 @@ bool  SigmaExtendedVerifier::verify(
         Scalar fi_sum(uint64_t(0));
         for (uint64_t i = I[j] + 1; i < n; i++)
             fi_sum += f_[j*n + i];
-        pow += fi_sum * xj.pow * f_part_product[m - j - 1];
+        f_i_[N - 1] += fi_sum * xj.pow * f_part_product[m - j - 1];
```
psolstice, line 87: is `pow` and `f_i_[N-1]` the same before the loop? They are calculated in two independent ways. I believe it's clearer (and slightly faster) to leave the code here as is and then assign instead of emplacing.

src/sigma/sigmaplus_verifier.hpp
```diff
@@ -106,10 +103,9 @@ bool SigmaPlusVerifier<Exponent, GroupElement>::verify(
             Exponent fi_sum(uint64_t(0));
             for (int i = I[j] + 1; i < n; i++)
                 fi_sum += f[j*n + i];
-            pow += fi_sum * xj * f_part_product[m - j - 1];
+            f_i_[N - 1] += fi_sum * xj * f_part_product[m - j - 1];
```
psolstice, line 106: See comment in creation part

psolstice (changes requested):

src/liblelantus/sigmaextended_verifier.cpp
levonpetrosyan93: Yes `pow` and `f_i_[N-1]` are the same before this step. I don't think it will change something in terms of performance, but yea may be it is more clear. will change it

src/liblelantus/sigmaextended_verifier.h
```diff
@@ -43,11 +43,18 @@ class SigmaExtendedVerifier{
...
+
 private:
     GroupElement g_;
     std::vector<GroupElement> h_;
     uint64_t n;
     uint64_t m;
+    mutable Scalar* ptr;
+    mutable Scalar* start_ptr;
+    mutable Scalar* end_ptr;
 };
```
psolstice, line 58: not a nice way to do it. Please consider moving these to `compute_fis`/`compute_batch_fis` arguments

psolstice (changes requested):

psolstice (approved):

PR #949, opened 2020-12-19
Update copyright ©️ in LicenseInfo ℹ️

KuroGuo:
There was a duplicate in the copyright :copyright: information and the year didn't match, I updated it. :beers:

image

image


HysMagus:
Shouldn't this also have the DASH guys listed on the copyright?

KuroGuo:
I don't think Dash should be listed because Dash is not Firo's dad :older_man:

reubenyap:
Dash should be included because of the usage of their masternode infrastructure work.
The double mention was probably due to a find/replace when porting over masternode code.

KuroGuo:
Added at the bottom.

`This product includes Masternodes software developed by the Dash Core
developers <https://www.dash.org>.
`

HysMagus:
utACK

Issue #939, opened 2020-12-04
Segmentation fault in GUI of v0.14.1.2 in Ubuntu 18.04


Sviluppo718:
For some reason I can only get the text-based firod and firo-cli to run on this system (the qt version built and runs on a different computer running Pop!_OS _based on Ubuntu 20.4 with a comparatively minimal set of packages_).  **On this system, firo-qt shows the splash screen and then segmentation faults.**   The debug.log doesn't show any helpful information (it simply seems to close in the middle of the reindexing that I started).

[e.g. the last few lines in debug.log:]
```
2020-12-04 05:28:42 CheckTransaction nHeight=69950, isVerifyDB=false, isCheckWallet=false, txHash=e8f75f1b41298e660769734a2458fd818893a42d0479a1555bcbd854ece602a3
2020-12-04 05:28:42 CheckMintFiroTransaction txHash = 9d18c9c003e8da8b7459b2e4cb3185a2bca8e6670f78af65f3c0e816684c66d0
2020-12-04 05:28:42 nValue = 10000000000
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 ConnectTipZC: mint added denomination=100, id=60
2020-12-04 05:28:42 UpdateTip() pindexNew.nHeight=69950
```

After firo-qt closes I've tried:

```  $ echo $? ```

which outputs a return code: 139 

Trying it with system-trace:

```  $ strace firo-qt```

The last few lines are:

```
stat("/etc/xdg/Firo/Firo-Qt.conf", 0x7fffdf4e3a70) = -1 ENOENT (No such file or directory)
stat("/etc/xdg/xdg-ubuntu/Firo.conf", 0x7fffdf4e3aa0) = -1 ENOENT (No such file or directory)
stat("/etc/xdg/xdg-ubuntu/Firo.conf", 0x7fffdf4e3aa0) = -1 ENOENT (No such file or directory)
stat("/etc/xdg/xdg-ubuntu/Firo.conf", 0x7fffdf4e3a70) = -1 ENOENT (No such file or directory)
[...]

PR #950, opened 2020-12-21
Lelantus unittests and fixes

### LINKED ISSUE #946: Daemon and QT crashes
On Windows, the QT can sometimes crash with this error in logs:
`Windows Exception: EXCEPTION_ACCESS_VIOLATION`

On Linux, might be related to #939

levonpetrosyan93:
What brings this PR

1. Fixing and adding unit tests
2. Fixing bug at coin choosing algorithm
3. Fixing bug resulting wrong debug log at joinsplit creation
4. Fixing minor bug at joinsplit creation (our implementation was not resulting it, but in unit tests we have such case)
5. Fixing crashes when upgrading with non-hd wallet
6. Fixing issues and crashes when -disablewallet arg used, #939 
7. Removed wrong background color when dark mode is used.
8. Fixes #946, the reason was incorrect mutex locks.
9. Fixed crash after try to create lelantus mint when wallet not yet synced, Prevented 0 mint creation at Ui level.

### DIFF
--- a/qa/rpc-tests/lelantus_mintspend.py
+++ b/qa/rpc-tests/lelantus_mintspend.py
@@ -21,8 +21,8 @@ class LelantusMintSpendTest(BitcoinTestFramework):
 
         mint_trans = list()
         coin = 100000000;
-        mint_trans.append(self.nodes[0].mintlelantus(1 * coin))
-        mint_trans.append(self.nodes[0].mintlelantus(2 * coin))
+        mint_trans.append(self.nodes[0].mintlelantus(1))
+        mint_trans.append(self.nodes[0].mintlelantus(2))
 
         # Get last added transaction and fee for it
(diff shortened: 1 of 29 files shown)


qa/rpc-tests/lelantus_mintspend.py
```diff
@@ -0,0 +1,139 @@
...
+
+        start_bal = self.nodes[0].getbalance()
+
+        mint_trans = list()
+        coin = 100000000;
+        mint_trans.append(self.nodes[0].mintlelantus(1))
+        mint_trans.append(self.nodes[0].mintlelantus(2))
+
+        # Get last added transaction and fee for it
+        info = self.nodes[0].gettransaction(mint_trans[-1])
```
psolstice, line 28: Test fails here

psolstice (changes requested):
please fix failing test

psolstice (approved):

PR #947, opened 2020-12-19
Sporks for management of certain network features

psolstice:
## PR intention
Introduce sporks used to disable or limit some of features of firo

## Code changes brief
New type of transaction which is once mined have specific effects on consensus rules in subsequent blocks. Some reorganization in lelantus code.

PR #951, opened 2020-12-23
Raising lelantus joinsplit limit to 5000

psolstice:
## PR intention
Raise lelantus limits to 5001 per transaction and 5100 per block. Sporks (#947 ) can be used to lower this limit temporarily.

## Code changes brief
Change of several constants

### DIFF
--- a/src/zerocoin_params.h
+++ b/src/zerocoin_params.h
@@ -130,19 +130,19 @@ static const int64_t DUST_HARD_LIMIT = 1000;   // 0.00001 FIRO mininput
 #define ZC_SIGMA_INPUT_LIMIT_PER_TRANSACTION            35
 
 // Value of lelantus spends allowed per block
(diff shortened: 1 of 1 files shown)


reubenyap:
The idea behind this is a balance of allowing adoption of Lelantus tx to happen especially as we move to a privacy by default model while balancing with mitigating risk of a new privacy protocol. It is intended for this limit to be raised as we grow more confident on the protocol.

Related to this is #947 which allows the core team to turn off Lelantus in case of an attack but with a limited validity period of one year as we allow the protocol to mature.

levonpetrosyan93 (dismissed):

a-bezrukov (approved):

levonpetrosyan93 (approved):

skironDotNet:
@reubenyap can we increase the limit on transparent funds to at least 50k FIRO, the rationale is here

https://github.com/firoorg/firo/issues/1215#issuecomment-1475105910

PR #966, opened 2021-01-08
To support multiple inputs in Lelantus spend tx in Elysium

nopslide:
Increase to support multiple inputs in Lelantus spend tx in Elysium layer, due to with 4k, it will support only one spend input, which impractical to support multiple spend input in a spend tx.

### DIFF
--- a/src/script/standard.h
+++ b/src/script/standard.h
@@ -27,7 +27,7 @@ public:
     CScriptID(const uint160& in) : uint160(in) {}
 };
 
-static const unsigned int MAX_OP_RETURN_RELAY = 4096;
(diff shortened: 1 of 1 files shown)


a-bezrukov (approved):
I think we don't need an HF for increasing this size

psolstice (approved):

psolstice (changes requested):
`elysium_build_tx_tests/wallettxbuilder_create_normal_b` is failing

nopslide:
In the new version, I deleted class B transaction type, so this error from class B transaction tests will not be an issue

sproxet:
This was merged into the elysium-lelantus branch in 6a740d535c0d41864863267f38eb0713a80d00a7.

PR #969, opened 2021-01-17
Update chainparams.cpp

xinya-123:
Updating checkpoint can better protect the security of blockchain！


reubenyap:
Not needed with Chainlocks but will do a manual checkpoint prior to chainlocks later.

PR #971, opened 2021-01-17
Fix for coin not found issue.

levonpetrosyan93:
- Fix for sigma hd mint bug, where coin group id is not being set, it results an coin not found in chin issue during Lelantus joinsplit creation, The reason wast that, setting group id in SetMintSeedSeen() was missing, and it was remaining -1 in sum circumstances, the code is adding that,   
- Making error messages more accurate.

### DIFF
--- a/src/hdmint/wallet.cpp
+++ b/src/hdmint/wallet.cpp
@@ -487,12 +487,17 @@ bool CHDMintWallet::SetMintSeedSeen(CWalletDB& walletdb, std::pair<uint256,MintP
     }
 
     LogPrintf("%s: Creating mint object.. \n", __func__);
+
+    int id;
+    std::tie(std::ignore, id) = sigma::CSigmaState::GetState()->GetMintedCoinHeightAndId(sigma::PublicCoin(bnValue, denom));
+
     // Create mint object
     CHDMint dMint(mintCount, seedId, hashSerial, bnValue);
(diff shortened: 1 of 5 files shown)


src/hdmint/wallet.cpp
```diff
@@ -487,12 +487,17 @@ bool CHDMintWallet::SetMintSeedSeen(CWalletDB& walletdb, std::pair<uint256,MintP
...
+
+    int id;
+    std::tie(std::ignore, id) = sigma::CSigmaState::GetState()->GetMintedCoinHeightAndId(sigma::PublicCoin(bnValue, denom));
+
     // Create mint object
     CHDMint dMint(mintCount, seedId, hashSerial, bnValue);
     int64_t amount;
     DenominationToInteger(denom, amount);
     dMint.SetAmount(amount);
     dMint.SetHeight(nHeight);
```
psolstice, line 499: `nHeight` is passed to this function as `const int &` which is something *really* weird. Please change it to just `int`

psolstice (changes requested):

src/hdmint/wallet.cpp
levonpetrosyan93: okay,

psolstice (approved):

a-bezrukov (approved):

PR #974, opened 2021-01-18
JSplit fee estimation and calculation optimized

levonpetrosyan93:
- In Lelantus coincontrol fee estimation was not set to be calculated for joinsplit, added the check if transaction is Lelantus joinsplit, it uses EstimateJoinSplitFee() function for calculation,
- During joinsplit creation it was picking minimum fee, and if SubtractFeeFromAmount was mot set, it was doing one unneeded iteration, so before starting to create transaction, fee estimation is called and returned value is being chosen as transaction fee.
- In fee estimation GetAvailableCoins() and balance calculation for sigma was called in each iteration, it is moved out of loop,


reubenyap:
@levonpetrosyan93 needs to be updated

a-bezrukov (approved):

psolstice (approved):

PR #981, opened 2021-01-24
Fixing Insufficient Funds error when minting big amounts

levonpetrosyan93:
- Fixes a corner case issue, when you had balance on a single address bigger then mint limit and you was trying to mint such amount that the change from first mint will be used for second mint, for example you have 7k on one address and no other utxo and you try to mint 6k, 
The reason is that during coin selection it calls GetCredit() function for transaction, which utxo is trying to use, as it is mint transaction, it is trying to read HDMint from db, but as it is not broadcasted yet, it is not added into db, so it is not recognized yours, the simple fix is this 5a77ae1d5b3075e8f0d5e3ed2b2eb42d58c9ea51

- Second issue was, that when you had very small utxo, which is a single one for and address, less then min transaction fee, it was trying to create mint with negative value and during broadcast  it was throwing and exception, the fix is to skip such utxo's during minting,

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -5308,7 +5308,7 @@ bool CWallet::CreateLelantusMintTransactions(
                 CAmount nValueToSelect, mintedValue;
 
                 std::set<std::pair<const CWalletTx *, unsigned int>> setCoins;
-                bool skipIfNegative = false;
+                bool skipCoin = false;
                 // Start with no fee and loop until there is enough fee
                 while (true) {
                     mintedValue = valueToMintInTx;
@@ -5322,7 +5322,7 @@ bool CWallet::CreateLelantusMintTransactions(
 
                     if (!MoneyRange(mintedValue) || mintedValue == 0) {
                         valueAndUTXO.erase(itr);
-                        skipIfNegative = true;
+                        skipCoin = true;
                         break;
                     }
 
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -5320,6 +5320,12 @@ bool CWallet::CreateLelantusMintTransactions(
                         nValueToSelect = mintedValue + nFeeRet;
                     }
 
+                    if (!MoneyRange(mintedValue) || mintedValue == 0) {
+                        valueAndUTXO.erase(itr);
+                        skipIfNegative = true;
```
psolstice, line 5325: bad variable name. It is known by this time the value is negative, no ifs. Really confusing

src/wallet/wallet.cpp
```diff
@@ -5536,79 +5542,85 @@ bool CWallet::CreateLelantusMintTransactions(
...
-                    size_t nLimitAncestors = GetArg("-limitancestorcount", DEFAULT_ANCESTOR_LIMIT);
-                    size_t nLimitAncestorSize = GetArg("-limitancestorsize", DEFAULT_ANCESTOR_SIZE_LIMIT) * 1000;
-                    size_t nLimitDescendants = GetArg("-limitdescendantcount", DEFAULT_DESCENDANT_LIMIT);
-                    size_t nLimitDescendantSize = GetArg("-limitdescendantsize", DEFAULT_DESCENDANT_SIZE_LIMIT) * 1000;
-                    std::string errString;
-                    if (!mempool.CalculateMemPoolAncestors(entry, setAncestors, nLimitAncestors, nLimitAncestorSize,
-                                                           nLimitDescendants, nLimitDescendantSize, errString)) {
-                        strFailReason = _("Transaction has too long of a mempool chain");
-                        return false;
+                if(!skipIfNegative) {
```
psolstice, line 5545: can we just `continue` instead of huge `if` clause?

psolstice (changes requested):

src/wallet/wallet.cpp
levonpetrosyan93: yes, will change, it will look better

psolstice (approved):

a-bezrukov (approved):

PR #1004, opened 2021-02-21
Fee estimation optimized

levonpetrosyan93:


a-bezrukov (approved):

src/qt/coincontroldialog.cpp
```diff
@@ -523,8 +523,11 @@ void CoinControlDialog::updateLabels(WalletModel *model, QDialog* dialog, bool a
     // calculation
     if (nQuantity > 0)
     {
-        if(anonymousMode){
-            std::tie(nPayFee, nBytes) = model->getWallet()->EstimateJoinSplitFee(nPayAmount,CoinControlDialog::fSubtractFeeFromAmount, coinControl);
+        if (anonymousMode) {
+            // 956 is constant part, mainly Schnorr and Range proof, 2560 is for each sigma/aux data
+            // 83 assuming 1 jmint, 34 is the size of each normal vout,  10 is the size of empty transaction, 52 other constant parts
+            nBytes = 956 + 2560 * vOutputs.size() + 83 + CoinControlDialog::payAmounts.size()  * 34  + 10 + 52;
+            nPayFee = CWallet::GetMinimumFee(nBytes, nTxConfirmTarget, mempool);
```
psolstice, line 530: Does this formula take new Qk Schnorr proof into account?
We really need to create a set of functions to deal with size estimates and use it instead of such hardcoded estimates. Sooner or later we will end up with estimate which is too low and transactions will be rejected at mempool level.

psolstice (changes requested):

src/qt/coincontroldialog.cpp
levonpetrosyan93: no, I will update this,  I have implemented such function, but it works slow and  is making coincontrol work laggy, may be better to include this for now. after some refactorings and optimizations may be such function will be useful.

psolstice (approved):

a-bezrukov (approved):

PR #1012, opened 2021-04-02
Lelantus improvements

levonpetrosyan93:


psolstice (approved):

src/liblelantus/lelantus_primitives.h
```diff
@@ -27,6 +28,8 @@ struct NthPower {
 
     void go_next() {
         pow *= num;
+        if (pow == Scalar(uint64_t(1)))
+            throw std::invalid_argument("NthPower resulted 1");
```
psolstice, line 32: I think this has DoS potential because `NthPower` is used in muptiple places and not only with challenge value. I.e. attacker can set some value in proof to 1 and if the exception is not properly caught it'll crash all the nodes. It's better to move the check to higher level.

psolstice (changes requested):

src/liblelantus/lelantus_primitives.h
levonpetrosyan93: I have put this call in try-catch in verifier side. I think that is enoygh for preventing such risks.

levonpetrosyan93: will doublecheck to verify that I did not missed it somewhere.

psolstice (dismissed):

PR #1013, opened 2021-04-13
Bip47 lelantus

a-bezrukov:
## PR intention
An implementation of RAP (bip47) payment codes. 

## Code changes brief
The secret point is calculated using Lelantus private/public keys. This required a minor change into the Lelantus tx creation. Other that this, there is no significant changes to the existing code and the BIP47 code is done as an add-on.


levonpetrosyan93 (dismissed):
LGTM

src/bip47/address.cpp
```diff
@@ -0,0 +1,18 @@
+#include "bip47/address.h"
+#include "bip47/bip47utils.h"
+
+namespace bip47 {
+
+CAddress::CAddress(CBitcoinAddress const & address, bool isUsed)
+:address(address), isUsed_(isUsed)
+{}
+
+void CAddress::isUsed(bool isUsed) {
```
margaryan, line 10: woudln't be better to rename this setUseed or like that ?

a-bezrukov: Fixed, thanks!

levonpetrosyan93 (approved):

src/wallet/wallet.cpp
```diff
@@ -6936,3 +7337,137 @@ bool CMerkleTx::AcceptToMemoryPool(const CAmount &nAbsurdFee, CValidationState &
...
+    LogBip47("Sending pcode: %s to naddress: %s\n", pchannel.getMyPcode().toString(), notifAddr.ToString());
+
+    CValidationState state;
+    if (!pwallet->CommitTransaction(wtxNew, reservekey, g_connman.get(), state)) {
+        strError = strprintf("Error: The transaction was rejected! Reason given: %s", state.GetRejectReason());
+        throw JSONRPCError(RPC_WALLET_ERROR, strError);
+    }
+}
+
+void SendNotificationTxLelantus(CWallet * const pwallet, bip47::CPaymentChannel const & pchannel, CWalletTx& wtxNew)
```
psolstice, line 7402: Definitely should be a method of `CWallet` class.

src/wallet/wallet.cpp
```diff
@@ -6936,3 +7337,137 @@ bool CMerkleTx::AcceptToMemoryPool(const CAmount &nAbsurdFee, CValidationState &
 
 bool CompSigmaHeight(const CSigmaEntry &a, const CSigmaEntry &b) { return a.nHeight < b.nHeight; }
 bool CompSigmaID(const CSigmaEntry &a, const CSigmaEntry &b) { return a.id < b.id; }
+
+namespace {
+//I am leaving this proc in place as it may be useful for further urgent cases.
```
psolstice, line 7342: What was the original urgent case? Comment doesn't look good. Please consider removing the function altogether.

psolstice (changes requested):

src/wallet/wallet.cpp
a-bezrukov: It is Lelantus disabled.

src/wallet/wallet.cpp
a-bezrukov: I think it should not. If speak about `PrepareAndSendNotificationTx`, then maybe it should. But what for, what would be the benefits of this? It's a utility function as it just uses ::CWallet services to send a very special transaction.

src/wallet/wallet.cpp
a-bezrukov: I'll save it into some other repo

psolstice (approved):

levonpetrosyan93 (approved):

PR #1016, opened 2021-04-16
Fix listransactions performance

a-bezrukov:
## PR intention
Improves performance of listtransaction calls when users don't need the tx to be marked as masternode output

## Code changes brief
Added an additional flag to listtransactions and switched back to the caching schema.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1627,7 +1627,7 @@ UniValue listtransactions(const JSONRPCRequest& request)
             "2. count          (numeric, optional, default=10) The number of transactions to return\n"
             "3. skip           (numeric, optional, default=0) The number of transactions to skip\n"
             "4. include_watchonly (bool, optional, default=false) Include transactions to watch-only addresses (see 'importaddress')\n"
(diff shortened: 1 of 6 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -1625,6 +1627,7 @@ UniValue listtransactions(const JSONRPCRequest& request)
             "2. count          (numeric, optional, default=10) The number of transactions to return\n"
             "3. skip           (numeric, optional, default=0) The number of transactions to skip\n"
             "4. include_watchonly (bool, optional, default=false) Include transactions to watch-only addresses (see 'importaddress')\n"
+            "5. omit_mnpayments   (bool, optional, default=false) Do not handle masternode outputs (improves performance)\n"
```
psolstice, line 1630: Help string and variable name suggest we remove znode payments from the list which is not true, we just don't check if txout is znode payment.

psolstice (changes requested):

psolstice (approved):

levonpetrosyan93 (approved):

PR #1017, opened 2021-04-19
Fixing transaction list issue after recovery

levonpetrosyan93:


a-bezrukov (approved):

psolstice (changes requested):
Could you please write a short description what was a problem in the first place?

reubenyap:
When recovering the wallet with a mnemonic, transactions list has empty 0.0000 FIRO entries. 

image

image

psolstice (approved):

PR #1019, opened 2021-04-27
Remove unnecessary zero checks from Lelantus protocol

AaronFeickert:
Removes unnecessary zero checks from range and one-of-many proof verifiers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ firo.creator
 firo.creator.user
 firo.files
 firo.includes
+
+# Backups
+*.backup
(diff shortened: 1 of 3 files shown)


reubenyap:
Test failing please kindly resolve lelantus_mintspend.py

levonpetrosyan93 (approved):

levonpetrosyan93 (dismissed):

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

psolstice (changes requested):
Technically it's a hard fork. Thus proper management of block number is needed and checks should be dropped only after specific block. Otherwise it's possible to create a transaction which causes forks in our network shortly after rolling out the update. Given that these checks are relatively fast I don't think it's worth it

PR #1037, opened 2021-05-21
Replace part of balance verification with multiscalar multiplication operation

AaronFeickert:
Updates balance proof verification to use multiscalar multiplication operations where applicable.

Review requested.

### DIFF
--- a/src/liblelantus/lelantus_verifier.cpp
+++ b/src/liblelantus/lelantus_verifier.cpp
@@ -227,15 +227,24 @@ bool LelantusVerifier::verify_schnorrproof(
         const std::vector<PublicCoin>& Cout,
         const LelantusProof& proof,
         unique_ptr<ChallengeGenerator>& challengeGenerator) {
-    GroupElement A;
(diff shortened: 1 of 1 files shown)


levonpetrosyan93 (changes requested):
The idea is good, but I will suggest to leave computation of `A` as is, as previously it was doing `Cout.size()` times additions and one multiplication, but now it is doing several multiplications,
I tried to test, and in my setup I got this 716 for old version, 491 for your case, and 400 in case what I am suggesting, numbers are in clock ticks.

AaronFeickert:
> The idea is good, but I will suggest to leave computation of `A` as is, as previously it was doing `Cout.size()` times additions and one multiplication, but now it is doing several multiplications,
> I tried to test, and in my setup I got this 716 for old version, 491 for your case, and 400 in case what I am suggesting, numbers are in clock ticks.

Updated.

levonpetrosyan93 (approved):

psolstice (approved):

PR #1043, opened 2021-06-17
Add Lelantus Support to Elysium

sproxet:
This PR adds Lelantus support to (and removes Sigma support from) our Elysium layer.

See https://gist.github.com/sproxet/494254e29adcc06cd9ec5dc75679790c examples of how to use it.

You'll need to pass `--enable-elysium` at build time and `-elysium` at run time.


levoncrypto (changes requested):

reubenyap:
Closing this PR so that a new one can be issued due to time elapsed.

PR #1045, opened 2021-06-21
Adding private balance value into getBalance rpc result

levonpetrosyan93:


a-bezrukov (approved):

psolstice (changes requested):
My suggestion is instead of modifying existing RPC (and losing ability to get transparent balance with RPC call) we add another RPC to get private balance

psolstice (approved):

a-bezrukov (approved):

PR #1046, opened 2021-06-23
Bip47 GUI fixes

a-bezrukov:
## PR intention
This is a series of GUI improvements:
1. Improved sending to a RAP address workflow - removed showing the secret address, made impossible to send unless a notification tx confirmed.
2. Implemented the receiving RAP address relabeling and RAP addresses creation during the wallet recover
3. Renamed pcodes to be RAP addresses for RPCs
4. Minor fixes - logging, naming
5. Fixed the bug with subsequently accpeted blocks


levonpetrosyan93 (approved):

levonpetrosyan93 (dismissed):

levonpetrosyan93 (approved):

src/qt/walletmodel.cpp
```diff
@@ -1462,18 +1491,51 @@ int WalletModel::getDefaultConfirmTarget() const
...
+        ctx.delayRelock(60);
+        {
+            std::lock_guard<std::mutex> _(queueMutex);
+            for(std::deque<bip47::CAccountReceiver const *>::iterator iter = receiverAccountNumQueue.begin(); iter != receiverAccountNumQueue.end(); ++iter)
+                bip47::utils::AddReceiverSecretAddresses(**iter, *wallet);
+            receiverAccountNumQueue.clear();
+        }
+        LOCK(cs_main);
+        if (pBlockIndex != chainActive.Tip()) {
+            wallet->ScanForWalletTransactions(pBlockIndex, false, false);
```
psolstice, line 1538: 1. do we need to check if reorg happened and `pBlockIndex` is not on the chain anymore? (`chainActive.Contains`?)
2. why start with this [presumably already scanned] block and not the next one (`chainActive.Next(pBlockIndex)`?)

psolstice, line 1538: 1. do we need to check if reorg happened and `pBlockIndex` is not on the chain anymore? (`chainActive.Contains`?)
2. why start with this [presumably already scanned] block and not the next one (`chainActive.Next(pBlockIndex)`?)

psolstice (changes requested):

src/qt/walletmodel.cpp
a-bezrukov: 1. This is possible theoretically, so fixed, please review
2. There can be other bip47 transactions in this block that's why we unlock wallet and rescan all the blocks from this one to the tip

src/qt/walletmodel.cpp
```diff
@@ -1534,6 +1534,9 @@ void WalletModel::handleBip47Keys(int receiverAccountNum, void * pBlockIndex_)
             receiverAccountNumQueue.clear();
         }
         LOCK(cs_main);
+        for (; !chainActive.Contains(pBlockIndex); pBlockIndex = pBlockIndex->pprev) {
+            if (!pBlockIndex) return;
+        }
```
psolstice, line 1539: you don't need to scan the entire history of blocks, if active chain contains `pBlockIndex` it contains all its ancestors

psolstice (changes requested):

src/qt/walletmodel.cpp
a-bezrukov: I don't get your point. It does not scan the entire history. It goes down to the first block which is on the active chain and starts rescanning from there.

psolstice: `pBlockIndex->pprev` is an ancestor i.e. _previous_ block, not the next one. You're scanning the entire block chain up to genesis block here. Testing if `pBlockIndex` is in the chain is enough

psolstice: Or sorry, what are you trying to compute here, maybe I didn't understand. Fork point?

psolstice: There is a member of `CChain` called `FindFork` to do this

a-bezrukov: Yes, it looks up for a fork point.

psolstice (dismissed):

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

psolstice (approved):

PR #1047, opened 2021-06-28
Using absolute path for dumpwallet rpc

levonpetrosyan93:
This PR fixes dumpwallet path issues, when on some platforms dump is not saved at correct directory. 
It also returns the path as an output.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -693,10 +693,10 @@ UniValue dumpwallet(const JSONRPCRequest& request)
 
     EnsureWalletIsUnlocked(pwallet);
 
-    ofstream file;
-    // need change to std::filesystem::absolute() for C++17
     boost::filesystem::path filePath = request.params[0].get_str();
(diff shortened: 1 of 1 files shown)


a-bezrukov (dismissed):

a-bezrukov (approved):

a-bezrukov (approved):

src/wallet/rpcdump.cpp
```diff
@@ -693,8 +693,12 @@ UniValue dumpwallet(const JSONRPCRequest& request)
 
     EnsureWalletIsUnlocked(pwallet);
 
+    boost::filesystem::path filePath = request.params[0].get_str();
+    filePath = boost::filesystem::absolute(filePath);
```
psolstice, line 697: I suggest using `canonical` instead of `absolute` if you really want to show users where their file ends up. E.g. `/home/user/mydir/../somedir/wallet.dump` isn't something user can quickly parse

psolstice (changes requested):

psolstice (approved):

PR #1048, opened 2021-06-30
Batching fail fix

levonpetrosyan93:
The PR fixes batch verification failing issue in case of reindex/sync, The fail was happening when you had reorgs in chain, During DisconnectBlock it should remove proofs from container, but it was not removed and verification failed as proof from wrong chain is not valid after reorg. 
Removing was not being done correctly as third element in key, was being set as true, https://github.com/firoorg/firo/pull/1048/commits/7c9550af534691ee8c1e695400bb562bd87b02bc#diff-755c76a1b8be7607bc371ca3a5b4ea39a91bf46946913fd5fd67b4d06b808d7dR97 , 
but actually it should be false, but now in removeLelantus() function is being removed also sigmaToLelantus proofs.

### DIFF
--- a/src/batchproof_container.cpp
+++ b/src/batchproof_container.cpp
@@ -69,35 +69,13 @@ void BatchProofContainer::add(lelantus::JoinSplit* joinSplit,
 
 void BatchProofContainer::removeSigma(const sigma::spend_info_container& spendSerials) {
     for (auto& spendSerial : spendSerials) {
-        bool foundAtSigma =  false;
         for (auto& itr :sigmaProofs) {
             if (itr.first.first == spendSerial.second.denomination && itr.first.second.first == spendSerial.second.coinGroupId) {
                 auto& vProofs = itr.second;
                 for (auto dataItr = vProofs.begin(); dataItr != vProofs.end(); dataItr++) {
                     if (dataItr->coinSerialNumber == spendSerial.first) {
                         vProofs.erase(dataItr);
-                        foundAtSigma = true;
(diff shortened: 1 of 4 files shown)


src/batchproof_container.cpp
```diff
@@ -69,60 +69,50 @@ void BatchProofContainer::add(lelantus::JoinSplit* joinSplit,
...
 
-        for (auto dataItr = vProofs->begin(); dataItr != vProofs->end(); dataItr++) {
-            if (dataItr->serialNumber == spendSerial.first) {
-                vProofs->erase(dataItr);
-                break;
-            }
+void BatchProofContainer::erase(std::vector<LelantusSigmaProofData>* vProofs, const Scalar& serial) {
+    for (auto dataItr = vProofs->begin(); dataItr != vProofs->end(); dataItr++) {
+        if (dataItr->serialNumber == serial) {
+            vProofs->erase(dataItr);
```
psolstice, line 115: This will crash if serial is the last element of `vProofs` (`dataItr` will be past the end of vector). Also `std::vector<T>::erase` will invalidate iterators in some cases. I'd do it with `std::remove_if`.

psolstice (changes requested):

src/batchproof_container.cpp
levonpetrosyan93: agree, done

psolstice (approved):

a-bezrukov (approved):

PR #1056, opened 2021-07-19
Fixing performance issue by removing uneeded GroupElement conversions…

levonpetrosyan93:

### DIFF
--- a/src/secp256k1/src/cpp/GroupElement.cpp
+++ b/src/secp256k1/src/cpp/GroupElement.cpp
@@ -555,6 +555,10 @@ std::size_t GroupElement::hash() const
     return std::hash<std::string>()(std::string(coord.begin(), coord.end()));
 }
 
+std::size_t GroupElement::lowestbyte() const {
+    return reinterpret_cast<secp256k1_gej *>(g_)->x.n[0];
+}
+
 const void* GroupElement::get_value() const {
     return g_;
 }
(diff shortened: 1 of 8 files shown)


a-bezrukov (dismissed):

src/secp256k1/src/cpp/GroupElement.cpp
```diff
@@ -555,6 +555,12 @@ std::size_t GroupElement::hash() const
     return std::hash<std::string>()(std::string(coord.begin(), coord.end()));
 }
 
+std::size_t GroupElement::lowest_limb() const {
+    secp256k1_fe x = reinterpret_cast<secp256k1_gej *>(g_)->x;
+    secp256k1_fe_normalize(&x);
+    return x.n[0];
```
psolstice, line 561: 64-bit implementation of `secp256k1` has 58-bit limb. This will not be uniform distribution in [0, SIZE_MAX] range (because it's less than 2^58). As a result (depending on implementation) unordered map/set may have just 1/64 buckets used decreasing performance. Examining C++ STL source is not for faint hearted so it's easier to just use for hash something like `x.n[0] ^ (x.n[1] << 8)` which will give you all bits in 64-bit integer and will be uniformly distributed. Then you need to rename it to something like `get_hash()`

psolstice (changes requested):

src/secp256k1/src/cpp/GroupElement.cpp
psolstice: er, it's 52 bits and hence `x.n[0] ^ (x.n[1] << 16)` should be used. This will work both in 64-bit and 32-bit cases

psolstice (approved):

a-bezrukov (approved):

PR #1058, opened 2021-08-01
Multi threading on Lelantus-sigma prover and sigma/lelantus batch verification 

levonpetrosyan93:
1. Adding simple thread pool,
2. Applying multithrading on  Lelantus-sigma prover 
3. Applying multithrading on  batch verification during sync/reindex.


a-bezrukov:
Thread sanitizer does not report anything new for the batching.

reubenyap:
Could you post the before/after benchmarks here?

levonpetrosyan93:
Running benchmarks on  i7-11700K 3.60GHz × 16 

number of coins 1, parallel proof time 1624, regular proof time 1616
number of coins 2, parallel proof time 1672, regular proof time 3155
number of coins 8, parallel proof time 2046, regular proof time 12598
number of coins 12 parallel proof time 2602 regular proof time 18833

src/liblelantus/threadpool.h
```diff
@@ -0,0 +1,109 @@
...
+        for (std::thread &t: threadsToJoin)
+            t.join();
+    }
+
+    // Post a task to the thread pool and return a future to wait for its completion
+    std::future<Result> PostTask(std::function<Result()> task) {
+        std::packaged_task<Result()> packagedTask(std::move(task));
+        std::future<Result> ret = packagedTask.get_future();
+
+        std::unique_lock<std::mutex> lock(task_queue_mutex);
```
levoncrypto, line 90: use std::scoped_lock here (if we have C++17, if no then std::lock_guard)

src/batchproof_container.cpp
```diff
@@ -171,75 +195,93 @@ void BatchProofContainer::batch_lelantus() {
...
+                for (auto& proofData : itr->second) {
+                    serials.emplace_back(proofData.serialNumber);
+                    setSizes.emplace_back(proofData.anonymitySetSize);
+                    proofs.emplace_back(proofData.lelantusSigmaProof);
+                    challenges.emplace_back(proofData.challenge);
+                }
 
-        lelantus::SigmaExtendedVerifier sigmaVerifier(params->get_g(), params->get_sigma_h(), params->get_sigma_n(),
-                                            params->get_sigma_m());
+                lelantus::SigmaExtendedVerifier sigmaVerifier(params->get_g(), params->get_sigma_h(), params->get_sigma_n(),
```
levoncrypto, line 255: I think you can move this outside of loop

src/liblelantus/lelantus_prover.cpp
```diff
@@ -101,32 +102,70 @@ void LelantusProver::generate_sigma_proofs(
...
-        Yk[i].resize(params->get_sigma_m());
-        a[i].resize(params->get_sigma_n() * params->get_sigma_m());
-        sigmaProver.sigma_commit(C_, indexes[i], rA[i], rB[i], rC[i], rD[i], a[i], Tk[i], Pk[i], Yk[i], sigma[i], sigma_proofs[i]);
+
+    std::size_t threadsMaxCount = std::min((unsigned int)N, std::thread::hardware_concurrency());
+    std::vector<std::shared_future<void>> parallelTasks;
+    parallelTasks.reserve(threadsMaxCount);
+    ParallelOpThreadPool<void> threadPool(threadsMaxCount);
+
+    std::vector<std::vector<GroupElement>> C_;
```
levoncrypto, line 111: you can store only one array, and clear() it at the start of each iteration

src/liblelantus/lelantus_prover.cpp
```diff
@@ -101,32 +102,70 @@ void LelantusProver::generate_sigma_proofs(
...
+                rA[i].randomize();
+                rB[i].randomize();
+                rC[i].randomize();
+                rD[i].randomize();
+                Tk[i].resize(params->get_sigma_m());
+                Pk[i].resize(params->get_sigma_m());
+                Yk[i].resize(params->get_sigma_m());
+                a[i].resize(params->get_sigma_n() * params->get_sigma_m());
+
+                auto& sigma_i = sigma[i];
```
levoncrypto, line 140: const auto&

levoncrypto (changes requested):

src/liblelantus/lelantus_prover.cpp
levonpetrosyan93: We are passing all values by reference, storing and clearing will not work.

src/batchproof_container.cpp
levonpetrosyan93: done

src/liblelantus/threadpool.h
levonpetrosyan93: done

src/liblelantus/threadpool.h
```diff
@@ -0,0 +1,109 @@
...
+                    // In case of shutdown thread list will be empty and destructor will wait for this thread completion
+                    std::thread::id currentId = std::this_thread::get_id();
+                    auto pThread = std::find_if(threads.begin(), threads.end(), [=](const std::thread &t) { return t.get_id() == currentId; });
+                    if (pThread != threads.end()) {
+                        pThread->detach();
+                        threads.erase(pThread);
+                    }
+                    break;
+                }
+                job = std::move(task_queue.front());
```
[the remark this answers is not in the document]
levonpetrosyan93: done

src/liblelantus/threadpool.h
```diff
@@ -0,0 +1,109 @@
...
+    std::queue<std::packaged_task<Result()>>  task_queue;
+    std::mutex                                task_queue_mutex;
+    std::condition_variable                   task_queue_condition;
+
+    bool                                      shutdown;
+    size_t const                              number_of_threads;
+
+    void ThreadProc() {
+        for (;;) {
+            std::packaged_task<Result()> job;
```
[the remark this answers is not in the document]
levonpetrosyan93: done

src/liblelantus/threadpool.h
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+public:
+    ParallelOpThreadPool(std::size_t thread_number) : number_of_threads(thread_number), shutdown(false) {}
+
+    ~ParallelOpThreadPool() {
+        std::list<std::thread> threadsToJoin;
+
+        {
+            std::unique_lock<std::mutex> lock(task_queue_mutex);
```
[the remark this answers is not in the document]
levonpetrosyan93: done

src/liblelantus/lelantus_prover.cpp
levonpetrosyan93: no, we need to fill i-th vector

levoncrypto (dismissed):

firstcryptoman (approved):
It looks good for me.

src/liblelantus/lelantus_prover.cpp
```diff
@@ -101,32 +102,70 @@ void LelantusProver::generate_sigma_proofs(
...
+                    return prover.sigma_commit(commits, index, rA_i, rB_i, rC_i, rD_i, a_i, Tk_i, Pk_i, Yk_i, sigma_i, proof);
+                }));
+            } else
+                break;
+        }
+
+        for (auto& th : parallelTasks) {
+            try {
+                th.get();
+            } catch (std::exception& except) {
```
psolstice, line 163: if exception other than `std::exception` is thrown from inside `prover.sigma_commit()` it'll be rethrown on `th.get()`. Imagine situation where the first task in the list results in exception which is not supposed to abort the program. We're getting out of `LelantusProver::generate_sigma_proofs` while all other tasks are still running and referencing local variables. This will result in segfault. If `sigma_commit()` can throw you need to wrap it into catch

psolstice (changes requested):

src/liblelantus/lelantus_prover.cpp
psolstice: also in case of client shutdown (even graceful!) all the threads are cancelled (`boost::thread_group::interrupt_all()`). In this case `th.get()` will throw boost exception, call frame will cease to exist and you'll get corrupted data or segfault

src/liblelantus/lelantus_prover.cpp
```diff
@@ -101,32 +103,78 @@ void LelantusProver::generate_sigma_proofs(
...
+        bool isFail = false;
+        for (auto& th : parallelTasks) {
+            try {
+                th.get();
+            } catch (std::exception& except) {
+                LogPrintf(except.what());
+                isFail = true;
+            }  catch (...) {
+                LogPrintf("Lelantus proof creation failed.");
+                isFail = true;
```
psolstice, line 170: look for docs there
https://www.boost.org/doc/libs/1_54_0/doc/html/thread/thread_management.html#thread.thread_management.tutorial.interruption

you can't use `catch(...)` here (without rethrowing at least). This this catch `boost::thread_interrupted` and then throw `std::runtime_error`. Meaning every time you gracefully shutdown there will be runtime error

please use the method from zerocoin code to block thread interruption while in function

psolstice (changes requested):

src/batchproof_container.cpp
```diff
@@ -159,39 +160,70 @@ void BatchProofContainer::erase(std::vector<LelantusSigmaProofData>* vProofs, co
...
-            fPadding.emplace_back(proofData.fPadding);
-            setSizes.emplace_back(proofData.anonymitySetSize);
-            proofs.emplace_back(proofData.sigmaProof);
-        }
+    else
+        return;
+    std::size_t threadsMaxCount = std::min((unsigned int)sigmaProofs.size(), boost::thread::hardware_concurrency());
+    std::vector<boost::future<bool>> parallelTasks;
+    parallelTasks.reserve(threadsMaxCount);
+    ParallelOpThreadPool<bool> threadPool(threadsMaxCount);
```
psolstice, line 168: DoNotDisturb here? and everywhere where ParallelOpThreadPool is used

psolstice (changes requested):

src/batchproof_container.cpp
levonpetrosyan93: We added multithreading in Lelantus-sigma roof creation code and batch proving code.
1 Added DoNotDisturb here for proof creation, I think we need it inside the function which is called in thread. https://github.com/firoorg/firo/pull/1058/files#diff-35b15f4d3b021c6dc41983092fa0d3666801234fe9f9ce790b41c65297a68483R32
2 Fir batching I don't think we need don't disturb as shutdown waits until batch verification finishes, [ ](https://github.com/firoorg/firo/blob/78e29d68b7354be702965fdd4f033709750776e5/src/init.cpp#L258)

src/batchproof_container.cpp
```diff
@@ -159,39 +160,71 @@ void BatchProofContainer::erase(std::vector<LelantusSigmaProofData>* vProofs, co
...
+                    serials.emplace_back(proofData.coinSerialNumber);
+                    fPadding.emplace_back(proofData.fPadding);
+                    setSizes.emplace_back(proofData.anonymitySetSize);
+                    proofs.emplace_back(proofData.sigmaProof);
+                }
 
-        auto params = sigma::Params::get_default();
-        sigma::SigmaPlusVerifier<Scalar, GroupElement> sigmaVerifier(params->get_g(), params->get_h(), params->get_n(), params->get_m());
+                parallelTasks.emplace_back(threadPool.PostTask([=]() {
+                    DoNotDisturb dnd;
```
psolstice, line 203: dnd should be in the same function frame that gets `std::future` result, not in lambda

psolstice (changes requested):

src/batchproof_container.cpp
levonpetrosyan93: done

psolstice (approved):

PR #1080, opened 2021-10-26
GUI Style and Colors

notshillo:
## PR intention
Update the GUI colors and styling


a-bezrukov:
image
This is HiDPI image (4k with 1.75 menu and bars scaling)

notshillo:
> image image
> 
> This is HiDPI image (4k with 1.75 menus and bars scaling)

apparently QSS and hiDPI don't mix well with font size. I'll look into it

notshillo:
As far as I can see the problem lies with element sizes set in the stylesheet which do not scale. My Qt experience is from before they introduced all the HiDPI scaling stuff so I am unsure of the best way forward. 

The tableview header columns can be dragged to size (which I also need to do with the original GUI if I set my monitor to use 175% scaling at 4k).

The toolbar buttons in the current code are auto-generated when the actions are created and added to the toolbar in bitcoingui.cpp. The size is then set in the stylesheet (which doesn't scale). Ideally the toolbar should be rebuilt as buttons in a horizontal layout which should then allow dynamic sizing based on button content size to occur. This would require a decent amount of work to rewrite. It may be possible to load separate stylesheets or modify values via regex based on the scaling factor.

Also, when I set my monitor to 4k and 175% scaling I do not get the same result as in the images above. Possibly due to different version of Qt? I am using 5.9 (from the depends build).

notshillo:
I'll read up some more and see if I can find an easier solution

notshillo:
Setting the font size made it scale properly for me under win10. 
The main toolbar item text will not get cut off anymore. Not an ideal fix as the buttons will not be of equal length (due to the limitation mentioned in https://github.com/firoorg/firo/pull/1080#issuecomment-956614987).

a-bezrukov:
It's a long living issue that when I start GUI now, I can see a QT warning in the console:
```
Attribute Qt::AA_EnableHighDpiScaling must be set before QCoreApplication is created.
```
If I move setting of this attribute higher in bitcoin.cpp:
```
diff --git a/src/qt/bitcoin.cpp b/src/qt/bitcoin.cpp
index db26437bb..2cecde71e 100644
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -658,7 +658,6 @@ int main(int argc, char *argv[])
     Q_INIT_RESOURCE(bitcoin);
     Q_INIT_RESOURCE(bitcoin_locale);
 
-    BitcoinApplication app(argc, argv);
 #if QT_VERSION > 0x050100
     // Generate high-dpi pixmaps
     QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
@@ -666,6 +665,8 @@ int main(int argc, char *argv[])
 #if QT_VERSION >= 0x050600
     QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
+
+    BitcoinApplication app(argc, argv);
 #ifdef Q_OS_MAC
     QApplication::setAttribute(Qt::AA_DontShowIconsInMenus);
 #endif
```
the GUI starts looking ways better.

notshillo:
Yes it should be above `BitcoinApplication app(argc, argv);`

I missed that warning as for some reason I get zero console output on win10 when running firo-qt.exe via cmd.exe or powershell.

a-bezrukov (approved):

PR #1085, opened 2021-11-15
Fixing the crashing issue on listaddressgroupings rpc

firstcryptoman:


levonpetrosyan93 (approved):

a-bezrukov (changes requested):
Sorry, how does this fix a crash?

levonpetrosyan93:
@a-bezrukov spend transaction types have no standard inputs so in case of spend we don't need to go inside if, but this call  `pcoin->tx->IsLelantusJoinSplit()` is added out of braces by mistake, me need to move it into braces, `HasNoRegularInputs()` function checkes it for all spend transactions.

a-bezrukov (approved):

PR #1089, opened 2021-11-22
Strip unneeded MTP data

psolstice:
## PR intention
MTP data takes ~200kB in block. These changes will reduce network traffic and disk space by ~56GB for nodes synced from scratch.

## Code changes brief
Code introduces concise form of block storage. If `hashMTPRoot` for MTP block is all zeroes there will be no MTP data in this block. Protocol version is bumped to 90031.

Before threshold time nothing is changed. After it blocks will be stripped of MTP data if possible. Sender with newer version will send stripped blocks if receiver has version >= 90031. Receiver with newer version will strip received block if full block is received. This PR doesn't cause a hardfork.


levonpetrosyan93 (approved):
Looks good to me.

PR #1111, opened 2021-12-22
Fix for window width increasing on Mac

levoncrypto:
## PR intention
This PR fixes an issue where the firo-qt app window was expanding infinitely on a Mac.


levonpetrosyan93 (approved):
Looks it works

PR #1129, opened 2022-01-14
Fixing lock icon view in coincontrol

firstcryptoman:
**What fixes the PR**
Padlock icon is invisible when the tx is locked and selected.
image


levonpetrosyan93 (approved):
This is not very pretty way, but as the colors for all icons is hardcoded here [link](https://github.com/firoorg/firo/pull/1129/files#diff-04ddda5e656ef0b20f1588b899cdc4ce1801dfe1237394e8eeb9516d9ac1bf27L44) ,
Just we did not want to break the design.

PR #1133, opened 2022-01-18
Faster header sync for ProgPoW

psolstice:
## PR intention
Header sync seemed to be very slow for ProgPoW. This PR fixes it.

## Code changes brief
There are two functions for checking PP hash: `GetProgPowHashFull` and `GetProgPowHashLight`. The former does proper check but is relatively slow. The latter doesn't check everything and fast. This PR uses light function during `CheckBlockHeader()` call and postpones full check until call to `ConnectBlock()`. This doesn't make the sync process any faster but it makes "syncing headers" stage of full sync take less time.

### DIFF
--- a/src/test/progpow_tests.cpp
+++ b/src/test/progpow_tests.cpp
@@ -45,11 +45,17 @@ struct ProgpowTestingSetup : public TestChain100Setup
         SetMockTime(0);
     }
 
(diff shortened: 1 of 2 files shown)


levonpetrosyan93 (approved):
LGTM

PR #1134, opened 2022-01-18
Fixing dropdown menu alignment in transaction tab

firstcryptoman:


src/qt/transactionview.cpp
```diff
@@ -71,7 +68,11 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     headerLayout->addWidget(watchOnlyWidget);
 
     instantsendWidget = new QComboBox(this);
-    instantsendWidget->setFixedWidth(24);
+    if (platformStyle->getUseExtraSpacing()) {
+        instantsendWidget->setFixedWidth(121);
+    } else {
```
levonpetrosyan93, line 73: Why we need this if/else here?

levonpetrosyan93 (approved):

levonpetrosyan93 (approved):

PR #1137, opened 2022-02-01
Explain new simpler process for building (depends) 

firo-sm:
Explain simple usage of the self hosted dependency building (depends) and promote this over manually building dependencies, which doesn't really work now anyway due to some versioning issues on most platforms.

Since the primary build instructions are now in the top level README.md there is a note added to build-unix.md to point anybody looking there back to the top.

## PR intention
There were a few issues with the documentation that can be easily fixed:
- The top level README linked to the depends build instructions but still primarily encouraged users to build manually. But the manual build instructions no longer work as described on newer systems. The depends system should be the primary promoted method, with the manual methods as a fallback instead of the other way around.
- The build-unix.md docs which is where most people look did not refer to the depends build process at all.
- Headless builds using depends needs some extra explanation otherwise it can break on some missing QT dependencies.
- For using the depends build, a bunch of extra system libraries were still recommended that are provided by the depends system anyway.

## Code changes brief
- The change points people looking at the normal build-unix.md to the top level, where the instructions have been moved to.
- Explains how to install just the required system dependencies for compiling, including for some additional distributions.
- Explains how to properly build for headless or GUI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,51 +79,70 @@ docker start firod
 Linux Build Instructions and Notes
 ==================================
 
-Dependencies
+Firo contains build scripts for its dependencies to ensure all component versions are compatible. For additional options
+such as cross compilation, read the [depends instructions](depends/README.md)
+
+Alternatively, you can build dependencies manually. See the full [unix build instructions](doc/build-unix.md).
+
+Development Dependencies (compiler and build tools)
 ----------------------
-1.  Update packages
 
-        sudo apt-get update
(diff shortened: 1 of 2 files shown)


AaronFeickert (approved):
Tested the top-level `depends` build instructions on a fresh Ubuntu 20.04.3 virtual machine. Instructions were correct for a headless build with tests enabled, and for a GUI build.

psolstice:
Merged changes into #1138 , closing this one now

PR #1144, opened 2022-02-18
Improve performance on macOS

psolstice:
## PR intention
This PR makes one particular optimisation that should improve performance on all the platforms, but it affects macOS the most.

## Code changes brief
Removed unnecessary copying of `Consensus::Params()`


levonpetrosyan93 (approved):
LGTM

Issue #1122, opened 2022-01-08
Group element deserialization accepts invalid values


AaronFeickert:
Group element deserialization in `GroupElement::deserialize` ignores the return value of this function call: https://github.com/firoorg/firo/blob/master/src/secp256k1/src/cpp/GroupElement.cpp#L528

On invalid input, this results in a `GroupElement` object being assigned an invalid underlying value. It can be mitigated by calling `isMember` on the object, but this is not documented.

Deserialization should throw an exception on invalid input, which matches the behavior of the `Scalar::deserialize` operation.

PR #1147, opened 2022-02-21
Added validity check in GroupElement deserialization

levonpetrosyan93:
fix for #1122

### DIFF
--- a/src/bip47/bip47utils.cpp
+++ b/src/bip47/bip47utils.cpp
@@ -168,7 +168,11 @@ GroupElement GeFromPubkey(CPubKey const & pubKey)
     std::copy(pubKey.begin()+ 1, pubKey.end(), std::back_inserter(serializedGe));
     serializedGe.push_back(*pubKey.begin() == 0x02 ? 0 : 1);
     serializedGe.push_back(0x0);
-    result.deserialize(&serializedGe[0]);
+    try {
+        result.deserialize(&serializedGe[0]);
+    } catch (...) {
+        result = GroupElement();
+    }
     return result;
 }
 
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -398,13 +398,21 @@ public:
 
         for (const auto& str : lelantus::lelantus_blacklist) {
             GroupElement coin;
-            coin.deserialize(ParseHex(str).data());
+            try {
+                coin.deserialize(ParseHex(str).data());
+            } catch (...) {
+                continue;
+            }
             consensus.lelantusBlacklist.insert(coin);
         }
 
         for (const auto& str : sigma::sigma_blacklist) {
             GroupElement coin;
-            coin.deserialize(ParseHex(str).data());
+            try {
+                coin.deserialize(ParseHex(str).data());
+            } catch (...) {
+                continue;
+            }
             consensus.sigmaBlacklist.insert(coin);
         }
 
@@ -677,7 +685,11 @@ public:
 
         for (const auto& str : lelantus::lelantus_testnet_blacklist) {
             GroupElement coin;
-            coin.deserialize(ParseHex(str).data());
+            try {
+                coin.deserialize(ParseHex(str).data());
+            } catch (...) {
+                continue;
+            }
             consensus.lelantusBlacklist.insert(coin);
         }
 
--- a/src/lelantus.cpp
+++ b/src/lelantus.cpp
@@ -384,6 +384,12 @@ bool CheckLelantusJoinSplitTransaction(
             REJECT_MALFORMED,
             "CheckLelantusJoinSplitTransaction: invalid joinsplit transaction");
     }
+    catch (...) {
+        return state.DoS(100,
+                         false,
+                         REJECT_MALFORMED,
+                         "CheckLelantusJoinSplitTransaction: failed to deserialize joinsplit");
+    }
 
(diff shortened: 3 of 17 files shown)


AaronFeickert:
Recommend adding a unit test to trigger validity check failure.

src/bip47/bip47utils.cpp
```diff
@@ -168,7 +168,11 @@ GroupElement GeFromPubkey(CPubKey const & pubKey)
     std::copy(pubKey.begin()+ 1, pubKey.end(), std::back_inserter(serializedGe));
     serializedGe.push_back(*pubKey.begin() == 0x02 ? 0 : 1);
     serializedGe.push_back(0x0);
-    result.deserialize(&serializedGe[0]);
+    try {
+        result.deserialize(&serializedGe[0]);
+    } catch (...) {
+        result = GroupElement();
```
AaronFeickert, line 174: Has it been determined that defaulting to the zero element is safe in this context?

src/secp256k1/src/cpp/GroupElement.cpp
```diff
@@ -529,6 +529,10 @@ const unsigned char* GroupElement::deserialize(const unsigned char* buffer) {
     result.infinity = (int)infinity;
 
     secp256k1_gej_set_ge(reinterpret_cast<secp256k1_gej *>(g_), &result);
+
+    if (!secp256k1_ge_is_valid_var(&result)) {
+        throw "GroupElement: deserialize failed";
```
AaronFeickert, line 534: Would be useful to have a unit test showing this works as expected.

AaronFeickert (changes requested):

src/bip47/bip47utils.cpp
levonpetrosyan93: Yes, it just will generate invalid address, which also would be possible previously.

src/secp256k1/src/cpp/GroupElement.cpp
levonpetrosyan93: done

src/chainparams.cpp
```diff
@@ -398,13 +398,21 @@ class CMainParams : public CChainParams {
 
         for (const auto& str : lelantus::lelantus_blacklist) {
             GroupElement coin;
-            coin.deserialize(ParseHex(str).data());
+            try {
+                coin.deserialize(ParseHex(str).data());
+            } catch (...) {
+                continue;
```
AaronFeickert, line 404: Is this `catch` reached given the current chain?

src/chainparams.cpp
```diff
@@ -398,13 +398,21 @@ class CMainParams : public CChainParams {
...
             consensus.lelantusBlacklist.insert(coin);
         }
 
         for (const auto& str : sigma::sigma_blacklist) {
             GroupElement coin;
-            coin.deserialize(ParseHex(str).data());
+            try {
+                coin.deserialize(ParseHex(str).data());
+            } catch (...) {
+                continue;
```
AaronFeickert, line 414: Is this `catch` reached given the current chain?

src/lelantus.cpp
```diff
@@ -698,18 +704,6 @@ bool CheckLelantusTransaction(
 {
     Consensus::Params const & consensus = ::Params().GetConsensus();
 
-
-    if(tx.IsLelantusJoinSplit()) {
-        CAmount nFees;
-        try {
-            nFees = lelantus::ParseLelantusJoinSplit(tx)->getFee();
```
AaronFeickert, line 705: Is this set anywhere else, since it's now removed?

src/sigma/test/serialize_test.cpp
```diff
@@ -19,6 +19,25 @@ BOOST_AUTO_TEST_CASE(group_element_serialize)
...
+    unsigned char buffer [initial.memoryRequired()];
+    initial.serialize(buffer);
+    //Shift all elements to right by 2, to make it invalid
+    for (int i = initial.memoryRequired() - 1; i > 1; i--)
+        buffer[i] = buffer[i - 2];
+    buffer[0] = std::rand();
+    buffer[1] = std::rand();
+
+    // Making the point not infinity as it is allowed in deserialization
+    buffer[33] = 0;
```
AaronFeickert, line 35: Are there nonstandard representations of the zero point that need to be checked for validity?

src/chainparams.cpp
levonpetrosyan93: no, added just in case,

src/chainparams.cpp
levonpetrosyan93: the same

src/sigma/test/serialize_test.cpp
levonpetrosyan93: I don't think there are such representations.

src/lelantus.cpp
levonpetrosyan93: nFees value is not used in this context, it was being used previously, and missed to remove this during refactorin

AaronFeickert (approved):
utACK

src/lelantus.cpp
reubenyap: Was this resolved?

AaronFeickert: Relying on @levonpetrosyan93's comment that it is unused.

PR #1151, opened 2022-03-09
Update ci-master.yml

arjundashrath:
GitHub asks users to define workflow permissions, see https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ and https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token for securing GitHub workflows against supply-chain attacks.

The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. 

The `Token-Permissions` category has a score of 0/10 in Scorecards.

This file was fixed automatically using the open-source tool https://github.com/step-security/secure-workflows. If you like the changes and merge them, please consider starring the repo.

### DIFF
--- a/.github/workflows/ci-master.yml
+++ b/.github/workflows/ci-master.yml
@@ -8,6 +8,9 @@ on:
     - master
 env:
   SOURCE_ARTIFACT: source
+permissions:
+  contents: read
+
 jobs:
   create-source-distribution:
     name: Create Source Distribution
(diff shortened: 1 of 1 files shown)


levonpetrosyan93:
I don't think we need this, as I see  [here](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) that "Any permission that is absent from the list will be set to none.", that meas all permissions have been set `none` before.

Issue #1112, opened 2021-12-23
Minimum height of QT wallet is greater than my screen


xorzero777:
_**Minimum height of QT wallet is greater than my screen and resize cannot reduce. Also there is no full screen view. Maybe consider putting sliders on the side of each tab?**_

![image](https://user-images.githubusercontent.com/42816019/147263522-16df880c-8852-4012-9434-dcfb4f63c4a6.png)

- v0.14.9.2-g738308bfc

- OS: Ubuntu 20.04 (Gnome)

KuroGuo:
I think the font size in css is too big

xorzero777:
Does QT graphics use CSS?

KuroGuo:
> Does QT graphics use CSS?

Yes. https://github.com/firoorg/firo/blob/master/src/qt/res/css/firo.css#L15

xorzero777:
Ok new style QT .. I wrote/used QT before but the old style.

xorzero777:
Ok so how do I fix my Issue ...

KuroGuo:
> Ok so how do I fix my Issue ...

Wait for the developers to fix it

xorzero777:
Ok .. thanks for clarification ;-)

KuroGuo:
I found that the font size in [css](https://github.com/firoorg/firo/commits/master/src/qt/res/css/firo.css) was changed twice, the two developers seem to be unaware of each other, maybe 10pt is more appropriate.

![2c5259d499e8a342a61d62e507040ea](https://user-images.githubusercontent.com/8684548/150287513-cb97419e-e06c-4fb0-ad0b-280b00927dff.png)

WonderLand33:
same to me

justanwar:
Fixed in #1152

xorzero777:
![image](https://user-images.githubusercontent.com/42816019/163557836-101745d8-eaec-4f0e-996b-7c5324889b73.png)
For me this showed in 0.14.9.5 ... 

OS: Ubuntu 20.04 (Gnome)
v0.14.9.5-gb764c08df

Although I cannot see the lowest extent of the window .. this is now usable so thanks and no complaint ;-)

levonpetrosyan93:
> ![image](https://user-images.githubusercontent.com/42816019/163557836-101745d8-eaec-4f0e-996b-7c5324889b73.png) For me this showed in 0.14.9.5 ...
> 
[...]

PR #1152, opened 2022-03-11
Fixed minimum height issue

levonpetrosyan93:
Fix for issue #1112

### DIFF
--- a/src/sigma/test/serialize_test.cpp
+++ b/src/sigma/test/serialize_test.cpp
@@ -21,21 +21,10 @@ BOOST_AUTO_TEST_CASE(group_element_serialize)
 
 BOOST_AUTO_TEST_CASE(group_element_invalid)
 {
-    secp_primitives::GroupElement initial;
-    initial.randomize();
-    unsigned char buffer [initial.memoryRequired()];
-    initial.serialize(buffer);
-    //Shift all elements to right by 2, to make it invalid
-    for (int i = initial.memoryRequired() - 1; i > 1; i--)
-        buffer[i] = buffer[i - 2];
-    buffer[0] = std::rand();
-    buffer[1] = std::rand();
-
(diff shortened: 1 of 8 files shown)


justanwar:
Tested. Main wallet window minimum height reduced but recovery UI is still tall. 

Need testing on 1366x768 screens.

levonpetrosyan93:
Tested on 1280x800

justanwar:
Tested on 1280x768, mnemonic recovery screen shows fully.

src/sigma/test/serialize_test.cpp
```diff
@@ -21,21 +21,10 @@ BOOST_AUTO_TEST_CASE(group_element_serialize)
...
-    for (int i = initial.memoryRequired() - 1; i > 1; i--)
-        buffer[i] = buffer[i - 2];
-    buffer[0] = std::rand();
-    buffer[1] = std::rand();
-
-    // Making the point not infinity as it is allowed in deserialization
-    buffer[33] = 0;
-
+    // Invalid GroupElement generated in advance
+    std::string buffer = "956f89850d79edd157360bb8608b8a336ca6e00dc057b4182540db5536745ef57b00";
```
AaronFeickert, line 25: Can you provide an explanation for how this particular invalid serialization was derived, and why it is suitable to test the desired functionality?

levonpetrosyan93: Previous test code sometimes failed there is possibility to generate valid point from such semi-random string, I generated an invalid point checked that, and put it here.

AaronFeickert: Sure, but what makes it a good test candidate? I'm of the opinion that magic numbers should be easily explainable or reproducible where possible.

KuroGuo:
Tested on 1280x768 (Windows 11, zh-CN), mnemonic recovery screen shows fully.

image

Issue #1163, opened 2022-04-15
Lelantus mint crashes wallet


spointerz:
When Lelantus mint dust smaller than fee, amount is negative and wallet will crash and corrupt the chaindata(mine did).
TODO; add mint limit like 1 coin.

levonpetrosyan93:
Please give the steps how it can be reproduced.

reubenyap:
Closing due to inactivity.

PR #1174, opened 2022-05-09
Catch other exceptions from Joinsplit creation

levonpetrosyan93:
Possible fix for #1163

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3557,6 +3557,9 @@ UniValue joinsplit(const JSONRPCRequest& request) {
     catch (const std::exception& e) {
         throw JSONRPCError(RPC_WALLET_ERROR, e.what());
     }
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -3557,6 +3557,9 @@ UniValue joinsplit(const JSONRPCRequest& request) {
     catch (const std::exception& e) {
         throw JSONRPCError(RPC_WALLET_ERROR, e.what());
     }
+    catch (...) {
```
psolstice, line 3560: `catch(...)` will catch SEGFAULTs on some systems (Windows comes to mind). It's probably not a great idea to catch this because ignoring SEGFAULTs can have memory corruption as a side effect. And potentially turn DDoS attack vector into RCE

psolstice (changes requested):

Issue #1170, opened 2022-04-28
Allow different masternode ports


mksdev:
When trying to register master node on different port than default 8168 there is hardcoded check introduced in https://github.com/firoorg/firo/commit/8d54a4e33b069a9f3f95c4967742a158fe01318d which ends up in error if trying to use different one.
It would be nice to be able to run masternode on non-default port in case port is already taken by different service, or when running multiple masternodes behind one public ip.

Since all neccesary data is already stored on blockchain, it should be enough to just revert this commit (from my understanding by skimming through the codebase).

skironDotNet:
They won't do it. They are CROWN developers and few orders (except PIVX) stick to this sick idea that if you just multiple MN on same VPS you threat the network on case you VPS goes down.
To me this is such a full of shit argument. I host 12 MNs of PIVX thanks to them allowing IPv6. If my VPS goes down then it maybe caused by data center going down and then we have same problem even if paid for 12 VPS with each single IPv4.

I think your argument is valid about port being taken but also invalid because what's on your server that the port is taken?
If you want to host multiple MNs from single VPS you should be using same port but different IP v6 (free and available)

Now to you developers. Forcing people to host from IPv4 is bigger threat than the bs one about VPS going down. 
Single FIRO instance takes 2.5GB of ram, so on 32GB ram you can host only 12 nodes. If they all go down let say due to server restart that's not a threat having other 4000 in the network. Now why ipv4 is bigger threat. Ipv4 is expensive, once the coin dumps in price it's not worth to host MN, this forces people to abandon the idea and then you really losing MNs.

Look at CRWN, hosting SN gives 50% APY but the coin is do cheap it can't pay for cheapest VPS. So my advice is allow IPv6 ASAP to make sure people will create nodes. The major cost of hosting is IP i can get 16GB RAM VPS from certain hosting company for $10 but only 2 IPv4 attached where 1st is free (part of vps cost) an the other cost 3 Euro a month. While i have whole /64 range of IPv6 addresses. So what's easier to host FIRO or PIVX?
Current price of FIRO $1.8 makes it bearly worth the cost + management time. So if it keeps dumping even current owner will sell their 1K of FIRO and you are gone

mksdev:
I am not using any datacenter VPS, but running self hosted servers from multiple locations. It seems like everyone automatically assumes (even MN tutorials) that you go and rent VPS. When running master nodes from small office or home there is inherent problem that you generally cannot get IPv6 or multiple IPv4. 

[...]

PR #1179, opened 2022-05-31
Enable masternodes to run on non-default port.

mksdev:
This reverts commit 8d54a4e33b069a9f3f95c4967742a158fe01318d. Related issue #1170.

## PR intention
Enable master node to be registered and run on non-default port (mainnet or not). Address and port is stored in registration transaction and correctly used to connect to the master node. This just enables transactions with non-default port to be registered.

### DIFF
--- a/src/evo/providertx.cpp
+++ b/src/evo/providertx.cpp
@@ -27,15 +27,6 @@ static bool CheckService(const uint256& proTxHash, const ProTx& proTx, CValidati
         return state.DoS(10, false, REJECT_INVALID, "bad-protx-addr");
     }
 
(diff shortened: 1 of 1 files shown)


reubenyap:
Could you explain the rationale for this?

mksdev:
Anything else that I should provide for this pull request to unstuck it?

SPC-beta:
this would create a network fork

mksdev:
it could be included in next mandatory release, same as 14.10, which also breaks all previous versions and basically forks the chain

reubenyap:
See Issue #1170

PR #1180, opened 2022-05-31
First iteration of Spark integration

levonpetrosyan93:


src/libspark/keys.cpp
```diff
@@ -12,6 +13,19 @@ SpendKey::SpendKey(const Params* params) {
...
 }
 
+SpendKey::SpendKey(const Params* params, const Scalar& r_) {
+    this->params = params;
+    this->r = r_;
+    std::vector<unsigned char> data;
+    r.serialize(data.data());
+    std::vector<unsigned char> result(CSHA512().OUTPUT_SIZE);
+    CHash512 hash512;
+    hash512.Write(data.data(), data.size()).Finalize(&result[0]);
```
AaronFeickert, line 23: Recommend using domain separation for this hash instantiation to mitigate any risk of collision.

src/libspark/keys.cpp
```diff
@@ -12,6 +13,19 @@ SpendKey::SpendKey(const Params* params) {
...
+
+    this->s1.memberFromSeed(&result[0]);
+    this->s2.memberFromSeed(&result[32]);
```
AaronFeickert, line 26: If desired, it is also safe (with appropriate domain separation and any other KDF considerations) to hash `r` to obtain `s2`, and then hash `s2` to obtain `s1`.

AaronFeickert (changes requested):

src/libspark/keys.cpp
```diff
@@ -115,4 +134,47 @@ const GroupElement& Address::get_Q2() const {
...
+    ss << std::hex;
+    for (const auto b : buffer) {
+        ss << (b >> 4);
+        ss << (b & 0xF);
+    }
+
+    return ss.str();
+}
+
+void Address::SetHex(const std::string& str) {
```
sproxet, line 175: In the event that `str` is invalid and an exception is thrown, `this` will be trampled.

src/spark.cpp
```diff
@@ -0,0 +1,9 @@
+#include "spark.h"
+
+namespace spark {
+
+bool GetOutPoint(COutPoint& outPoint, const spark::Coin coin) {
```
sproxet, line 5: This function is called in a hot loop in `CWallet::GetAvailableSparkCoins`. Equivalent Lelantus functions perform a disk read here.

src/wallet/wallet.cpp
```diff
@@ -2804,6 +2812,90 @@ CRecipient CWallet::CreateLelantusMintRecipient(
...
+    }
+
+    std::set<COutPoint> lockedCoins = setLockedCoins;
+
+    // Filter out coins that have not been selected from CoinControl should that be used
+    coins.remove_if([lockedCoins, coinControl](const std::pair<spark::Coin, CSparkMintMeta>& coin) {
+        COutPoint outPoint;
+
+        // ignore if the coin is not actually on chain
+        if (!spark::GetOutPoint(outPoint, coin.first)) {
```
sproxet, line 2840: If this is like the equivalent Lelantus call, it will take a long time and make this call require a disk read for every Spark coin in our wallet.

src/libspark/keys.cpp
levonpetrosyan93: What you mean trampled? Anyway you should handle all cases and catch exceptions.

src/spark.cpp
levonpetrosyan93: I see your point, actually for Lelantus much time takes to get mints from db, reading block from disk is very fast compared to reading data from db, I will suggest to keep this for now, and try to improve during testnet testing.

levonpetrosyan93: For spark it is not looking into db every time, we keep mints in memory, and only touch db during adding or removing coins.

PR #1185, opened 2022-06-27
Adjustment for block numbers in consensus following faster blocks hardfork

psolstice:
## PR intention
After PR #1176 any planned change to the network based on block number will come sooner than before because blocks come faster now. This PR fixes this.

## Code changes brief
Changes were made to the following consensus parameters:
- height of the second halving
- number of blocks between halvings starting from the second one
- height of the block at which block has no subsidy anymore
- number of block for spork functionality to stop

Also minimum peer version is set to 0.14.10


levonpetrosyan93 (approved):
LGTM

PR #1186, opened 2022-06-27
Use QT5 connect new syntax

firstcryptoman:
Changing qt signal-slot connect syntax from String-Based to Functor-Based

What's the problem with String-Based syntax?

1. No compile time check: All the checks are done at run-time by parsing the strings. That means if you do a typo in the name of the signal or the slot, it will compile but the connection will not be made, and you will only notice a warning in the standard output.

2. Since it operates on the strings, the type names of the slot must match exactly the ones of the signal. And they also need to be the same in the header and in the connect statement. This means it won't work nicely if you want to use typedef or namespaces

Pros for new syntax

1. The ability to compile time check of the existence of the signals and slot, of the types, or if the Q_OBJECT is missing, rather than in the runtime

2. Reducing compilation time and final binary size.

The new syntax requires MOC for signals to work, but not for slots. If a class only has slots and no signals, it need not have the Q_OBJECT macro and hence may not invoke the MOC, which not only reduces the final binary size but also reduces compilation time (no MOC call and no subsequent compiler call for the generated *_moc.cpp file).

3. Argument can be by typedefs or with different namespace specifier, and it works.

4. The ability to connect to any member function of QObject, not only slots and lambda functions.

5. The SIGNAL and SLOT macros ignore the access levels of methods, so new syntax protect ourselves from errors when we try to connect from the outside to a private slot

Cons for new syntax

1. Need to specify the type of your object(In my opinion its good to explisitly know the type of object in code)

2. Complicated syntax in cases of overloads(qOverload helps)
Example:

        connect(mySpinBox, SIGNAL(valueChanged(int)), mySlider, SLOT(setValue(int));

cannot be simply converted to:

        connect(mySpinBox, &QSpinBox::valueChanged, mySlider, &QSlider::setValue);

because QSpinBox has two signals named valueChanged() with different arguments. Instead, the new code needs to be:

        connect(mySpinBox, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), mySlider, &QSlider::setValue);

but with qOverload it we can write it like this:

        connect(mySpinBox, qOverload<int>(&QSpinBox::valueChanged), mySlider, &QSlider::setValue);

so its not very hard syntax.

3. Default arguments in slot is not supported anymore.


levonpetrosyan93 (approved):

psolstice (dismissed):

justanwar:
To sync with v0.14.11.1, enabling testing.

justanwar:
On the latest commit, the following problem was observed:
Transactions: double-clicking a transaction in the list does not show transaction details

justanwar:
Tested the following, they working as expected:

Anonymize all
Send (private/transparent)
Receive (address generation/address removal)
Transactions
Masternodes list (all/my masternodes)
Lelantus (manual anonymization)
RAP Address (creation)
Debug  Window - Information/Console/Network traffic/Peers
Coin Control (send private/send transparent/Lelantus)

PR #1198, opened 2022-07-30
Fix Various Instances of Undefined Behaviour

sproxet:
## PR intention
This fixes several instances of UB.

## Code changes brief
- `CQuorumBlockProcessor::GetCommitmentsFromBlock` - std::move() was used incorrectly, leading to a use after free issue.
- bls - An uninitialised boolean value is read.
- `CDKGSessionHandler::SleepBeforePhase` - An invalid implicit cast from `UINT64_MAX` to `double` is performed in some cases.

### DIFF
--- a/src/bls/bls.h
+++ b/src/bls/bls.h
@@ -42,7 +42,7 @@ class CBLSWrapper
     friend class CBLSPublicKey;
     friend class CBLSSignature;
 
-    bool fLegacy;
+    bool fLegacy{fLegacyDefault};
 
 protected:
     ImplType impl;
--- a/src/llmq/quorums_blockprocessor.cpp
+++ b/src/llmq/quorums_blockprocessor.cpp
@@ -321,7 +321,8 @@ bool CQuorumBlockProcessor::GetCommitmentsFromBlock(const CBlock& block, const C
                 return state.DoS(100, false, REJECT_INVALID, "bad-qc-dup");
             }
 
-            ret.emplace((Consensus::LLMQType)qc.commitment.llmqType, std::move(qc.commitment));
+            auto llmqTypeTemp = (Consensus::LLMQType)qc.commitment.llmqType;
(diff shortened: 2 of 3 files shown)


levonpetrosyan93 (approved):

src/bls/bls.h
```diff
@@ -42,7 +42,7 @@ class CBLSWrapper
     friend class CBLSPublicKey;
     friend class CBLSSignature;
 
-    bool fLegacy;
+    bool fLegacy{false};
```
psolstice, line 45: I don't see a code path that leaves `fLegacy` uninitialized. And if anything it should be assigned to `fLegacyDefault`, not to false

levonpetrosyan93 (dismissed):

src/llmq/quorums_blockprocessor.cpp
```diff
@@ -321,7 +321,8 @@ bool CQuorumBlockProcessor::GetCommitmentsFromBlock(const CBlock& block, const C
                 return state.DoS(100, false, REJECT_INVALID, "bad-qc-dup");
             }
 
-            ret.emplace((Consensus::LLMQType)qc.commitment.llmqType, std::move(qc.commitment));
+            auto temp = (Consensus::LLMQType) qc.commitment.llmqType;
```
psolstice, line 324: don't name variable `temp` please. `llmqType` is much better name

psolstice (changes requested):

src/llmq/quorums_blockprocessor.cpp
sproxet: Use of the name temp is an indicator that the variable is semantically required. Can document in a comment instead if that's preferable.

psolstice (approved):

levonpetrosyan93 (approved):

PR #1205, opened 2022-09-26
Fixed compilation issues on newer macOS versions

psolstice:
## PR intention
Fix compilation issues on macOS 12.6

## Code changes brief
`tor` binaries can't be compiled, so the patch was made to ensure only libraries are built


levonpetrosyan93 (approved):
LGTM

PR #1218, opened 2022-12-09
Review fixes

AaronFeickert:
Addresses issues FLS-01, FLS-02, FLS-03, FLS-07 from the HashCloak review.

### DIFF
--- a/src/libspark/bpplus.cpp
+++ b/src/libspark/bpplus.cpp
@@ -137,7 +137,7 @@ void BPPlus::prove(
     std::vector<GroupElement> A_points;
     std::vector<Scalar> A_scalars;
     A_points.reserve(2*N*M + 1);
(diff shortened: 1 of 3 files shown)


onurinanc:
For the FLS-02, you're only making the checks inside the BPPlus::verify(). You should also make the checks inside BPPlus::proof()

AaronFeickert:
> For the FLS-02, you're only making the checks inside the BPPlus::verify(). You should also make the checks inside BPPlus::proof()

Yep. It was unclear how adding those failure modes to the prover would have other effects elsewhere in the codebase. It's not a security issue for the prover to generate such a proof anyway, since the verifier will reject it. And the failure mode can't be triggered in practice anyway, since it's effectively a hash function preimage attack.

PR #1238, opened 2023-05-13
Fix undefined behaviour in CHDChain.

ghost:
## PR intention
Fix instances of UB in CHDChain.

## Code changes brief
- Give safe defaults for variables.
- Prevent `nExternalChainCounters` from being initialised with uninitialised values when it is deserialised.

### DIFF
--- a/src/wallet/walletdb.h
+++ b/src/wallet/walletdb.h
@@ -71,18 +71,23 @@ typedef std::pair<uint32_t,bool> Component;
 class CHDChain
 {
 public:
-    uint32_t nExternalChainCounter; // VERSION_BASIC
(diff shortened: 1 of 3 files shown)


src/wallet/walletdb.h
```diff
@@ -102,9 +107,7 @@ class CHDChain
         nVersion = CHDChain::CURRENT_VERSION;
         masterKeyID.SetNull();
         nExternalChainCounter = 0;
-        for(int index=0;index<N_CHANGES;index++){
-            nExternalChainCounters.push_back(0);
-        }
+        nExternalChainCounters = std::vector<uint32_t>(N_CHANGES, 0);
```
psolstice, line 110: `std::vector::assign()` is better function to call here

psolstice (changes requested):

PR #1244, opened 2023-05-19
Integrate spark into qt

giangnamnabka:
Integrate spark into qt


levonpetrosyan93:
@giangnamnabka Please rebase onto spark_ branch

PR #1249, opened 2023-05-24
Integrate spark into Qt

giangnamnabka:
integrate spark

### DIFF
--- a/src/spark/state.cpp
+++ b/src/spark/state.cpp
@@ -638,6 +638,9 @@ bool CheckSparkSpendTransaction(
         while (index != coinGroup.firstBlock && index->GetBlockHash() != idAndHash.second)
             index = index->pprev;
 
+        // take the hash from last block of anonymity set
+        std::vector<unsigned char> set_hash = GetAnonymitySetHash(index, idAndHash.first);
+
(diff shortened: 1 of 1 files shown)


src/qt/transactionrecord.cpp
```diff
@@ -229,6 +229,12 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
                 // Mint to self
                 parts.append(TransactionRecord(hash, nTime, TransactionRecord::Anonymize, "",
                     -(nDebit - nChange), 0));
+            } else if (wtx.tx->IsSparkMint()) {
+                parts.append(TransactionRecord(hash, nTime, TransactionRecord::MintSparkToSelf, "",
+                    -(nDebit - nChange), 0));
+            } else if (wtx.tx->IsSparkSpend()) {
```
levonpetrosyan93, line 235: You need to get nTxFee for spark spend transaction like it is done for lelantus joinsplit, at line 67, as spark amounts are shielded, you are unable to calculate fee by inputs - outputs.

giangnamnabka: Updated. Thanks!

PR #1251, opened 2023-06-02
Fix docker build crash on Fedora system 

mksdev:
Fixes crash in final build step when libraries from build image are copied to the final image.


erciccione (approved):

levonpetrosyan93 (approved):
LGTM

justanwar (approved):

PR #1254, opened 2023-06-09
minor repo cleanup

erciccione:
remove INSTALL.md, _config.yml and lgtm.yml
remove broken badges from readme
adapt contributing to firo
edit doc/README.md
remove not needed files in /doc


doc/files.md
reubenyap, line 1: What's the reasoning behind the deletion of this file? No particular objections but just wanted tot understand.

erciccione: Looks like it's an old file mostly valid for very old versions of bitcoin core. If in doubt we can keep it IMO.

justanwar: We can also keep `bips.md`.

erciccione: @justanwar  My goal here is to remove as many references to Bitcoin as possible, because i see the risk of random visitors and/or developers interested in Firo dropping their interests if they start thinking of Firo as a copycat of Bitcoin with something added on top. That would be a distortion of what Firo is.

Removing references to Bitcoin and Bitcoin-specific documentation from the repo is a small, but necessary effort to improve the image of the project and overall look of the repo. If i'm not mistaken, BIPS are very specific to bitcoin and have no role in Firo, that's why i was removing `bips.md`. Of course if keeping it is preferred i have no problem with reverting.

justanwar (approved):

reubenyap (approved):

PR #1255, opened 2023-06-10
Get compilation working on latest OSX.

sproxet:
This fixes issues prevent compilation on the latest Macs. It upgrades Boost.


reubenyap:
Fixed in #1265

PR #1260, opened 2023-06-19
Add the -checkblockindexpow flag, reducing startup time by 70%.

sproxet:
Stop checking PoW of already blocks that we've already persisted (and therefore have already checked the PoW of). Add a flag -checkblockindexpow if we really want to do this anyway.

This cuts startup time by around 70%.


reubenyap:
Superseded by #1420

PR #1261, opened 2023-06-20
Add CMakeLists.txt for easier IDE usage.

sproxet:
This file allows CLion to pick up all our source files and dependencies. It has no effect otherwise.


psolstice:
If you can come with a real CMake file that works, that would be a different story. But fake one is not a good thing to have especially considering we need to maintain it to keep it up to date. You can keep it locally (like some keep local VS code scripts)

PR #1263, opened 2023-07-06
Integrate Spark into QT

giangnamnabka:
Integrate Spark into QT


src/qt/bitcoinaddressvalidator.cpp
```diff
@@ -97,5 +96,22 @@ QValidator::State BitcoinAddressCheckValidator::validate(QString &input, int &po
...
+
+bool BitcoinAddressCheckValidator::validateSparkAddress(const std::string& address) const
+{
+    const spark::Params* params = spark::Params::get_default();
+    unsigned char network = spark::GetNetworkType();
+    unsigned char coinNetwork;
+    spark::Address addr(params);
+    try {
+        coinNetwork = addr.decode(address);
+    } catch (...) {
```
levonpetrosyan93, line 113: Better to catch std::invalid_argument here

src/qt/automintnotification.h
```diff
@@ -36,4 +37,30 @@ private Q_SLOTS:
     LelantusModel *lelantusModel;
 };
 
+class AutomintSparkNotification : public QDialog
```
levonpetrosyan93, line 40: Why we need separate type here? We can just check if Spark already active set spark model instead of lelantus model,

src/libspark/coin.cpp
```diff
@@ -56,9 +56,9 @@ Coin::Coin(
 	// Type-specific elements
 	//
 
+	this->v = v;
+	this->address = address;
```
levonpetrosyan93, line 60: I am not agree to do this change, Very risky to add v and address into Spark::Coin, even if we don't serailize them,  wallet knows this data inside CSparkMintMeta, please use it to get this data, and revert this two lines.

src/libspark/coin.h
```diff
@@ -102,6 +102,7 @@ class Coin {
 	GroupElement S, K, C; // serial commitment, recovery key, value commitment
 	AEADEncryptedData r_; // encrypted recipient data
 	uint64_t v; // value
+	Address address;
```
levonpetrosyan93, line 105: Remove this too.

src/qt/automintdialog.h
```diff
@@ -51,4 +51,44 @@ private Q_SLOTS:
     void ensureLelantusModel();
 };
 
+enum class AutoMintSparkMode : uint8_t {
```
levonpetrosyan93, line 54: Use above classes instead of copy-pasting the same code.

src/qt/automintdialog.h
```diff
@@ -51,4 +51,44 @@ private Q_SLOTS:
...
+    MintAll, // come from overview page
+    AutoMintAll // come from notification
+};
+
+class AutoMintSparkDialog : public QDialog
```
levonpetrosyan93, line 59: The same here, Just add there spark related thins and check when use Lelantus related parts when spark, no need to have separate Dialogs.

src/qt/automintmodel.h
```diff
@@ -111,4 +112,67 @@ public Q_SLOTS:
...
+    Success,
+    WaitUserToActive,
+    FailToMint,
+    NotEnoughFund,
+    UserReject,
+    FailToUnlock,
+    Close
+};
+
+class AutoMintSparkModel : public QObject
```
levonpetrosyan93, line 133: Again, why we need separate types and classes, we can just reused what we already have,

PR #1267, opened 2023-07-17
Update linking tag

AaronFeickert:
Updates linking tags in preparation for a possible one-of-many proof upgrade. Makes corresponding changes to authorizing proofs.

Because this PR changes how linking tags and authorizing proofs are computed, it is a **breaking change** against [this commit](https://github.com/firoorg/firo/tree/89024560ba73b05796d08b930233e4011a727be5), which is apparently the current devnet and testnet commit.

### DIFF
--- a/src/libspark/chaum.cpp
+++ b/src/libspark/chaum.cpp
@@ -3,8 +3,8 @@
 
 namespace spark {
 
-Chaum::Chaum(const GroupElement& F_, const GroupElement& G_, const GroupElement& H_, const GroupElement& U_):
-    F(F_), G(G_), H(H_), U(U_) {
+Chaum::Chaum(const GroupElement& F_, const GroupElement& G_, const GroupElement& H_):
+    F(F_), G(G_), H(H_) {
 }
 
 Scalar Chaum::challenge(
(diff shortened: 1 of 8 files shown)


AaronFeickert:
This [preprint PR](https://github.com/firoorg/spark-paper/pull/19) contains updates reflecting this change.

levonpetrosyan93 (commented):
@AaronFeickert I think we need to remove U also from params.h/cpp as it  is no more used.

src/libspark/chaum.cpp
```diff
@@ -43,7 +42,7 @@ void Chaum::prove(
         throw std::invalid_argument("Bad Chaum statement!");
     }
     for (std::size_t i = 0; i < n; i++) {
-        if (!(F*x[i] + G*y[i] + H*z[i] == S[i] && T[i]*x[i] + G*y[i] == U)) {
+        if (!(F*x[i] + G*y[i] + H*z[i] == S[i] && (T[i]*x[i] + G*y[i]).isInfinity())) {
```
man2706kum, line 45: this condition forces if x[i] is zero then so is y[i] and vice versa. in that case the distribution of x and y might not be uniformly distributed.

AaronFeickert:
This proof change unfortunately is flawed, as setting `s = r = 0` would allow an adversary to reveal multiple tags against the same statement `S`. The original proof is fine, as `s != 0` is automatically enforced by the independence of the generators `U` and `G`, and which in turn ensures that only a single tag can be revealed for a given statement `S`.

Closing.

PR #1348, opened 2023-11-11
Refactor Public Transaction Creation

sproxet:
This PR refactors public transaction creation, which is a major change requiring careful review.


src/primitives/transaction.h
```diff
@@ -171,7 +171,7 @@ class CTxOut
 public:
     CAmount nValue;
     CScript scriptPubKey;
-    int nRounds;
+    int nRounds = -10;
```
psolstice, line 174: Why -10 specifically?

src/llmq/quorums_instantsend.h
```diff
@@ -113,6 +110,9 @@ class CInstantSendManager : public CRecoveredSigsListener
     std::atomic_bool isNewInstantSendEnabled{false};
 
 public:
+    CCriticalSection cs;
+    CInstantSendDb db;
+
```
psolstice, line 115: Is this the only way to implement the test? Wallet changes can certainly be done with an extra member function.

src/wallet/wallet.cpp
```diff
@@ -193,7 +405,7 @@ CPubKey CWallet::GetKeyFromKeypath(uint32_t nChange, uint32_t nChild, CKey& secr
         MnemonicContainer mContainer = mnemonicContainer;
         DecryptMnemonicContainer(mContainer);
         SecureVector seed = mContainer.GetSeed();
-        masterKey.SetMaster(&seed[0], seed.size());
+        masterKey.SetMaster(&seed.at(0), seed.size());
```
psolstice, line 408: Using std::vector::data() method is the only possible improvement here. Other methods work but at() is just as bad or even worse than []

src/wallet/wallet.cpp
```diff
@@ -84,6 +85,217 @@ CFeeRate CWallet::minTxFee = CFeeRate(DEFAULT_TRANSACTION_MINFEE);
...
+
+bool CTransparentTxout::IsLLMQInstantSendLocked() const {
+    if (_isMockup)
+        return _mockupIsLLMQInstantSendLocked;
+
+    assert(wallet);
+    AssertLockHeld(wallet->cs_wallet);
+    AssertLockHeld(llmq::quorumInstantSendManager->cs);
+
+    return llmq::quorumInstantSendManager->db.GetInstantSendLockByTxid(GetHash()) != nullptr;
```
psolstice, line 241: why not use quorumInstantSendManager->IsLocked() here?

src/wallet/wallet.cpp
```diff
@@ -4320,374 +4533,339 @@ bool CWallet::ConvertList(std::vector <CTxIn> vecTxIn, std::vector <CAmount> &ve
...
+// strFailReason to a human-readable description of the failure. coinControl may be set or unset; if it is unset an
+// arbitrary transaction selection algorithm will be used. If coinControl->nMaxSize is set, transaction creation will
+// fail if the size of the transaction with MAXIMUM length signatures would exceed the transaction size. Actual
+// generated transactions may produce signatures below this size, which may result in a transaction that would be
+// smaller than coinControl->nMaxSize being rejected for size reasons; this is because at the time the transaction is
+// made, we do not know how long the signatures will be. If sign is false, the transaction will not be signed; if it is
+// true, the wallet must be unlocked. nExtraPayloadSize should be set to the number of extra bytes in the transaction
+// outside inputs/outputs (ie. LLMQ-related things). If fUseInstantSend is true, we will consider both locked and
+// confirmed UTXOs to be eligible for input; if it is not, only confirmed UTXOs will be used as inputs.
+bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletTx& wtxNew, CReserveKey* reservekey,
```
psolstice, line 4709: I don't think changing function signature (reserveKey parameter) is justified here, this function is inherited from bitcoin and subsequent merges with bitcoin PRs may create quite of lot of surprises

src/wallet/wallet.h
```diff
@@ -647,12 +648,66 @@ class LelantusJoinSplitBuilder;
...
  */
 class CWallet : public CCryptoKeyStore, public CValidationInterface
 {
+public:
+    /**
+     * Used to keep track of spent outpoints, and
+     * detect and report conflicts (double-spends or
+     * mutated transactions where the mutant gets mined).
+     */
+    typedef std::multimap<COutPoint, uint256> TxSpends;
```
psolstice, line 709: it's only a type but I can't find its usage of TxSpends outside of the CWallet class, why change it to public then?

src/wallet/wallet.h
```diff
@@ -665,6 +720,346 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+    void AddToSpends(const uint256 &wtxid);
+
+    CAmount GetFee(const CCoinControl* coinControl, size_t txSize);
+    std::vector<CTransparentTxout> GetTransparentTxouts() const;
+
+    // Get all the vRelevantTransactions that can be used. Inputs specifically selected in coinControl are put into
+    // vCoinControlInputs, and all other inputs compatible with coinControl and available for use are put into
+    // vAvailableInputs.
+    template<typename AbstractTxout>
+    void GetAvailableInputs(const std::vector<AbstractTxout>& vRelevantTransactions,
```
psolstice, line 734: This is a huge function to leave it in the header file. Please consider moving it to cpp file and then explicitly instantiate it with all the possible AbstractTxout types

src/wallet/wallet.h
```diff
@@ -665,6 +720,346 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+    // depending on the value of fSubtractFeeFromAmount); nConstantSize must be the size of the transaction minus the
+    // inputs, input count, and P2SH_OUTPUT_SIZE in the event that input and output amounts aren't exactly equal. If
+    // fUseInstantSend is not set, we will ignore UTXOs without a confirmation; if it is set, we will ignore only
+    // un-is-locked UTXOs. Fees will be automatically determined if not set in coinControl. If fAllowPartial is set, we
+    // will collect as many inputs as we can and return even if we don't have enough to fulfill nRequired. We return
+    // true if we were able to collect enough inputs to fulfill nRequired, and false otherwise. If fAllowPartial is set,
+    // vInputs may be returned with a 0-length; this will occur if there vRelevantTransactions is empty or if no
+    // transactions could cover the required fees.
+    template<typename AbstractTxout>
+    bool GetInputsForTx(const std::vector<AbstractTxout>& vRelevantTransactions, std::vector<AbstractTxout>& vInputs,
```
psolstice, line 810: same as above

src/wallet/coincontrol.h
```diff
@@ -16,13 +18,16 @@ enum class CoinType
...
-    WITH_MINTS = 8
+    WITH_MINTS = 8,
+    WITH_1000 = 9
 };
 
 /** Coin Control Features. */
 class CCoinControl
 {
 public:
+    std::set<COutPoint> setSelected;
```
psolstice, line 29: You know I don't like these changes of visibility :)

psolstice (changes requested):

reubenyap:
@coderabbitai review

aleflm (approved):
After reviewing the code and running the tests, this look good to me. Thanks.

src/wallet/txbuilder.cpp
```diff
@@ -19,27 +19,28 @@
...
 {
 }
 
-TxBuilder::TxBuilder(CWallet& wallet) noexcept : wallet(wallet)
+SigmaTxBuilderSuperclass::SigmaTxBuilderSuperclass(CWallet& wallet) noexcept : wallet(wallet)
 {
 }
 
-TxBuilder::~TxBuilder()
+SigmaTxBuilderSuperclass::~SigmaTxBuilderSuperclass()
```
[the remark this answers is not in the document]
aleflm: I would remove the word `Super` from this name too. @psolstice do you prefere to keep it with `Super` or do you agree we should remove the word `Super` from this name.

src/validation.cpp
```diff
@@ -1196,6 +1196,7 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
                     if (pfMissingInputs) {
                         *pfMissingInputs = true;
                     }
+                    LogPrintf("%s(): Couldn't find input %s-%d for tx %s\n", __func__, txin.prevout.hash.GetHex(), txin.prevout.n, tx.GetHash().GetHex());
```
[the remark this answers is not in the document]
aleflm: I think the logging here is clear enough. Not very necessary to have extra comment.

src/script/script.cpp
```diff
@@ -217,6 +217,16 @@ unsigned int CScript::GetSigOpCount(const CScript& scriptSig) const
...
 
+bool CScript::IsPayToPublicKey() const {
+    if (size() != 35) return false;
+
+    opcodetype opcode;
+    const_iterator pc = begin();
+    GetOp(pc, opcode);
+    GetOp(pc, opcode);
+    return opcode == OP_CHECKSIG;
+}
```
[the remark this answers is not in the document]
aleflm: This is a valid suggestion:
```
+   // Expect the script to start with a push operation of the public key (33 bytes) followed by OP_CHECKSIG
+   if (!GetOp(pc, opcode) || opcode != OP_PUSHDATA1) return false;
```

Second suggestion about optimization is not that valid. Any compiler will optimize this.

@psolstice I can change the code an submit or we can keep what we have. . But depends on your final decision.

src/wallet/wallet.h
```diff
@@ -647,12 +648,66 @@ class LelantusJoinSplitBuilder;
...
+enum AbstractTxoutType {
+    Transparent
+};
+
+class CTransparentTxout {
+public:
+    static bool IsTransparentTxout(const CTxOut& txout);
+
+    CTransparentTxout() = default;
+    CTransparentTxout(COutPoint outpoint, CTxOut txout): outpoint(outpoint), txout(txout), _isMockup(true) {}
```
aleflm, line 661: Although we might not apply this comment right now. But it worth keep this in mind. 

I personally don't like to have test variables in production code. IMHO we can use something like test macro to eliminate test code from production code. 

For example this came up in my search: 

https://github.com/apache/incubator-resilientdb/blob/0541fe43bd1bbd69ecac5a1bc0d0cdd5ac04bc56/platform/statistic/stats.cpp#L44-L46

```
#ifdef TEST_MODE
  monitor_sleep_time_ = 1;
#endif

```

aleflm (dismissed):
I went through the code one more time. I didn't see any deal breaker. Just added a few comments here and there to make it more readable.

justanwar:
When trying to transparent send with 28 inputs, error message `Some coin control inputs could not be selected` appeared (if coin control was used to select inputs) or `Insufficient funds` if coin control was not used to select inputs, then the error `Transaction creation failed` appared.`

The inputs have a minimum of 5 confirmations. With 6 confirmations, no errors were observed.

PR #1374, opened 2023-12-05
Add size and type checks to coin deserialization

### LINKED ISSUE #1373: Spark coin deserialization should enforce fixed sizes
Spark coins contain encrypted data whose size is fixed by the coin type:
- mint coin encrypted data should occupy 84 bytes: 16-byte encrypted diversifier, 32-byte encoded scalar, 32-byte padded memo
- spend coin encrypted data should occupy 90 bytes: 8-byte encoded value, 16-byte encrypted diversifier, 32-byte encoded scalar, 32-byte padded memo

These sizes should be enforced.

AaronFeickert:
## PR intention
Adds size and type checks during coin deserialization that fail on known bad data.

Closes #1373.

## Code changes brief
Certain aspects of coin data must be checked during deserialization. These include the validity of the coin type (mint or spend) and the size of encrypted recipient data. This PR adds these checks directly into deserializers, and will throw errors on bad data.

### DIFF
--- a/src/libspark/aead.h
+++ b/src/libspark/aead.h
@@ -25,7 +25,7 @@ struct AEADEncryptedData {
 
 		// Key commitment must be the correct size, which also includes an encoded size
 		READWRITE(key_commitment);
-		if (key_commitment.size() != 1 + AEAD_COMMIT_SIZE) {
+		if (key_commitment.size() != AEAD_COMMIT_SIZE) {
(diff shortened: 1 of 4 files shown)


AaronFeickert:
This still fails non-`libspark` tests, where AEAD key commitments are 64 bytes instead of 32 bytes. Not sure why this occurs.

AaronFeickert:
Superseded by #1379, which fixes test failures.

PR #1381, opened 2023-12-18
Unpad memos

### LINKED ISSUE #1380: Remove memo padding during coin identification
Spark transactions include an encrypted memo. To reduce metadata, the memo is padded to a fixed length before encryption. This is somewhat subpar, as the padding is not removed.

A straightforward way to improve the user experience and avoid confusion would be to prepend the memo length before padding, and include this length in the encrypted data.

AaronFeickert:
When a recipient decrypts a memo during coin identification, the memo is still padded, which is poor user experience.

This PR includes the memo length as part of encrypted recipient data, and removes the padding during coin identification. This ensures that the original unpadded memo is provided to the recipient.

Closes #1380.

### DIFF
--- a/src/libspark/coin.cpp
+++ b/src/libspark/coin.cpp
@@ -52,6 +52,9 @@ Coin::Coin(
 	std::vector<unsigned char> padded_memo(memo_bytes);
 	padded_memo.resize(this->params->get_memo_bytes());
 
+	// Prepend the unpadded memo length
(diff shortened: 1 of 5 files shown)


AaronFeickert:
Marking as draft for a request to include the length encoding as part of the existing structure to avoid a change in overall encrypted data length.

levonpetrosyan93 (approved):
Looks good to me

psolstice (approved):

PR #1384, opened 2023-12-21
Segfault fixed

levonpetrosyan93:


src/spark/primitives.h
```diff
@@ -49,6 +49,7 @@ struct CSparkMintMeta
         READWRITE(memo);
         READWRITE(serial_context);
         READWRITE(type);
+        coin.setParams(spark::Params::get_default());
```
psolstice, line 52: I think this should be done only when __reading__ coin from the stream, writing shouldn't modify anything

psolstice (changes requested):

src/spark/primitives.h
levonpetrosyan93: okay, agree, will change

psolstice (approved):

justanwar (approved):

PR #1395, opened 2024-01-16
WIP: spend transaction claims

AaronFeickert:
This PR is an initial draft of a design for spend transaction claims.

A claim is a proof against an existing spend transaction that asserts the prover knew the spend key used to authorize the transaction. It also binds an arbitrary message into the proof, which can be useful to avoid replay.

How this should be structured in the codebase is up for debate. Because a claim is structured like a Chaum authorizing proof, it uses the same `ChaumProof` data structure and serialization. It is generated and verified using the static `SpendTransaction::proveClaim` and `SpendTransaction::verifyClaim` functions.

The prover must provide the spend transaction, the (secret) input coin data representing the coins that were consumed in the spend transaction, its full view and spend keys, an arbitrary message, and an identifier that the verifier can use to obtain its own view of the spend transaction.

The verifier must provide the spend transaction, arbitrary message, and identifier. It is _very important_ that the verifier use its own view of the spend transaction! Otherwise, the prover could lie about its contents.

Once the prover produces the `ChaumProof` data structure representing the claim, it should be sent to the verifier in a serialized package containing:
- the `ChaumProof`
- the spend transaction identifier
- the arbitrary message

The verifier then uses its view of the ledger to look up the spend transaction using the identifier. It verifies the claim by checking the `ChaumProof` while binding in the identifier and message.

No tests exist for this yet, in order to first get feedback about the design.


AaronFeickert:
Note that this also includes the change proposed in #1396.

levonpetrosyan93 (commented):
Looks good to me,
Please add unittests too.

AaronFeickert:
Added basic tests.

reubenyap:
@coderabbitai review

PR #1396, opened 2024-01-17
Update binding hash function signature

AaronFeickert:
## PR intention
Updates `SpendTransaction::hash_bind_inner` to make it static and take `S1` and `T` as arguments.

## Code changes brief
Previously, the function in question was not static, and inferred `S1` and `T` from the object. This was inconsistent and had a bad code smell.

This PR makes the function static and is explicit about all fields used in the hash.


psolstice (approved):

AaronFeickert:
This shouldn't affect verification, but should be tested with a reindex to confirm.

levonpetrosyan93 (approved):

PR #1410, opened 2024-02-07
Fix sync issue on mac

levonpetrosyan93:
The PR fixes the bug resulting sync issues on some platforms, due to unordered_map being serialized and hashed.
Used map instead of unordered_map.


psolstice (approved):

AaronFeickert (approved):
utACK

PR #1427, opened 2024-04-07
checkifmncollateral rpc added

levonpetrosyan93:


src/rpc/misc.cpp
```diff
@@ -1430,6 +1431,47 @@ UniValue getsparklatestcoinid(const JSONRPCRequest& request)
...
+    } catch (std::logic_error const & e) {
+        throw std::runtime_error(std::string("An exception occurred while parsing parameters: ") + e.what());
+    }
+
+    uint256 txid = uint256S(strTxId);
+
+    CTransactionRef tx;
+    uint256 hashBlock;
+    if(!GetTransaction(txid, tx, Params().GetConsensus(), hashBlock, true))
+        throw std::runtime_error("Unable to get the transction");
```
psolstice, line 1462: typo in "transaction" plus wouldn't it be better to say something like "Unknown transaction"?

src/rpc/misc.cpp
```diff
@@ -1430,6 +1431,47 @@ UniValue getsparklatestcoinid(const JSONRPCRequest& request)
...
+
+    auto mnList = deterministicMNManager->GetListAtChainTip();
+    COutPoint o(txid, index);
+
+    bool mnexists = false;
```
psolstice, line 1467:
```
bool fMnExists = deterministicMNManager->IsProTxWithCollateral(tx, index) || mnList.HasMNByCollateral(o);
```
if we need to fix the message, let's fix the beauty of the code too pls. :)

psolstice (changes requested):

psolstice (approved):

PR #1433, opened 2024-04-15
Faster initial block index load for low memory machines

psolstice:
## Code changes brief
Linux only: detect if there is <2GB available RAM and then lower the number of block headers to perform a full check on during startup

## Summary by CodeRabbit

- **New Features**
	- Enhanced startup performance for systems with lower RAM by adjusting the number of blocks checked.


levonpetrosyan93 (approved):
LGTM

PR #1436, opened 2024-04-29
improve warning when sending transparent tx

### LINKED ISSUE #1403: Make this a private transaction warning
Currently when sending to transparent address it shows this warning:

image

"... If it is masternode transaction, you do not have to go private."

It may be good to include also info about exchanges as at least Binance requires to send from transparent address. And once exchanges will use exchange addresses it will have to be send from transparent for sure.

erciccione:
## PR intention
Improve warning message when sending a transparent transaction. Feel free to suggest better wording.

Resolves #1403

## Summary by CodeRabbit

- **New Features**
	- Updated user interface messages to emphasize privacy options for transactions, except when sending to exchanges or conducting masternode transactions.

### DIFF
--- a/src/qt/forms/sendcoinsentry.ui
+++ b/src/qt/forms/sendcoinsentry.ui
@@ -162,7 +162,7 @@
       <item>
         <widget class="QLabel" name="textWarning">
           <property name="text">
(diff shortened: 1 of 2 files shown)


erciccione:
Just noticed #1434, which does the same thing, but  goes deeper. Keeping this open as a fall-back option.

PR #1437, opened 2024-05-06
Qt scaling fix

firstcryptoman:
### New Features
- **Dynamic Resizing**: Implemented dynamic resizing logic in various components, ensuring UI elements adjust appropriately to different window sizes and display settings. This feature enhances usability across diverse hardware environments.

### Bug Fixes
- **Consistent UI Across Resolutions**: Fixed issues where UI components would not scale correctly on resizing, leading to inconsistent layouts and usability problems in different screen resolutions.

### Refactor
- **Code Cleanup for UI Scaling**: Refactored `resizeEvent` handlers across multiple UI components to utilize dynamic sizing calculations instead of fixed pixel values. This change simplifies future UI adjustments and enhances code maintainability.

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -484,6 +484,7 @@ void OverviewPage::resizeEvent(QResizeEvent* event)
     const int newWidth = event->size().width();
     const int newHeight = event->size().height();
     adjustTextSize(newWidth, newHeight);
+
     // Determine widths for specific widgets as percentages of total width
     int labelWidth = static_cast<int>(newWidth * 0.5);
     int labelMinWidth = static_cast<int>(newWidth * 0.15);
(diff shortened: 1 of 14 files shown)


src/qt/overviewpage.cpp
```diff
@@ -398,8 +403,25 @@ void OverviewPage::onRefreshClicked()
...
+
+    if(status == WalletModel::Locked) {
+
+        AskPassphraseDialog dlg(AskPassphraseDialog::Unlock, this, info);
+        dlg.setModel(walletModel);
+        dlg.exec();
+        ui->textWarning2->setText(tr("To complete the migration of %1").arg(migrateAmount));
+        ui->textWarning1->hide();
+        ui->warningFrame->show();
+    } else {
```
levonpetrosyan93, line 421: Your logic will open the wallet unlock window, but after unlock you will need to click again to do the migration, I think we need to do the following logic,
- if wallet is locked
    Show the window and ask to unlock
- if wallet is not locked or it is unlocked by upper step, so the migration

levonpetrosyan93 (changes requested):

levonpetrosyan93 (approved):

PR #1438, opened 2024-05-07
Transaction large error message change

levoncrypto:


src/qt/bitcoinstrings.cpp
```diff
@@ -553,7 +553,7 @@ QT_TRANSLATE_NOOP("firo-core", "Transaction has too long of a mempool chain"),
 QT_TRANSLATE_NOOP("firo-core", "Transaction must have at least one recipient"),
 QT_TRANSLATE_NOOP("firo-core", "Transaction not valid."),
 QT_TRANSLATE_NOOP("firo-core", "Transaction too large for fee policy"),
-QT_TRANSLATE_NOOP("firo-core", "Transaction too large"),
+QT_TRANSLATE_NOOP("firo-core", "This transaction exceeds the transaction size limit of 100 kB, please select less inputs or consolidate your UTXOs before trying to send again"),
```
erciccione, line 556: I suggest something like

```suggestion
QT_TRANSLATE_NOOP("firo-core", "transaction is too large (size limit: 100Kb). Select less inputs  or consolidate your UTXOs"),
```

to make things clearer for non-technical users

PR #1447, opened 2024-06-06
Gettotalsupply

levonpetrosyan93:


reubenyap:
Only the Zerocoin supply figure is confirmed while the Lelantus figures does not account for funds that we managed to blacklist during the attack.
Sigma pool is going to be closed and still has positive figure of 182,643 though it hasn't seen movement for a while.
Would need for Lelantus window to close to audit supply there and is likely smaller number.
Calculation working required for the CVE.

reubenyap:
Putting this on KIV until Lelantus window is closed.

PR #1448, opened 2024-06-07
Automate spend from spark to EX-address in Qt

psolstice:
## PR intention
When spending private funds to EX-address in Qt client automatically create an intermediate transaction.

## Code changes brief
Changed Qt code handling "Send" page to support this scenario.


levonpetrosyan93 (approved):
LGTM

PR #1449, opened 2024-06-09
New rpc calls for mobile

levonpetrosyan93:
Added new rpc calls,
- `getusedcoinstagstxhashes`, this rpc returns all spent coin serials attached to the hash of spending transaction, previously we had `getusedcoinstags` , the new call allows to get also in which transaction is spent the coin, to do this we add field inside block index, and in spark state, when tx is being checked during block acceptance, we add the pair of ltag and tx hash into block index, and each time we construct the state, we get the data from the index. this fields are constructed in case `-mobile` is passed,
- `getmempooltxids` gives the list of spark tx hashes inside mempool, this allow to get spark state not only for blockchian but also from mempool.
- `getmempooltxs` returns spark metadata for each tx hash, existing in mempool, in case there is no data for particular tx, that means tx is removed from mempool, confirmed or removed for another reason.

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -1488,11 +1488,11 @@ UniValue getsparklatestcoinid(const JSONRPCRequest& request)
     return UniValue(latestCoinId);
 }
 
-UniValue getmempooltxids(const JSONRPCRequest& request)
+UniValue getmempoolsparktxids(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 0)
         throw std::runtime_error(
-                "getmempooltxids\n"
+                "getmempoolsparktxids\n"
                 "\nReturns all mempool transaction ids.\n"
         );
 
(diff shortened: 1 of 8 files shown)


psolstice (changes requested):
Could you please add the description of what's the problem and how it's fixed. I see the coderabbit's comments, but this is not enough to do the review. Especially please explain the reasons behind adding new members of `CBlockIndex`.

levonpetrosyan93:
@psolstice the description added.

src/rpc/misc.cpp
```diff
@@ -1431,6 +1488,105 @@ UniValue getsparklatestcoinid(const JSONRPCRequest& request)
     return UniValue(latestCoinId);
 }
 
+UniValue getmempooltxids(const JSONRPCRequest& request)
```
psolstice, line 1491: This name must have "spark" in it. Otherwise, someone will find this RPC and complain that it's not working as he wants it to. As a matter of fact, all new RPCS should follow the same rule

src/rpc/misc.cpp
```diff
@@ -1431,6 +1488,105 @@ UniValue getsparklatestcoinid(const JSONRPCRequest& request)
...
+    for (auto it = txs.begin(); it != txs.end(); it++) {
+        if (!it->tx->IsSparkTransaction())
+            continue;
+        result.push_back(EncodeBase64(it->tx->GetHash().begin(), it->tx->GetHash().size()));
+    }
+
+    return result;
+}
+
+UniValue getmempooltxs(const JSONRPCRequest& request)
```
psolstice, line 1510: same, very misleading name

src/rpc/misc.cpp
levonpetrosyan93: agree, updated

src/rpc/misc.cpp
levonpetrosyan93: updated

psolstice (dismissed):

PR #1463, opened 2024-06-25
Update macOS Build Instructions for Dependency and Configuration IssuesMac os build fix

firstcryptoman:
### Description
This pull request revises the macOS build instructions for the Firo wallet, addressing dependency recognition and configuration issues on macOS Sonoma 14.5 with an M chip Mac.

#### Key Changes Include:
1. **Boost Library Configuration**:
   - Adjusted the build instructions to ensure proper recognition of the Boost libraries, which were not initially recognized. This includes specifying the correct paths for the Boost headers and libraries in the configuration settings.

2. **Dependency Installation Adjustments**:
   - Updated the instructions to include `m4` and `python-setuptools` in the Homebrew installation commands, addressing issues where these tools were not recognized post-installation. This ensures a more streamlined setup and reduces configuration errors during the build process.

3. **Environment Setup**:
   - Outlined the necessary steps for setting up the development environment on macOS, with considerations for the specific challenges posed by the latest macOS versions and hardware configurations.

#### Purpose:
The updates aim to streamline the setup process for developers on macOS, ensuring all key dependencies like Boost, `m4`, and `python-setuptools` are installed and recognized correctly, enhancing the clarity and accuracy of the build instructions.

#### Testing:
These revised instructions have been tested on macOS Sonoma 14.5 with an M chip Mac.

### DIFF
--- a/doc/build-macos.md
+++ b/doc/build-macos.md
@@ -4,14 +4,14 @@ The commands in this guide should be executed in a Terminal application.
 The built-in one is located in `/Applications/Utilities/Terminal.app`.
 
 ## Preparation
-1. **Install macOS Command Line Tools** (if not already installed):
+1. Install macOS Command Line Tools (if not already installed):
    ```bash
    xcode-select --install
    ```
    When the popup appears, click `Install`.
 
 
-2. **Install Homebrew** (if not already installed):
+2. Install Homebrew (if not already installed):
    ```bash
    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
    ```
@@ -101,7 +101,10 @@ Download and install the community edition of [Qt Creator](https://www.qt.io/dow
 Uncheck everything except Qt Creator during the installation process.
 
 1. Make sure you installed everything through Homebrew mentioned above
-2. Do a proper  ``` ./configure --prefix=`pwd`/depends/`depends/config.guess` --enable-debug   ```
+2. Properly configure the build environment:
+   ```bash
+   ./configure --prefix=`pwd`/depends/`depends/config.guess` --enable-debug
+   ```
 3. In Qt Creator do "New Project" -> Import Project -> Import Existing Project
 4. Enter "bitcoin-qt" as project name, enter `src/qt` as location
 5. Leave the file selection as it is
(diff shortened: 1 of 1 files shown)


levonpetrosyan93 (approved):

doc/build-macos.md
```diff
@@ -3,24 +3,29 @@ macOS Build Instructions and Notes
...
-----------------------
-
-    brew install automake berkeley-db4 libtool boost miniupnpc openssl pkg-config protobuf python qt libevent qrencode
 
 In case you want to build the disk image with `make deploy` (.dmg / optional), you need RSVG
 
-      brew install librsvg
+    brew install librsvg
       
 Berkley DB
```
AaronFeickert, line 30: This renders as a second-level heading that also contains the Firo Core build instructions, which is probably not what was intended. It might be more clear to have it under the dependency header, and then have the build instructions as a higher-level heading.

AaronFeickert (changes requested):
This fails on an M1 MacBook Air. The build reports no suitable `m4` installation, and prompts to have the build tools installed again. This will keep occurring.

From discussions elsewhere, it sounds like this may be related to whether or not you have Xcode installed (not just the `xcode-select`-installed tools), but I have not verified this.

doc/build-macos.md
firstcryptoman: That’s an interesting point. My tests were on Macs that all had Xcode installed, which likely explains why I didn’t encounter the same problem. Considering the size of the full Xcode, its installation is mostly unnecessary for users who are just trying to manually build the project. I’ll continue to look into this issue to find a workaround. If I can't find a better solution, I'll update our documentation to suggest installing Xcode. Also, I’ve taken your advice and reorganized the build instructions for clearer navigation. Thanks for your insightful comments!

AaronFeickert: Plus it's apparently necessary to obtain an Apple ID in order to even obtain Xcode, which not all users may wish to do, and which imposes another effective build requirement.

doc/build-macos.md
```diff
@@ -3,56 +3,91 @@ macOS Build Instructions and Notes
...
+
+In case you want to build the disk image with `make deploy` (.dmg / optional), you need RSVG:
+```bash
+brew install librsvg
+```
+### Ensure `m4` is Found
+After installing `m4`, ensure that it is correctly linked and available in your PATH:
+```bash
+brew link m4
+```
```
AaronFeickert, line 34: This returned a warning to pass `--force` to the command to properly set symlinks.

AaronFeickert (approved):
Tested builds with and without GUI, as well as the `deploy` target.

AaronFeickert (approved):
LGTM, except I haven't tested the Berkeley DB installation.

levonpetrosyan93 (approved):

PR #1423, opened 2024-04-01
Fix compiler warnings

aleflm:
Includes two commits:

- Update lib `immer` to a more recent version (v0.8.1 5875f7739a6c642ad58cbedadb509c86d4212).  This fixes a lot of random warnings we have in lib immer.
- Fixed warnings in Linux / Clang.
- Fixed warnings in Win cross-compile.
- no-gui build only.

## Summary by CodeRabbit

- **Refactor**
  - Improved code readability and performance in various library components through restructuring and optimizations.
  - Enhanced error handling and memory management across several modules.
- **New Features**
  - Introduced new member functions and improvements in mutable versions of `map`, `set`, and `table` classes.
- **Bug Fixes**
  - Adjusted file path handling logic to correctly handle absolute paths.
- **Chores**
  - Updated configuration with macros for C++17 features, exception handling, debugging, and compiler-specific optimizations.
- **Documentation**
  - Added necessary include directives across multiple files for better code management and clarity.

PR #1470, opened 2024-07-15
Fix warnings in build process

psolstice:
Supersedes #1423 

Update lib immer to a more recent version (v0.8.1 5875f7739a6c642ad58cbedadb509c86d4212). This fixes a lot of random warnings we have in lib immer.
Fixed warnings in Linux / Clang.
Fixed warnings in Win cross-compile.


levonpetrosyan93 (approved):
Looks good to me.

justanwar:
Except for `EXCEPTION_ACCESS_VIOLATION` when starting Qt wallet with `-disablewallet` flag (not that you should), everything looking good so far.

PR #1475, opened 2024-08-15
UI improvements

levonpetrosyan93:
- Stripped lelantus Ui
- improved performance on big wallets


src/qt/clientmodel.cpp
```diff
@@ -80,19 +83,28 @@ void ClientModel::setMasternodeList(const CDeterministicMNList& mnList)
...
+    }
     setMasternodeList(deterministicMNManager->GetListAtChainTip());
 }
 
 int ClientModel::getNumBlocks() const
 {
-    LOCK(cs_main);
+    TRY_LOCK(cs_main,lock);
+    if (!lock) {
+        return -1;
```
psolstice, line 106: in code sometimes this snippet is used
```
setNumBlocks(model->getNumBlocks(), ... )
```
I'm pretty sure it'll cause crashes if `getNumBlocks()` can return -1

src/qt/clientmodel.cpp
```diff
@@ -138,7 +156,7 @@ quint64 ClientModel::getTotalBytesSent() const
 
 QDateTime ClientModel::getLastBlockDate() const
 {
-    LOCK(cs_main);
+    TRY_LOCK(cs_main,lock);
```
psolstice, line 159: Just proceed if TRY_LOCK failed?

src/qt/clientmodel.cpp
```diff
@@ -161,7 +179,7 @@ double ClientModel::getVerificationProgress(const CBlockIndex *tipIn) const
     CBlockIndex *tip = const_cast<CBlockIndex *>(tipIn);
     if (!tip)
     {
-        LOCK(cs_main);
+        TRY_LOCK(cs_main,lock);
```
psolstice, line 182: Another one with not checking

src/qt/sparkmodel.cpp
```diff
@@ -37,7 +37,12 @@ CAmount SparkModel::getMintableSparkAmount()
 {
     std::vector<std::pair<CAmount, std::vector<COutput>>> valueAndUTXO;
     {
-        LOCK2(cs_main, wallet->cs_wallet);
+        TRY_LOCK(cs_main,lock_main);
+        if (!lock_main)
+            return 0;
```
psolstice, line 42: Can't it lead to the case when a number on overview page jump from 0 to the real value and back quickly?

src/qt/transactiontablemodel.cpp
```diff
@@ -131,7 +136,12 @@ class TransactionTablePriv
             }
             if(showTransaction)
             {
-                LOCK2(cs_main, wallet->cs_wallet);
+                TRY_LOCK(cs_main,lock_main);
+                if (!lock_main)
+                    return;
```
psolstice, line 141: This (and many other lines like this) may break transaction list completely. I'd suggest some other approach (caching list and atomically adding all the missing txs?)

psolstice (changes requested):

psolstice (approved):

PR #1493, opened 2024-11-03
Add memo field in UI

levoncrypto:


src/spark/sparkwallet.cpp
```diff
@@ -742,7 +742,8 @@ std::vector<CRecipient> CSparkWallet::CreateSparkMintRecipients(
         script.insert(script.end(), serializedCoins[i].begin(), serializedCoins[i].end());
         unsigned char network = spark::GetNetworkType();
         std::string addr = outputs[i].address.encode(network);
-        CRecipient recipient = {script, CAmount(outputs[i].v), false, addr};
+        std::string memo = outputs[i].memo;
+        CRecipient recipient = {script, CAmount(outputs[i].v), false, addr, memo};
```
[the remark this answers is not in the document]
levonpetrosyan93: I think this is a useful suggestion. Better to apply this, use get_memo_bytes() to get the max memo size.

src/spark/primitives.h
```diff
@@ -105,6 +107,7 @@ class CSparkOutputTx
     inline void SerializationOp(Stream& s, Operation ser_action) {
         READWRITE(address);
         READWRITE(amount);
+        READWRITE(memo);
```
levonpetrosyan93, line 110: you need a check here, to prevent issues in case you have an old type data without this field .

src/qt/sendcoinsentry.cpp
```diff
@@ -55,13 +56,36 @@ SendCoinsEntry::SendCoinsEntry(const PlatformStyle *_platformStyle, QWidget *par
...
 }
 
 SendCoinsEntry::~SendCoinsEntry()
 {
     delete ui;
 }
 
+void SendCoinsEntry::on_MemoTextChanged(const QString &text)
+{
+    int maxLength = 256;
```
levonpetrosyan93, line 73: Hard coding numbers like this is wrong coding style, also this is not right number. use get_memo_bytes() function to get memo size limit.

levonpetrosyan93 (changes requested):

psolstice (approved):

levonpetrosyan93 (approved):

PR #1496, opened 2024-11-17
Implemented an rpc to get the number of addresses with positive balance

levonpetrosyan93:
Iterates through the whale address index to calculate the total number of addresses with a positive balance.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -322,12 +322,9 @@ size_t CBlockTreeDB::findAddressNumWBalance() {
         std::pair<char,CAddressIndexKey> key;
         if (pcursor->GetKey(key) && key.first == DB_ADDRESSINDEX && (key.second.type == AddressType::payToPubKeyHash || key.second.type == AddressType::payToExchangeAddress)) {
             CAmount nValue;
-            if (pcursor->GetValue(nValue)) {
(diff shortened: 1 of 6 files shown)


src/txdb.cpp
```diff
@@ -313,6 +313,31 @@ bool CBlockTreeDB::ReadAddressIndex(uint160 addressHash, AddressType type,
...
+        boost::this_thread::interruption_point();
+        std::pair<char,CAddressIndexKey> key;
+        if (pcursor->GetKey(key) && key.first == DB_ADDRESSINDEX && (key.second.type == AddressType::payToPubKeyHash || key.second.type == AddressType::payToExchangeAddress)) {
+            CAmount nValue;
+            if (pcursor->GetValue(nValue)) {
+                auto it = addrMap.find(key.second.hashBytes);
+                if (it != addrMap.end()) {
+                    it->second += nValue;
+                    if (it->second <= 0)
+                        addrMap.erase(it);
```
psolstice, line 330: If it's less that a zero due to the previous vulnerability the result would be incorrect because we erased non-zero amount from the map.

psolstice (changes requested):

psolstice (approved):

PR #1505, opened 2024-12-11
Changes to compile with C++20

gevorgvoskanyan:
## PR intention
The codebase now compiles with C++20

## Code changes brief
Requires extensive testing before it can ever end up in master, which will be taken care of in scope of the spats project testing itself.


gevorgvoskanyan:
@coderabbitai review

firstcryptoman (approved):

PR #1508, opened 2024-12-16
Fix wallet scanning for newly created wallet

levoncrypto:


src/wallet/wallet.cpp
```diff
@@ -2405,6 +2405,10 @@ CBlockIndex* CWallet::GetBlockByDate(CBlockIndex* pindexStart, const std::string
 CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex *pindexStart, bool fUpdate, bool fRecoverMnemonic)
 {
     CBlockIndex* ret = nullptr;
+    if (GetBoolArg("-newwallet", false)) {
+        LogPrintf("Created new wallet, not need to scan\n");
```
gevorgvoskanyan, line 2409: should either be "no need" or "not needed"

src/wallet/wallet.cpp
```diff
@@ -7327,10 +7331,10 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
             }
         }
 
-        uiInterface.InitMessage(_("Rescanning..."));
+        if (!(GetBoolArg("-newwallet", false))) {uiInterface.InitMessage(_("Rescanning..."));}
         nStart = GetTimeMillis();
         walletInstance->ScanForWalletTransactions(pindexRescan, true, fRecoverMnemonic);
-        LogPrintf(" rescan      %15dms\n", GetTimeMillis() - nStart);
+        if (!(GetBoolArg("-newwallet", false))) {LogPrintf(" rescan      %15dms\n", GetTimeMillis() - nStart);}
```
gevorgvoskanyan, line 7337: What's the reason for making the messages conditional but not the actual operation itself?

levoncrypto: The condition to prevent scanning if a new wallet has been created is placed inside the ScanForWalletTransactions function. This approach makes it easy to change the algorithm in the future without having to edit each function call. It also helps avoid code duplication and simplifies testing since all the logic is in one place.

gevorgvoskanyan: ok

gevorgvoskanyan (approved):

levonpetrosyan93 (approved):

PR #1532, opened 2025-03-03
Spark names

psolstice:
## PR intention
Spark name is a way to create an easily remembered alias to a spark address. This PR adds both core and UI functionality for spark names creation and management. Fee is required to register a spark name, and it depends on spark name length (the shorter the length, the more expensive spark name registration is).

## Code changes brief
Core functionality is added on top of spark spend transaction. When spark is spent to development fund address spark name metadata is expected to follow.


levonpetrosyan93 (dismissed):
The only comment, if no objections please move sparkname.h/.cpp into spark directory

levonpetrosyan93 (approved):

PR #1553, opened 2025-04-17
Build fix

levonpetrosyan93:


psolstice (changes requested):
Why `getownsparknames` is needed? There is already an argument for `getsparknames` doing that (a single argument set to true retrieves only own spark names)

psolstice (approved):

levoncrypto (approved):

PR #1613, opened 2025-05-16
Spats central logic

gevorgvoskanyan:
## PR intention
Biggest chunk of the spats feature implementation.

## Code changes brief
Mostly working overall, only spends are missing at this point as the big missing thing.


src/spats/wallet.cpp
```diff
@@ -0,0 +1,531 @@
...
+   const auto scalar_of_proof = compute_mint_asset_supply_serialization_scalar( action_params, serialized.as_bytes_span() );
+   const spark::OwnershipProof proof = spark_wallet_.makeDefaultAddressOwnershipProof( scalar_of_proof );
+   LogPrintf( "Ownership proof for mint asset supply (hex): %s\n", proof );
+   CDataStream proof_serialized( SER_NETWORK, PROTOCOL_VERSION );
+   proof_serialized << proof;
+   script.insert( script.end(), serialized.begin(), serialized.end() );
+   assert( script.IsSpatsMint() );
+   script.insert( script.end(), proof_serialized.begin(), proof_serialized.end() );
+   assert( script.IsSpatsMint() );
+   auto tx = spark_wallet_.CreateSparkSpendTransaction( { CRecipient{ std::move( script ), {}, false, admin_public_address, "spats mint" } },
```
levonpetrosyan93, line 331: Why we need to put his into the transaction here? And whole upper logic also under question.

gevorgvoskanyan: This was written before the coin part was available. When the coin part arrived, I integrated it into the existing structure with minimal disturbance, so that is why it is like this. But yes, in hindsight this is a duplication of data that is not strictly necessary. The proof and all of the params could be taken from the coin, except it is missing the initiator_address, which we would need both for verifying the ownership proof on, and for the registry to be able to validate the action. I can add the initiator_address into the coin part, right before ownership proof, and then extract all the needed data from the coin part directly, and get rid of this top-level mint script. This will require quite a bit of refactoring, and also care needs to be taken to differentiate from 'create' action because that can contain this coin part too, but it is doable.
You also mentioned that there is a security issue here, could you please elaborate?

levonpetrosyan93: The security issue is the following,
I can change the mint output which is generated at the last step, and your tx verification will pass. 
By creating transaction you should ensure that tx components are safe and not possible to modify, this is done by putting proofs which can be created by  the owner, and the proof should be linked to the sensitive data.
Please just use my function and just add checks where TODOs are left.

gevorgvoskanyan: I see, thanks.
At least if the value was to be maliciously changed, https://github.com/firoorg/firo/blob/spats_central_logic/src/spark/state.cpp#L407 would detect and reject it, but the other attributes of the coin would need similar protection too, in order to be safe against all attacks of the kind that you mentioned. I will go ahead with your suggested route, and will also double-check other actions' code to make sure they don't suffer from similar vulnerabilities.

levonpetrosyan93: That like does not protect anything; I can change the whole TX output, not only the script.

src/utils/lock_proof.hpp
levonpetrosyan93, line 1: why do we need this?

src/utils/constrained_value.hpp
levonpetrosyan93, line 1: why and where  do we need this?

src/spats/wallet.cpp
gevorgvoskanyan: The refactoring is now done.

src/utils/lock_proof.hpp
gevorgvoskanyan: This is used in the Registry class to provide compile-time guarantees for functions that require certain kinds of locks to be taken before the function can be called, such as a read lock or a write lock. This ensures that bugs where a lock is forgotten to be taken don't slip into the final executable, but are caught as compilation errors instead.

src/utils/constrained_value.hpp
gevorgvoskanyan: This is a useful utility for creating types whose values are constrained, i.e. they must satisfy a certain criteria. For example, https://github.com/firoorg/firo/blob/spats_central_logic/src/spats/spark_asset.hpp#L40 shows its use for creating the type of an asset's symbol, which is required to be a non-empty string where all characters are uppercase letters. Users of this type can 100% rely on the value being valid, and don't have to check it everywhere themselves explicitly like they would when dealing with naked general types. Invalid objects just cannot exist, structurally.

src/spats/wallet.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+std::optional< CWalletTx > Wallet::create_burn_asset_supply_transaction( asset_type_t const asset_type,
+                                                                         const asset_symbol_t &asset_symbol,
+                                                                         supply_amount_t const burn_amount,
+                                                                         CAmount &standard_fee,
+                                                                         const BurnActionUserConfirmationCallback &user_confirmation_callback ) const
+{
+   const auto &admin_public_address = my_public_address_as_admin();
+
+   if ( asset_type == base::asset_type ) {
```
levonpetrosyan93, line 363: Please remove this; base asset coins are not supposed to be burned. In some extraordinary cases, people can just use regular spend.

src/spats/wallet.cpp
```diff
@@ -0,0 +1,541 @@
...
+            return {};
+         }
+      }
+
+      return tx;
+   }
+
+   const BurnParameters action_params( asset_type, burn_amount, admin_public_address, asset_symbol );
+   CScript script;
+   script << OP_SPATSBURN;
```
levonpetrosyan93, line 389: Here is a logical error,  asset coin can be burnt by the owner of the coin, not by the manager, so please simplify this and remove all that logic , evern extra opcode from script.h,  assed burn can have only one tx output, wbich is already done in my function, just use it as I have done in rpcwallet.cpp

levonpetrosyan93 (changes requested):

src/spats/wallet.cpp
levonpetrosyan93: Just use regular spend crearion function from your UI and please remove this function all together.

src/spats/wallet.cpp
gevorgvoskanyan: Ok.

src/spats/wallet.cpp
gevorgvoskanyan: > Here is a logical error, asset coin can be burnt by the owner of the coin, not by the manager, so please simplify this and remove all that logic , evern extra opcode from script.h, assed burn can have only one tx output, wbich is already done in my function, just use it as I have done in rpcwallet.cpp

Thanks for noticing this. It's not a logical error however, just a naming issue. The field in the action is the initiator's address, and everyone can be the initiator for burn, not just the admin. I thought to put the burn initiators into public record, such that not only we know how much has been burned but also by whom, thinking it might be useful some day. If you think it's not necessary or desirable, I can remove it, by refactoring the code to contain one tx output only. I would like to do the refactoring after the chaum bugfix though, so I would be able to test it properly.

levonpetrosyan93: Sorry Gevorg, I dont mean it is not needed or desirable, bur there is a logical error.
Putting admin address and ownership proof can be done only by admin, but burn can be done by anyone having the coin.
For tracking the burrnt amount and asset supply there is already all necesaary data, such as identifier(inside acript) and vout in tx out.

gevorgvoskanyan: > Just use regular spend crearion function from your UI and please remove this function all together.

This function is useful to keep for handling the user confirmation business. After the refactoring it will become much simpler though, than it is now.
Update: over the phone discussion the possibility of detaching the user confirmation from here was floated, but that's a bigger undertaking we can visit later on.

gevorgvoskanyan: > Sorry Gevorg, I dont mean it is not needed or desirable, bur there is a logical error. Putting admin address and ownership proof can be done only by admin, but burn can be done by anyone having the coin. For tracking the burrnt amount and asset supply there is already all necesaary data, such as identifier(inside acript) and vout in tx out.

I was putting the initiator's address and its ownership proof, not the admin's. It was just a wrong variable name in this function, and anybody was able to burn. But the refactoring will get rid of this anyway, and over the phone we agreed that there is no need to track the burn initiators, so the refactored code no longer will.

gevorgvoskanyan: Hi @levonpetrosyan93, I refactored spats burn along these lines and pushed, but the chaum verify is still failing. Please take a look. Thanks.

levonpetrosyan93: when you removed unnecessary data, it fixed one bug which was deviating tx hash for signing, but there was one more on my side, pushed the fix.

levonpetrosyan93: @gevorgvoskanyan jan even in that case that is a logical error, coin owner's address is not directly connected to coin publicly, so ownership proof provided here does not proof anything, tx creator can put there any address and nothing will change.

gevorgvoskanyan: @levonpetrosyan93 jan got it, I see that it had a potential to be misused. At any rate, it is gone now after the refactoring, but thanks for your insight anyway!

levonpetrosyan93: sure, thanks, I  was just trying to make sure we both  enough understanding about  the  protocol and business logic, and necessity of proofs we are including into tx,

PR #1615, opened 2025-05-18
Fix a crash on startup when ulimits are set to unlimited.

cassandras-lies:
## PR intention
Fix a crash on startup.

## Code changes brief
If ulimit is set to unlimited, getrlimit will return a value of UINT_MAX, but it was being interpreted as an int, so this was evaluated as -1 and led to a crash on startup thinking resource limits had been hit.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -62,6 +62,7 @@
 #include "llmq/quorums_init.h"
 
 #include <stdint.h>
+#include <sys/resource.h>
 #include <stdio.h>
 #include <memory>
 
@@ -1070,9 +1071,9 @@ void InitLogging() {
 namespace { // Variables internal to initialization process only
 
 ServiceFlags nRelevantServices = NODE_NETWORK;
-int nMaxConnections;
-int nUserMaxConnections;
-int nFD;
+rlim_t nMaxConnections;
+rlim_t nUserMaxConnections;
+rlim_t nFD;
 ServiceFlags nLocalServices = NODE_NETWORK;
 
 }
@@ -1168,10 +1169,10 @@ bool AppInitParameterInteraction()
                 (mapMultiArgs.count("-bind") ? mapMultiArgs.at("-bind").size() : 0) +
                 (mapMultiArgs.count("-whitebind") ? mapMultiArgs.at("-whitebind").size() : 0), size_t(1));
     nUserMaxConnections = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
-    nMaxConnections = std::max(nUserMaxConnections, 0);
+    nMaxConnections = std::max(nUserMaxConnections, (rlim_t)0);
 
     // Trim requested connection counts, to fit into system limitations
-    nMaxConnections = std::max(std::min(nMaxConnections, (int)(FD_SETSIZE - nBind - MIN_CORE_FILEDESCRIPTORS - MAX_ADDNODE_CONNECTIONS)), 0);
+    nMaxConnections = std::max(std::min(nMaxConnections, (rlim_t)(FD_SETSIZE - nBind - MIN_CORE_FILEDESCRIPTORS - MAX_ADDNODE_CONNECTIONS)), (rlim_t)0);
     nFD = RaiseFileDescriptorLimit(nMaxConnections + MIN_CORE_FILEDESCRIPTORS + MAX_ADDNODE_CONNECTIONS);
     if (nFD < MIN_CORE_FILEDESCRIPTORS)
         return InitError(_("Not enough file descriptors available."));


reubenyap:
bugbot run

reubenyap:
@cursor review

src/init.cpp
```diff
@@ -116,7 +116,9 @@ static CDSNotificationInterface* pdsNotificationInterface = NULL;
 // accessing block files don't count towards the fd_set size limit
 // anyway.
 #define MIN_CORE_FILEDESCRIPTORS 0
+#define rlim_t unsigned long long
```
reubenyap, line 119: **[HIGH] `#define rlim_t unsigned long long` is a dangerous macro that redefines a POSIX type**

Using `#define` to create a type alias is fragile. Unlike `typedef`, a `#define` doesn't respect scope, can clash with any header that defines or uses `rlim_t`, and expands textually (e.g. `(rlim_t)x` becomes `(unsigned long long)x` — works by luck but is not a proper cast expression in all contexts). On non-Windows, `<sys/resource.h>` already provides `rlim_t`.

**Suggested fix:**
Use `typedef` instead:
```cpp
#ifdef WIN32
#define MIN_CORE_FILEDESCRIPTORS 0
typedef unsigned long long rlim_t;
#else
#include <sys/resource.h>
#define MIN_CORE_FILEDESCRIPTORS 150
#endif
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, around line 119, `#define rlim_t unsigned long long` is used inside the `#ifdef WIN32` block. Replace it with `typedef unsigned long long rlim_t;`. Keep the `#define MIN_CORE_FILEDESCRIPTORS 0` on the line before it. Keep the `#else` branch with `#include <sys/resource.h>` unchanged."

</details>

src/init.cpp
```diff
@@ -1070,9 +1072,9 @@ void InitLogging() {
 namespace { // Variables internal to initialization process only
 
 ServiceFlags nRelevantServices = NODE_NETWORK;
-int nMaxConnections;
-int nUserMaxConnections;
-int nFD;
+rlim_t nMaxConnections;
```
reubenyap, line 1075: **[HIGH] `nMaxConnections` is `rlim_t` but downstream consumers expect `int`**

Changing these variables to `rlim_t` (unsigned long long) introduces implicit narrowing conversions at several downstream sites:
- Line 2079: `connOptions.nMaxConnections = nMaxConnections;` — `CConnman::Options::nMaxConnections` is `int` (net.h:165)
- Line 2080: `connOptions.nMaxOutbound = std::min(MAX_OUTBOUND_CONNECTIONS, connOptions.nMaxConnections);`
- Line 1177: `RaiseFileDescriptorLimit(...)` takes `int` and returns `int` (util.h:121)
- Lines 1183/1462: `%d`/`%i` format specifiers used with `rlim_t` values — this is undefined behavior

The real fix should keep `nMaxConnections`/`nFD` as `int` but fix the narrowing at the `getrlimit` call site in `RaiseFileDescriptorLimit()`. Or if `rlim_t` is kept here, all downstream consumers must also be updated.

**Suggested fix:**
The safer approach: keep these as `int`, and fix `RaiseFileDescriptorLimit` to internally handle `rlim_t` from `getrlimit` and clamp to `INT_MAX` before returning. This avoids cascading type changes throughout the codebase.

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, lines 1075-1077 change `nMaxConnections`, `nUserMaxConnections`, and `nFD` from `int` to `rlim_t`. This causes implicit narrowing conversions downstream: `CConnman::Options::nMaxConnections` is `int` (src/net.h:165), `RaiseFileDescriptorLimit` takes and returns `int` (src/util.h:121), and format specifiers `%d`/`%i` at lines 1183 and 1462 expect `int` not `unsigned long long`. Either revert these to `int` and fix the root cause in `RaiseFileDescriptorLimit()` in `src/util.cpp` (line 771-788) to internally use `rlim_t` from `getrlimit` but clamp the return value to `INT_MAX`, or update all downstream consumers to accept `rlim_t`. The former approach is safer and more localized."

</details>

src/init.cpp
```diff
@@ -1168,10 +1170,10 @@ bool AppInitParameterInteraction()
                 (mapMultiArgs.count("-bind") ? mapMultiArgs.at("-bind").size() : 0) +
                 (mapMultiArgs.count("-whitebind") ? mapMultiArgs.at("-whitebind").size() : 0), size_t(1));
     nUserMaxConnections = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
```
reubenyap, line 1172: **[MEDIUM] `GetArg` returns signed `int64_t`, stored in unsigned `rlim_t` — negative values wrap**

`GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS)` returns `int64_t`. If a user passes `-maxconnections=-1`, the signed value `-1` is stored into unsigned `rlim_t` and becomes `ULLONG_MAX` (~1.8×10¹⁹). The old code used `int`, so `std::max(nUserMaxConnections, 0)` on line 1173 would clamp it to 0. With `rlim_t`, `std::max(nUserMaxConnections, (rlim_t)0)` compares two unsigned values, and the wrapped value is enormous, so the clamp has no effect.

**Suggested fix:**
Clamp the signed value before converting to unsigned:
```cpp
int64_t rawMaxConn = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
nUserMaxConnections = (rawMaxConn < 0) ? 0 : static_cast<rlim_t>(rawMaxConn);
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, line 1172, `nUserMaxConnections = GetArg(\"-maxconnections\", DEFAULT_MAX_PEER_CONNECTIONS);` stores a signed `int64_t` return value in unsigned `rlim_t`, causing negative values to wrap to huge unsigned numbers. Add an intermediate signed variable and clamp: `int64_t rawMaxConn = GetArg(\"-maxconnections\", DEFAULT_MAX_PEER_CONNECTIONS); nUserMaxConnections = (rawMaxConn < 0) ? 0 : static_cast<rlim_t>(rawMaxConn);`. Remove the now-redundant `std::max(nUserMaxConnections, (rlim_t)0)` on line 1173."

</details>

src/init.cpp
```diff
@@ -116,7 +116,9 @@ static CDSNotificationInterface* pdsNotificationInterface = NULL;
 // accessing block files don't count towards the fd_set size limit
 // anyway.
 #define MIN_CORE_FILEDESCRIPTORS 0
+#define rlim_t unsigned long long
 #else
+#include <sys/resource.h>
```
reubenyap, line 121: **[LOW] `<sys/resource.h>` include is buried inside a preprocessor block**

Placing `#include <sys/resource.h>` inside the `#else` (non-Windows) block at line 121 is unconventional. System includes are typically grouped at the top of the file for discoverability. While functionally correct, this makes it easy to miss during code review.

**Suggested fix:**
Move the include to the system includes section at the top of the file, wrapped in `#ifndef WIN32`:
```cpp
#ifndef WIN32
#include <sys/resource.h>
#endif
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, line 121, `#include <sys/resource.h>` is placed inside the `#ifdef WIN32 / #else` block around line 114-123. Move it to the system includes section near the top of the file (around lines 30-50), wrapped in `#ifndef WIN32 / #endif`. Remove it from its current location inside the preprocessor block. Keep the `#define MIN_CORE_FILEDESCRIPTORS 150` in the `#else` branch."

</details>

PR #1617, opened 2025-05-19
CMake: add daemon decl flag

aleflm:
## PR intention
Add support for one of the hidden flags: `-daemon`. 

By default is OFF, for enabling, you can pass: `-DHAVE_DECL_DAEMON=ON` to cmake.


cassandras-lies:
`daemon()` is in POSIX libc. This should be on by default on non-Windows platforms. (The normal way people would interact with firod is broken if it is off.)

PR #1642, opened 2025-06-22
Support Tor .onion address

levoncrypto:


src/netaddress.h
```diff
@@ -87,8 +91,23 @@ class CNetAddr
         template <typename Stream, typename Operation>
         inline void SerializationOp(Stream& s, Operation ser_action) {
             READWRITE(FLATDATA(ip));
-        }
 
+            if (ser_action.ForRead()) {
+                bool v3Flag;
+                READWRITE(v3Flag);
```
psolstice: I think it's not compatible with previously saved data, this read will throw an exception if there isn't more data to read. Where do we save IP addresses to (peers.dat, etc)?

psolstice (changes requested):

PR #1649, opened 2025-08-04
Add 'atomic' library to Boost configuration

aleflm:
## PR intention
Add atomic library to boost dependency.


reubenyap:
@aleflm Can you resolve the conflicts?

PR #1665, opened 2025-09-15
Check Spark spend ZK proofs on acceptance into the mempool and cache the result, so it's not re-checked later

psolstice:
## PR intention
When a lot of Spark spend transaction are mined the miner has a problem with performance. ZK proofs for a Spark spend are checked once on acceptance into the mempool, and then every time a block is updated and `getblocktemplate` is called every transaction is checked once again making the complexity of multiple checks O(N^2) where N is the number of Spark spend transaction in the block. This PR caches the result of the first check and makes the check multithreaded.

## Code changes brief
Use thread pool to defer the checks and cache the result.

### DIFF
--- a/src/spark/state.cpp
+++ b/src/spark/state.cpp
@@ -761,6 +761,10 @@ bool CheckSparkSpendTransaction(
                 }
                 else {
                     LOCK(cs_checkedSparkSpendTransactions);
+                    if (gCheckProofThreadPool.IsPoolShutdown())
+                        // if we are shutting down, don't start any new tasks
+                        return true;
+
                     // put the proof into the thread pool for verification
                     auto future = gCheckProofThreadPool.PostTask([spend, cover_sets]() {
                         try {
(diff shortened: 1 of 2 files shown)


src/spark/state.cpp
```diff
@@ -583,10 +637,6 @@ bool CheckSparkSpendTransaction(
 
     LogPrintf("CheckSparkSpendTransaction: tx metadata hash=%s\n", txHashForMetadata.ToString());
 
-    if (!fStatefulSigmaCheck) {
```
levonpetrosyan93, line 586: why do we need to move this check to lower lines?

levonpetrosyan93: okay I see your point, so you verify the proof even it is not stateful check in a separate thread,
so in case it is mempool verification it is stateful, checkblock and connectblock are running this function one with stateful check one without, so what is the purpose  makeing the code more complex by adding mmore threads and so on?

psolstice: yes, exactly

PR #1677, opened 2025-09-24
Sigma strip

levonpetrosyan93:


src/chain.h
```diff
@@ -243,7 +243,7 @@ class CBlockIndex
 
     //! Public coin values of mints in this block, ordered by serialized value of public coin
     //! Maps <denomination,id> to vector of public coins
-    std::map<std::pair<sigma::CoinDenomination, int>, std::vector<sigma::PublicCoin>> sigmaMintedPubCoins;
+    std::map<std::pair<std::uint8_t, int>, std::vector<sigma::PublicCoin>> sigmaMintedPubCoins;
```
psolstice: I think it'd be much safer to just leave `sigma::CoinDenomination` as is to be safe. It's not clear if `std::uint8_t` is always a good replacement

src/chain.h
```diff
@@ -259,7 +259,7 @@ class CBlockIndex
     std::unordered_map<GroupElement, std::pair<uint256, std::vector<unsigned char>>> sparkTxHashContext;
 
     //! Values of coin serials spent in this block
-    sigma::spend_info_container sigmaSpentSerials;
+    std::unordered_map<Scalar, sigma::CSpendCoinInfo> sigmaSpentSerials;
```
psolstice: same here, we could leave `sigma::CScalarHash` in the code. It's not immediately clear if this mismatch could create a problem, but the removal is simply not worth it

src/txdb.cpp
```diff
@@ -439,9 +439,6 @@ bool CBlockTreeDB::LoadBlockIndexGuts(boost::function<CBlockIndex*(const uint256
                     pindexNew->reserved[1] = diskindex.reserved[1];
                 }
 
-                pindexNew->sigmaMintedPubCoins   = diskindex.sigmaMintedPubCoins;
-                pindexNew->sigmaSpentSerials     = diskindex.sigmaSpentSerials;
-
```
psolstice, line 444: I wouldn't do it here. Might cause side effects.

PR #1685, opened 2025-09-29
getusedcoinstags rpc fixed

levonpetrosyan93:


src/spark/state.cpp
```diff
@@ -1101,6 +1105,14 @@ void CSparkState::AddLTagTxHash(const uint256& lTagHash, const uint256& txHash)
 
 void CSparkState::RemoveSpend(const GroupElement& lTag) {
     auto iter = usedLTags.find(lTag);
+    if (GetBoolArg("-mobile", false)) {
```
psolstice: It makes sense to move the whole mobile part behind `if (iter != usedLTags.end())` check to avoid long loop if there is no tag in our state

PR #1688, opened 2025-10-14
Fix Several Sources of Frequent Lock Contention

cassandras-lies:
## PR intention

Reduces lock contention in three areas that were causing unnecessary `cs_main` and `cs_vNodes` acquisitions multiple times per second.

### Issues fixed:

1. **`RemoveBannedNodeStates()` causes frequent `cs_main` contention**: This function acquires `LOCK2(cs_main, cs)` and was being called every ~100ms (every loop iteration of `WorkThreadMain`). Since it only cleans up local memory structures for banned nodes, acquiring `cs_main` ~10 times/second is unnecessary and creates significant contention with validation, block processing, and other `cs_main` consumers. **Fix**: Throttle to once every 30 seconds.

2. **Unnecessary full `vNodes` copy in `ThreadSocketHandler`**: Every ~50ms, the entire `vNodes` vector was copied (`std::vector<CNode*> vNodesCopy = vNodes`) even though typically 0 nodes need disconnecting per iteration. **Fix**: Iterate `vNodes` directly and collect only disconnected nodes (usually none), avoiding an O(n) heap allocation on every loop.

3. **100ms idle sleep too frequent**: The idle sleep in `WorkThreadMain` contributes to the overall contention frequency. **Fix**: Increase to 250ms — using 250ms rather than the originally proposed 1000ms since there is no wakeup mechanism for new signing work and longer delays could impact ChainLock/InstantSend latency (per reviewer feedback).

## Code changes brief

- `src/llmq/quorums_signing_shares.cpp`: Added 30-second throttle for `RemoveBannedNodeStates()` using `GetTimeMillis()` timestamp tracking. Changed idle sleep from 100ms to 250ms.
- `src/net.cpp`: Refactored `ThreadSocketHandler` disconnect handling to avoid copying the full `vNodes` vector. Instead, iterates directly and collects only disconnected nodes before processing them. Same locking scope and behavior preserved.

---

**Note**: This PR has been rebased onto the latest `master` branch. The idle sleep was adjusted from 1000ms to 250ms based on review feedback from @aleflm (no wakeup mechanism exists, so 1000ms could degrade ChainLock/InstantSend responsiveness). Rebased version is on branch `cursor/block-header-validation-issue-667d`.

<div><a href="https://cursor.com/agents/bc-375dc6b2-adbb-44eb-920f-f7c6a7b8263e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-375dc6b2-adbb-44eb-920f-f7c6a7b8263e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/llmq/quorums_signing_shares.cpp
+++ b/src/llmq/quorums_signing_shares.cpp
@@ -1338,6 +1338,7 @@ void CSigSharesManager::BanNode(NodeId nodeId)
 void CSigSharesManager::WorkThreadMain()
 {
     int64_t lastSendTime = 0;
+    int64_t lastRemoveBannedNodeStatesTime = 0;
 
     while (!workInterrupt) {
         if (!quorumSigningManager || !g_connman) {
@@ -1349,7 +1350,13 @@ void CSigSharesManager::WorkThreadMain()
 
         bool didWork = false;
 
-        RemoveBannedNodeStates();
+        // RemoveBannedNodeStates holds cs_main, and its functionality is only required for local memory management. It
+        // is much more performant to call this rarely.
+        if (GetTimeMillis() - lastRemoveBannedNodeStatesTime > 30000 /* 30s */) {
+            RemoveBannedNodeStates();
+            lastRemoveBannedNodeStatesTime = GetTimeMillis();
+        }
+
         didWork |= quorumSigningManager->ProcessPendingRecoveredSigs(*g_connman);
         didWork |= ProcessPendingSigShares(*g_connman);
         didWork |= SignPendingSigShares();
@@ -1364,7 +1371,7 @@ void CSigSharesManager::WorkThreadMain()
 
         // TODO Wakeup when pending signing is needed?
         if (!didWork) {
-            if (!workInterrupt.sleep_for(std::chrono::milliseconds(100))) {
+            if (!workInterrupt.sleep_for(std::chrono::milliseconds(1000))) {
                 return;
             }
         }


src/llmq/quorums_signing_shares.cpp
```diff
@@ -1364,7 +1371,7 @@ void CSigSharesManager::WorkThreadMain()
 
         // TODO Wakeup when pending signing is needed?
         if (!didWork) {
-            if (!workInterrupt.sleep_for(std::chrono::milliseconds(100))) {
+            if (!workInterrupt.sleep_for(std::chrono::milliseconds(1000))) {
```
aleflm, line 1374: Overall, PR looks correct to me. But I suspect that 1000ms is too high for this `sleep_for`. Maybe something like 250ms is more reasonable. Just a suggestion. Since this is not going to make a huge difference. 

Overall, I don't see any specific issues with the PR.

reubenyap:
bugbot run

PR #1690, opened 2025-10-16
Fix Symbolification with CMake on macOS

cassandras-lies:
## PR intention
Fix symbolification.

## Code changes brief
- Remove symbol stripping from macOS Debug builds.
- Add a cmake option -DUSE_ABSOLUTE_DSYM_PATHS to be friendly to debuggers.
- Add debugging information to fix issues with debug symbols not appearing for files with duplicated basenames.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -352,6 +352,63 @@ add_library(firo_node STATIC EXCLUDE_FROM_ALL
   ${CMAKE_CURRENT_SOURCE_DIR}/warnings.cpp
 )
 
+# Fix duplicate object file names to prevent debug symbol conflicts
+# This ensures that files with the same basename in different directories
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -352,6 +352,63 @@ add_library(firo_node STATIC EXCLUDE_FROM_ALL
   ${CMAKE_CURRENT_SOURCE_DIR}/warnings.cpp
 )
 
+# Fix duplicate object file names to prevent debug symbol conflicts
+# This ensures that files with the same basename in different directories
+# get unique object file names, preventing dsymutil from skipping debug symbols
+set_source_files_properties(
```
aleflm, line 358: I prefer not to have hardcoded object file names.

cassandras-lies:
Second commit is broken. Will hand to @aleflm to fix.

Issue #1530, opened 2025-02-25
Auto-anonymize crashes non-HD wallet


justanwar:
Due to the extremely low amount of users using a non-HD wallet currently, this can be considered a low priority bug.

Running latest master commit https://github.com/firoorg/firo/commit/0375ddef3e15b8da50a572c95b7ff50e31cfd780

```
2025-02-25 19:40:38 UpdateTip() pindexNew.nHeight=173606
2025-02-25 19:40:38 UpdateTip: new best=78eb2a30805c620d15ae1fe9c74addb2dfa66086b4f1f4e0e883df6c2d96c8ca height=173606 version=0x20001000 log2_work=46.627188 tx=353761 date='2025-02-25 19:40:39' progress=1.000000 cache=0.0MiB(48txo)
2025-02-25 19:40:40 GUI: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

2025-02-25 19:40:40 receive version message: /Satoshi:0.14.14/: version 90031, blocks=173606, us=210.186.24.16:61248, peer=12
2025-02-25 19:40:40 GUI: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

2025-02-25 19:40:41 

************************
EXCEPTION: Exception: type=std::logic_error, what="Spark feature requires HD wallet"
   0#: (0x007929DB) stl_list.h:414         - std::__cxx11::_List_base<CLelantusEntry, std::allocator<CLelantusEntry> >::_M_set_size(unsigned long long)
   1#: (0x007929DB) stl_list.h:1492        - std::__cxx11::list<CLelantusEntry, std::allocator<CLelantusEntry> >::swap(std::__cxx11::list<CLelantusEntry, std::allocator<CLelantusEntry> >&)
[...]

PR #1709, opened 2025-11-10
Crash on Non HD wallets fix

levonpetrosyan93:
Fix for issue #1530

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -334,7 +334,7 @@ void ReceiveCoinsDialog::copyAmount()
 
 void ReceiveCoinsDialog::displayCheckBox(int idx)
 {
-    if(ui->addressTypeCombobox->currentText() == "Spark"){
+    if(ui->addressTypeCombobox->currentText() == tr("Spark")){
         ui->reuseAddress->hide();
         ui->createSparkNameButton->setVisible(true);
     } else {


justanwar:
Tested. Crashes when generating Spark address in Receive tab (should show error/warning instead).

src/qt/receivecoinsdialog.cpp
```diff
@@ -327,7 +334,7 @@ void ReceiveCoinsDialog::copyAmount()
 
 void ReceiveCoinsDialog::displayCheckBox(int idx)
 {
-    if(idx==0){
+    if(ui->addressTypeCombobox->currentText() == "Spark"){
```
aleflm: This will cause issues with Localization. 

Something like this (roughly) might be better:

```
// Only works if "Spark" is translated exactly the same way here as in the .ui file
if(ui->addressTypeCombobox->currentText() == tr("Spark")){ 
```

src/qt/receivecoinsdialog.cpp
```diff
@@ -122,6 +122,13 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
         tableView->horizontalHeader()->setMinimumSectionSize(23);
         tableView->horizontalHeader()->setStretchLastSection(true);
 
+        auto wallet = _model->getWallet();
+        if (!wallet || !wallet->sparkWallet) {
+            ui->addressTypeCombobox->removeItem(0);
```
aleflm, line 127: This assumes item 0 is always “Spark”. Is this a correct assumption? ( I am not sure, just asking ).

aleflm (approved):
Other than two minor issues I raised. Everything looks good to me.

src/qt/receivecoinsdialog.cpp
levonpetrosyan93: Yes, item 0 at this point is always "Spark"

src/qt/receivecoinsdialog.cpp
levonpetrosyan93: agree, fixed

PR #1723, opened 2025-12-17
Reduce memory consumption of checking Spark spend transactions

psolstice:
## PR intention
Reduces memory footprint.

## Code changes brief
A number of measures including limiting a number of concurrently queried transactions and number of threads in the thread pool designated for the checks.

### DIFF
--- a/src/spark/state.cpp
+++ b/src/spark/state.cpp
@@ -24,7 +24,7 @@ struct ProofCheckState {
 static std::map<uint256, ProofCheckState> gCheckedSparkSpendTransactions;
 static CCriticalSection cs_checkedSparkSpendTransactions;
 
-static ParallelOpThreadPool<bool> gCheckProofThreadPool(std::max(boost::thread::hardware_concurrency(), 4u));
+static ParallelOpThreadPool<bool> gCheckProofThreadPool(std::min(boost::thread::hardware_concurrency(), 4u));
 
 static CSparkState sparkState;
 


src/spark/state.cpp
```diff
@@ -24,7 +24,7 @@ struct ProofCheckState {
 static std::map<uint256, ProofCheckState> gCheckedSparkSpendTransactions;
 static CCriticalSection cs_checkedSparkSpendTransactions;
 
-static ParallelOpThreadPool<bool> gCheckProofThreadPool(boost::thread::hardware_concurrency());
+static ParallelOpThreadPool<bool> gCheckProofThreadPool(std::max(boost::thread::hardware_concurrency(), 8u));
```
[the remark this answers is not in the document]
aleflm: @psolstice I confirm Coderabbit's comment here. `std::max` is not correct here.

src/spark/state.cpp
```diff
@@ -804,17 +804,23 @@ bool CheckSparkSpendTransaction(
...
-                        } catch (const std::exception &) {
-                            return false;
-                        }
-                    });
-                    auto &checkState = gCheckedSparkSpendTransactions[hashTx];
-                    checkState.fChecked = false;
-                    checkState.fResult = false;
-                    checkState.checkInProgress = std::make_shared<boost::future<bool>>(std::move(future));
+                    // don't post a request if there are too many tasks already
+                    if (gCheckProofThreadPool.GetPendingTaskCount() < (std::size_t)gCheckProofThreadPool.GetNumberOfThreads()/2) {
```
aleflm, line 808: Just asking. I am not 100% sure. It wouldn't be better to have this: 

```
if (gCheckProofThreadPool.GetPendingTaskCount() < ...) {
    gCheckProofThreadPool.PostTask(...);
}
``` 

as atomic operation?

psolstice: it's not that important right here, it's just an optimization, even if the number of tasks changes it won't cause any real damage

reubenyap:
@coderabbitai review

PR #1777, opened 2026-02-02
Subtract remaining amount after sigma and lelantus pull closed

levonpetrosyan93:
Sigma and Lelantus pools have already been closed,
As we allowed to spend sigma coins by Lelantus spend transactions, so we need to take in account remaining numbers from 2 pools.
Sigma number [https://explorer.firo.org/address/Sigma](here) : 182463.94882676
Lelantus number [https://explorer.firo.org/address/Lelantus](here) : -86771.07491993
As we had some amount of forged coins and we added that figures into our total supply number, it would be fair to subtract the remainder from this 2 figures, which is 95692.8739068 .
This PR introduces all required changes to do it.

---

> [!NOTE]
> **Medium Risk**
> Changes the `gettotalsupply` RPC calculation by subtracting additional pool balances, which can affect reported monetary supply used by explorers/analytics. Logic relies on `-addressindex` data and new address-type sums, so incorrect indexing or type selection could skew outputs.
> 
> **Overview**
> Adjusts the `gettotalsupply` RPC to also subtract the remaining balances in the **closed Sigma and Lelantus pools**, in addition to the existing Zerocoin/CVE/burn adjustments.
> 
> Introduces new addressindex-backed helpers and RPCs, `getsigmapoolbalance` and `getlelantuspoolbalance`, which compute pool remainders by summing `AddressType` index entries for the relevant mint/spend categories, and registers these commands in the RPC table.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5dcd7fced043479ed4385d63ad8ab8ea105cc6a5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -1941,6 +1941,52 @@ CAmount getzerocoinpoolbalance()
     return  nTotalAmount;
 }
 
+CAmount getsigmapoolbalance()
+{
+    CAmount nTotalAmount = 0;
+
+    // Iterate over all mints
+    std::vector<std::pair<CAddressIndexKey, CAmount> > addressIndex;
+    if (GetAddressIndex(uint160(), AddressType::sigmaMint, addressIndex)) {
+        for (auto& it : addressIndex) {
+            nTotalAmount += it.second;
+        }
+    }
+    addressIndex.clear();
+
+    // Iterate over all spends
+    if (GetAddressIndex(uint160(), AddressType::sigmaSpend, addressIndex)) {
+        for (auto& it : addressIndex) {
+            nTotalAmount += it.second;
+        }
+    }
+
+    return nTotalAmount;
+}
+
+CAmount getlelantuspoolbalance()
+{
+    CAmount nTotalAmount = 0;
+
+    // Iterate over all mints
+    std::vector<std::pair<CAddressIndexKey, CAmount> > addressIndex;
+    if (GetAddressIndex(uint160(), AddressType::lelantusMint, addressIndex)) {
+        for (auto& it : addressIndex) {
+            nTotalAmount += it.second;
+        }
+    }
+    addressIndex.clear();
+
+    // Iterate over all spends
+    if (GetAddressIndex(uint160(), AddressType::lelantusJSplit, addressIndex)) {
+        for (auto& it : addressIndex) {
+            nTotalAmount += it.second;
+        }
+    }
+
+    return nTotalAmount;
+}
+
 CAmount getCVE17144amount()
 {
     // CVE-2018-17144 was a critical bug that allowed double-spending of inputs
@@ -2090,6 +2136,7 @@ UniValue gettotalsupply(const JSONRPCRequest& request)
     total -= getzerocoinpoolbalance(); //498,397.00000000 The actual amount of coins forged during the Zerocoin attacks (the negative balance after the pool closed),
     total += getCVE17144amount(); //320,841.99803185 The cmount of forged coins during CVE-2018-17144 attacks,
     total -= 16810168037465;// burnt Coins sent to unrecoverable address https://explorer.firo.org/tx/0b53178c1b22bae4c04ef943ee6d6d30f2483327fe9beb54952951592e8ce368
+    total -= getsigmapoolbalance() + getlelantuspoolbalance(); // after closing sigma and lelantus pools, we have some amount of coins left in the pools. So we need to subtract them from the total supply.
     UniValue result(UniValue::VOBJ);
     result.push_back(Pair("total", total));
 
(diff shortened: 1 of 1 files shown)


navidR (approved):

reubenyap:
@cursor review

reubenyap:
Need to calculate end supply before tail emission hit before approving this.

reubenyap:
## End supply before tail emission

Per @reubenyap's request, here is the pre-tail-emission supply calculation, plus a review of the accounting change itself.

**Short version:** the chain emits *past* 21.4M before tail emission — it does not stop at it. With this PR applied, reported supply crosses 21,400,000 at roughly block **1,773,800** (~late July 2028) and tail emission begins ~24,900 blocks (~6 weeks) later at block **1,798,655**, by which point reported supply is **≈ 21,555,400 FIRO — about 155,400 above 21.4M**. Without this PR it would be ≈ 21,651,100 (~251k above), so the PR moves the figure toward 21.4M but does not land on it.

### Emission schedule

From `GetBlockSubsidyWithMTPFlag` (`src/validation.cpp:1695`) and mainnet params (`src/chainparams.cpp:199-219`). Tail emission is purely height-triggered at `stage4StartBlock + nSubsidyHalvingInterval` = 958,655 + 840,000 = **1,798,655**; there is no supply-based trigger anywhere (`MAX_MONEY` in `amount.h` is only a per-amount sanity bound).

| Era | Blocks | Reward | Emits |
|---|---|---|---|
| Launch → MTP (10-min) | 1 – 117,561 | 50 | 5,878,050 |
| MTP → 1st halving (5-min) | 117,562 – 302,437 | 25 | 4,621,900 |
| 1st halving → stage 3 (5-min) | 302,438 – 486,220 | 12.5 | 2,297,287.5 |
| Stage 3 → stage 4 (2.5-min) | 486,221 – 958,654 | 6.25 | 2,952,712.5 |
| Stage 4 (2.5-min, no reward cut) | 958,655 – 1,798,654 | 6.25 | 5,250,000 |
| **Total coinbase before tail** | | | **≈ 21,000,000** |
| Tail from 1,798,655 | forever | 1 (4·COIN ÷2÷2) | ~210,384/yr (~0.98% initial) |

Applying the `gettotalsupply` adjustments gives the reported figure when the tail begins:

```
  21,000,000.00   scheduled coinbase emission
 +    498,397.00   zerocoin pool negative balance (forged)
 +    320,842.00   CVE-2018-17144 forged (block 293526)
 -    168,101.68   burn tx 0b53178c…
 -     95,692.87   sigma + lelantus locked remainder  ← this PR
 = 21,555,400     (± exact MTP switch block, unclaimed coinbase, and Remint below)
```

If the requirement is "reach 21.4M before the tail starts", that is satisfied with ~6 weeks to spare. If the requirement is "end *at* 21.4M when the tail starts", it overshoots by ~155k. Landing exactly on 21.4M would need the stage-4 interval shortened from 840,000 to ~815,129 blocks (a consensus change), or an additional ~155k burned — otherwise ~21.55M is the terminal pre-tail figure and that should be stated deliberately rather than inherited.

### Review of the change

The concept and sign conventions are right. Summing sigma and lelantus **together** is exactly correct, since Lelantus joinsplits could spend sigma coins — which is why Sigma reads +182,463.95 and Lelantus −86,771.07, and only the sum is meaningful. The fee subtlety also cancels: joinsplit/sigma-spend fees are absent from `nFees` (inflating the DB supply counter) and equally absent from the index outflow entries (inflating the pool balance), so the subtraction nets them out.

**One real bug — the `zerocoinRemint` bucket is missed.** Zerocoin→sigma remints are indexed as `zerocoinRemint: −R` plus `sigmaMint: +R` (`src/txdb.cpp:617-630`, `720-730`). `getzerocoinpoolbalance()` sums only `zerocoinMint + zerocoinSpend`, and nothing sums `zerocoinRemint`. Those reminted coins therefore remain counted as phantom zerocoin-pool content *and*, with this PR, now also register as sigma inflow — subtracted twice, under-reporting supply by the total ever reminted. Suggest including `AddressType::zerocoinRemint` in the `gettotalsupply` subtraction (its sum is negative, so it adds R back). Worth confirming R on an addressindex node via `getaddressbalance {"addresses":["Remint"]}` before merging — note the correction moves end supply *further above* 21.4M.

**Minor:**
- `getlelantuspoolbalance()` omits `lelantusJMint`. Numerically harmless (JMint outputs carry `nValue = 0`), but the explorer's "Lelantus" pseudo-address is mint + jmint + jsplit (`src/rpc/misc.cpp:697-700`), so including it costs nothing and keeps parity.
- If `GetAddressIndex` fails, the new helpers silently return 0 and `gettotalsupply` quietly skips the adjustment. Mostly mitigated since `ReadTotalSupply` also only exists with `-addressindex`, but a stale DB key after disabling the index would yield a silently wrong total. Same pre-existing pattern as `getzerocoinpoolbalance`.
- The three pool functions are now copy-paste triplets; a shared helper taking two `AddressType`s would collapse them.
- Help text advertises `"total"` as a string but the RPCs return a raw numeric CAmount in satoshis.

**Recommendation:** hold approval until the Remint bucket is handled and its mainnet amount confirmed, since it shifts the reported figure in the wrong direction relative to 21.4M — and separately decide explicitly whether ~21.55M at tail start is the intended terminal supply.

---

PR #1790, opened 2026-03-13
Respect -onlynet flag when setting proxy network limits

reubenyap:
## PR intention
Fix the remaining `-onlynet` edge cases on top of the existing PR #1790 work, while keeping the behavior close to Bitcoin Core where it fits Firo's smaller supported network set.

This keeps the original goal of preventing `-onlynet` from being silently overridden, and also tightens the handling of supported network types in init and RPC paths.

## Code changes brief
- Keep PR #1790's explicit-limit tracking and async Tor auth callback protection so `NET_ONION` is not re-enabled when `-onlynet` explicitly excludes it.
- Narrow the `-onlynet` limiting loop in `src/init.cpp` to only the user-selectable networks (`NET_IPV4`, `NET_IPV6`, `NET_ONION`), and apply both `SetLimited()` and `SetNetworkExplicitlyLimited()` only to that set.
- Reject the contradictory combination `-onlynet=onion` with `-onion=0` using Bitcoin-style init error wording.
- Limit `getnetworkinfo` network reporting to `ipv4`, `ipv6`, and `onion` so unsupported enum values are excluded by default.
- Make `GetNetworkName()` retain the existing case labels but provide a safe debug-assert + release fallback of `"unknown"` for unexpected enum values.
- Keep the PR's deprecated `tor` alias compatibility, but emit the warning only once per process lifetime.
- Extend targeted coverage in `src/test/netbase_tests.cpp` and `qa/rpc-tests/proxy_test.py` for the updated helper behavior and RPC network list.

<div><a href="https://cursor.com/agents/bc-7cfd5bea-cb16-4196-abbb-77c61a354883"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7cfd5bea-cb16-4196-abbb-77c61a354883"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

___

## **CodeAnt-AI Description**
**Show the correct network for peers and keep Tor-related network limits in place**

### What Changed
- The peer list and `getpeerinfo` now show a `network` value, including `onion` for peers that connect through the node’s Tor hidden service instead of appearing as `127.0.0.1`
- `getnetworkinfo` now reports only the supported networks: `ipv4`, `ipv6`, and `onion`
- `-onlynet` restrictions are no longer overridden by proxy, Tor, or hidden-service setup paths, and `-onlynet=onion` now fails when Tor access is explicitly disabled with `-onion=0`
- The old `tor` network name still works, but now warns once that `onion` should be used instead

### Impact
`✅ Clearer peer network display`
`✅ Fewer hidden-service peers misreported as localhost`
`✅ More reliable Tor-only connection limits`

[🔄 Retrigger CodeAnt AI Review](https://api.codeant.ai/pr/retrigger_review?token=GIIhgHdO9byetx6I77YS8-HFzw2MXlVp09Ogf9CX-Zw&org=firoorg)<details>
<summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
This lets you have a chat with CodeAnt AI about your pull request, making it easier to understand and improve your code.

#### Example
<pre>
<code>@codeant-ai ask: Can you suggest a safer alternative to storing this secret?</code>
</pre>

### Preserve Org Learnings with CodeAnt
You can record team preferences so CodeAnt AI applies them in future reviews. Reply directly to the specific CodeAnt AI suggestion (in the same thread) and replace "Your feedback here" with your input:
<pre>
<code>@codeant-ai: Your feedback here</code>
</pre>
This helps CodeAnt AI learn and adapt to your team's coding style and standards.

#### Example
<pre>
<code>@codeant-ai: Do not flag unused imports.</code>
</pre>

### Retrigger review
Ask CodeAnt AI to review the PR again, by typing:
<pre>
<code>@codeant-ai: review</code>
</pre>

### Check Your Repository Health
To analyze the health of your code repository, visit our dashboard at [https://app.codeant.ai](https://app.codeant.ai). This tool helps you identify potential issues and areas for improvement in your codebase, ensuring your repository maintains high standards of code health.

</details>


levonpetrosyan93 (dismissed):

src/netbase.cpp
```diff
@@ -44,18 +44,23 @@ enum Network ParseNetwork(std::string net) {
     boost::to_lower(net);
     if (net == "ipv4") return NET_IPV4;
     if (net == "ipv6") return NET_IPV6;
-    if (net == "tor" || net == "onion")  return NET_ONION;
+    if (net == "onion") return NET_ONION;
```
reubenyap, line 55: **🔴 Breaking change: Removing "tor" alias from ParseNetwork()**

The old code accepted both `"tor"` and `"onion"` as valid network names:
```cpp
if (net == "tor" || net == "onion") return NET_ONION;
```

This PR removes `"tor"`, so `-onlynet=tor` will now fail with `"Unknown network specified in -onlynet: 'tor'"`. Users who have `onlynet=tor` in their `firo.conf` files will get a startup error after upgrading.

This is a backwards-incompatible change. Consider:
1. Keeping `"tor"` as a deprecated alias (log a warning but still accept it), or
2. Documenting this in release notes as a known breaking change

Bitcoin Core still accepts both `"tor"` and `"onion"` as of v28.

src/netbase.cpp
```diff
@@ -44,18 +44,23 @@ enum Network ParseNetwork(std::string net) {
...
     {
+    case NET_UNROUTABLE: return "not_publicly_routable";
     case NET_IPV4: return "ipv4";
     case NET_IPV6: return "ipv6";
     case NET_ONION: return "onion";
-    default: return "";
+    case NET_I2P: return "i2p";
+    case NET_CJDNS: return "cjdns";
+    case NET_INTERNAL: return "internal";
+    case NET_MAX: assert(false);
```
reubenyap: **🟡 Bug risk: `assert(false)` is UB in release builds with NDEBUG**

`case NET_MAX: assert(false);` and the trailing `assert(false);` become no-ops when compiled with `-DNDEBUG` (which is common in release builds). If `NET_MAX` is passed:
1. `assert(false)` → `((void)0)` → falls through the case
2. Trailing `assert(false)` → `((void)0)` → function returns without a value
3. This is **undefined behavior** for a function returning `std::string`

The intent (exhaustive switch with compiler warnings on missing cases) is good. But add a fallback for safety:
```cpp
case NET_MAX: break; // intentionally not handled
}
assert(false);
return "";
```

Or use `__builtin_unreachable()` on GCC/Clang to preserve the optimization hint while avoiding UB.

src/rpc/net.cpp
```diff
@@ -384,7 +384,7 @@ static UniValue GetNetworksInfo()
     for(int n=0; n<NET_MAX; ++n)
     {
         enum Network network = static_cast<enum Network>(n);
-        if(network == NET_UNROUTABLE)
+        if(network == NET_UNROUTABLE || network == NET_I2P || network == NET_CJDNS || network == NET_INTERNAL)
```
reubenyap: **🔵 Question: Hardcoded network filter vs future extensibility**

This filters `NET_I2P`, `NET_CJDNS`, and `NET_INTERNAL` from `getnetworkinfo` output. This makes sense for the current codebase since these networks aren't supported.

However, PR #1806 (I2P support) is open in parallel and would need to update this filter. Consider a more future-proof approach:
```cpp
if (network == NET_UNROUTABLE || network == NET_INTERNAL)
    continue;
```

Then `NET_I2P` and `NET_CJDNS` would appear in the RPC output as `limited: true, reachable: false` when not configured, which is informative rather than hidden. This matches Bitcoin Core's behavior where I2P appears in `getnetworkinfo` even when not configured.

src/net.cpp
```diff
@@ -289,6 +290,22 @@ bool IsLimited(const CNetAddr &addr)
     return IsLimited(addr.GetNetwork());
 }
 
+void SetNetworkExplicitlyLimited(enum Network net, bool fLimited)
```
reubenyap, line 293: **🟢 Praise: Clean explicit-limited API**

The `SetNetworkExplicitlyLimited()` / `IsNetworkExplicitlyLimited()` API elegantly solves the async callback problem. The bounds checking (`net == NET_UNROUTABLE || net >= NET_MAX`) in both functions prevents array out-of-bounds access, and the `LOCK(cs_mapLocalHost)` ensures thread safety for the async `auth_cb` callback.

Good that the default parameter is `bool fLimited = true` matching the existing `SetLimited()` signature for consistency.

src/torcontrol.cpp
```diff
@@ -465,7 +465,8 @@ void TorController::auth_cb(TorControlConnection& _conn, const TorControlReply&
             CService resolved(LookupNumeric("127.0.0.1", 9050));
             proxyType addrOnion = proxyType(resolved, true);
             SetProxy(NET_ONION, addrOnion);
-            SetLimited(NET_ONION, false);
+            if (!IsNetworkExplicitlyLimited(NET_ONION))
```
reubenyap, line 475: **🟢 Good: Correct async guard**

This is the key fix. The `auth_cb` callback runs asynchronously after `AppInitMain` completes, so it can't access the local `fOnlyNet`/`onlyNetNets` variables. Using `IsNetworkExplicitlyLimited(NET_ONION)` correctly queries the persistent global state set during init.

The guard `if (!IsNetworkExplicitlyLimited(NET_ONION))` ensures that if the user specified `-onlynet=ipv4` (which explicitly excludes onion), Tor authentication won't silently re-enable onion connectivity.

src/init.cpp
```diff
@@ -1529,18 +1529,21 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+    std::set<enum Network> onlyNetNets;
+    if (fOnlyNet) {
         BOOST_FOREACH(const std::string& snet, mapMultiArgs.at("-onlynet")) {
             enum Network net = ParseNetwork(snet);
             if (net == NET_UNROUTABLE)
                 return InitError(strprintf(_("Unknown network specified in -onlynet: '%s'"), snet));
-            nets.insert(net);
+            onlyNetNets.insert(net);
         }
         for (int n = 0; n < NET_MAX; n++) {
```
reubenyap: **🟢 Good: SetNetworkExplicitlyLimited called alongside SetLimited**

Calling `SetNetworkExplicitlyLimited(net)` inside the same loop as `SetLimited(net)` ensures the two arrays stay in sync. The "explicitly limited" flag is set for all networks NOT in the -onlynet list, which correctly represents the user's intent.

One minor note: `SetNetworkExplicitlyLimited` is only called in this one place (the -onlynet loop), so the `vfExplicitlyLimited[]` array purely reflects -onlynet state. The naming could be slightly more specific (e.g., `SetNetworkOnlynetExcluded`), but the current naming works fine and is more general-purpose.

reubenyap (commented):
## AI Code Review — PR #1790

**Overall Assessment: Approve with one breaking change concern**

This PR fixes a real, user-impacting bug where `-onlynet` restrictions were silently overridden by `-proxy`, `-onion`, and `-torsetup` code paths. The fix is clean and well-reasoned: local `fOnlyNet`/`onlyNetNets` variables capture the intent, and a new `vfExplicitlyLimited[]` parallel array exposes the state to the async `TorController::auth_cb` callback.

### Architecture Review

**The core fix is correct:** Every `SetLimited(NET_XXX, false)` call in init.cpp is now guarded by `if (!fOnlyNet || onlyNetNets.count(NET_XXX))`, which reads naturally: "only unlimit this network if -onlynet wasn't specified, or if this network was explicitly included." The `-torsetup` path correctly removes the old limit-then-unlimit no-op pattern.

**The async callback fix is well-designed:** `SetNetworkExplicitlyLimited()` persists the -onlynet decision into a global `vfExplicitlyLimited[]` array that `auth_cb` can query later. The bounds checking in both getter and setter (`net == NET_UNROUTABLE || net >= NET_MAX`) prevents out-of-bounds access. Both use `LOCK(cs_mapLocalHost)` for thread safety.

### Findings Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 breaking | netbase.cpp | Removing "tor" alias from ParseNetwork() |
| 2 | 🟡 bug risk | netbase.cpp | assert(false) in GetNetworkName is UB in release builds |
| 3 | 🔵 question | rpc/net.cpp | Hardcoded NET_I2P/NET_CJDNS filter vs future extensibility |
| 4 | 🟢 praise | net.cpp + torcontrol.cpp | Clean explicit-limited state for async callback |
| 5 | 🟢 praise | init.cpp | Consistent guard pattern across all SetLimited calls |

*Static analysis: No findings specific to changed lines. cppcheck found pre-existing uninitialized member warnings in headers.*

src/netbase.cpp
```diff
@@ -44,18 +44,28 @@ enum Network ParseNetwork(std::string net) {
...
     case NET_IPV4: return "ipv4";
     case NET_IPV6: return "ipv6";
     case NET_ONION: return "onion";
-    default: return "";
+    case NET_I2P: return "i2p";
+    case NET_CJDNS: return "cjdns";
+    case NET_INTERNAL: return "internal";
+    case NET_MAX: break;
     }
+    assert(false);
```
reubenyap: Test comment on assert(false)

reubenyap (commented):
Test review

src/init.cpp
```diff
@@ -1529,18 +1529,21 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
...
             enum Network net = (enum Network)n;
-            if (!nets.count(net))
+            if (!onlyNetNets.count(net)) {
                 SetLimited(net);
```
reubenyap, line 1607: test

reubenyap (commented):
line test

src/init.cpp
```diff
@@ -1529,18 +1529,21 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+                SetNetworkExplicitlyLimited(net);
```
reubenyap: test

reubenyap (commented):
line test

reubenyap (commented):
line test

src/rpc/net.cpp
reubenyap: test

src/init.cpp
```diff
@@ -1606,21 +1610,15 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
             if (!addrOnion.IsValid())
                 return InitError(strprintf(_("Invalid -onion address: '%s'"), onionArg));
             SetProxy(NET_ONION, addrOnion);
-            SetLimited(NET_ONION, false);
+            if (!fOnlyNet || onlyNetNets.count(NET_ONION))
```
reubenyap, line 1731: test

reubenyap (commented):
test

reubenyap (commented):
test

reubenyap (commented):
test

reubenyap (commented):
test

reubenyap (commented):
## PR #1790 Review: Respect -onlynet flag when setting proxy network limits

**[7 files changed, +70 / -27]** — Prevents `-proxy`, `-onion`, `-torsetup`, and the async `TorController::auth_cb` callback from silently overriding `-onlynet` network restrictions. Introduces `SetNetworkExplicitlyLimited`/`IsNetworkExplicitlyLimited` to persist the user's intent.

### Static Analysis
cppcheck: Clean on changed files (no new issues). flawfinder: 25 hits, all pre-existing (fopen, atoi, char buffer in torcontrol.cpp) — none introduced by this PR. semgrep (Firo rules): 117 hits across scanned files, all pre-existing — none in lines changed by this PR. clang-tidy: Not run (needs compile_commands.json).

### Findings Summary
| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | MEDIUM | `src/netbase.cpp:67` | `assert(false)` is UB in release builds |
| 2 | MEDIUM | `src/init.cpp:1606` | `-noonion` silently overrides `-onlynet=onion` |
| 3 | LOW | `src/rpc/net.cpp:387` | Hardcoded network skip list is fragile |
| 4 | LOW | `src/netbase.cpp:48` | Deprecation warning fires unconditionally |
| 5 | LOW | `src/init.cpp:1545` | Loop marks unused networks as explicitly limited |

> Each finding is posted as an inline comment on the relevant code — click to view, resolve when addressed. Each comment includes a **Prompt for AI agents** that can be copy-pasted into Cursor/Copilot/Claude Code to auto-fix the issue.

---
*Review assisted by AI + static analysis (cppcheck, flawfinder, semgrep). Posted by @reubenyap.*

src/netbase.cpp
reubenyap: **[MEDIUM] `assert(false)` is undefined behavior in release builds**

In release builds where `NDEBUG` is defined, `assert(false)` compiles to nothing. This means `GetNetworkName()` will silently fall through and return `""` for any unexpected `Network` value (e.g. if someone adds a new enum entry but forgets to update this switch). The test at line 52 of `netbase_tests.cpp` even acknowledges this by guarding with `#ifdef NDEBUG`.

**Suggested fix:**
Replace `assert(false); return "";` with a pattern that is safe in both debug and release builds, e.g.:
```cpp
default: return "unknown";
```
or if you want a crash in debug but safe fallback in release:
```cpp
assert(!"unexpected Network value");
return "";
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/netbase.cpp`, in the `GetNetworkName()` function around line 67, `assert(false)` is used as the fallback after the switch statement. In release builds (`NDEBUG`), this assert is compiled out, causing the function to silently return an empty string for unknown network values. Replace the `case NET_MAX: break; } assert(false); return \"\";` pattern with a safe fallback: remove the `assert(false)` and change the return to `return \"unknown\";` or use `assert(!\"unexpected Network value\"); return \"\";` which preserves the debug crash but provides a safe release fallback. Keep all existing case labels unchanged."

</details>

src/init.cpp
reubenyap, line 1731: **[MEDIUM] `-noonion` silently overrides `-onlynet=onion`**

If a user specifies `-onlynet=onion` (meaning "only connect via Tor") but also has `-noonion` or `-onion=0`, line 1606 unconditionally calls `SetLimited(NET_ONION)` — disabling the only network the user explicitly requested. This creates a contradictory state where all networks are limited, effectively preventing any outbound connections.

The `-onlynet` guard (`if (!fOnlyNet || onlyNetNets.count(...))`) is applied to the `SetLimited(..., false)` calls like on this line, but not to the `-noonion` branch above it (line 1605-1606). That path should either respect `-onlynet` or emit a clear error for the contradictory configuration.

**Suggested fix:**
Add a check before limiting onion:
```cpp
if (onionArg == "0") {
    if (fOnlyNet && onlyNetNets.count(NET_ONION)) {
        return InitError(_("Cannot use -noonion/-onion=0 together with -onlynet=onion"));
    }
    SetLimited(NET_ONION);
}
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, around line 1604-1606 in the `-onion` handling section, the `-noonion` (`onionArg == \"0\"`) branch unconditionally calls `SetLimited(NET_ONION)` which overrides any prior `-onlynet=onion` setting. Add a conflict check: before `SetLimited(NET_ONION)` on line 1606, check if `fOnlyNet && onlyNetNets.count(NET_ONION)` and if so, return an `InitError` explaining that `-noonion` and `-onlynet=onion` are contradictory. Keep the existing `SetLimited(NET_ONION)` call for the non-conflicting case."

</details>

src/rpc/net.cpp
reubenyap: **[LOW] Hardcoded network skip list in `GetNetworksInfo` is fragile**

The skip condition `network == NET_UNROUTABLE || network == NET_I2P || network == NET_CJDNS || network == NET_INTERNAL` hardcodes which networks to exclude from the RPC response. If a new `Network` enum value is added in the future, this list must be manually updated or the RPC will try to report on an unsupported network.

Consider using a positive inclusion list (only show NET_IPV4, NET_IPV6, NET_ONION) or adding a helper like `IsRoutableNetwork()` that can be maintained alongside the enum.

**Suggested fix:**
Replace the exclusion list with an inclusion approach:
```cpp
if (network != NET_IPV4 && network != NET_IPV6 && network != NET_ONION)
    continue;
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/rpc/net.cpp`, in the `GetNetworksInfo()` function around line 387, the condition `if(network == NET_UNROUTABLE || network == NET_I2P || network == NET_CJDNS || network == NET_INTERNAL)` uses a fragile exclusion list. Replace it with a positive inclusion list: `if (network != NET_IPV4 && network != NET_IPV6 && network != NET_ONION) continue;` This ensures that only known-supported networks are reported, and any new enum values are safely excluded by default."

</details>

src/netbase.cpp
```diff
@@ -44,18 +44,28 @@ enum Network ParseNetwork(std::string net) {
     boost::to_lower(net);
     if (net == "ipv4") return NET_IPV4;
     if (net == "ipv6") return NET_IPV6;
-    if (net == "tor" || net == "onion")  return NET_ONION;
+    if (net == "tor") {
+        LogPrintf("Warning: net name 'tor' is deprecated and will be removed in the future. You should use 'onion' instead.\n");
```
reubenyap: **[LOW] Deprecation warning for `tor` alias uses unconditional `LogPrintf`**

The deprecation message for the "tor" network name alias uses `LogPrintf` which prints unconditionally on every call to `ParseNetwork("tor")`. For users with `-onlynet=tor` in their config, this fires on every startup and spams the debug log. Consider using a static-once guard so the warning only appears once per process.

Also, the `-onlynet` help text (line 468) should mention that "tor" is deprecated in favor of "onion" so users know to update their configs.

**Suggested fix:**
Use a static-once pattern:
```cpp
if (net == "tor") {
    static bool warned = false;
    if (!warned) {
        LogPrintf("Warning: net name 'tor' is deprecated, use 'onion' instead.\n");
        warned = true;
    }
    return NET_ONION;
}
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/netbase.cpp`, in the `ParseNetwork()` function around line 47-50, the deprecation warning for the 'tor' network name alias uses `LogPrintf` which fires unconditionally. Add a `static bool warned = false;` guard so the warning only prints once per process lifetime, not on every call to `ParseNetwork(\"tor\")`. Keep the return value `NET_ONION` unchanged."

</details>

src/init.cpp
reubenyap: **[LOW] Loop marks unused networks (I2P, CJDNS, INTERNAL) as explicitly limited**

The loop at lines 1541-1547 iterates over all `Network` enum values from 0 to `NET_MAX`. Networks like `NET_I2P`, `NET_CJDNS`, and `NET_INTERNAL` will never be in `onlyNetNets` (users can't pass them via `-onlynet`), so they all get `SetNetworkExplicitlyLimited(true)`. This is currently harmless because `IsNetworkExplicitlyLimited` is only checked for `NET_ONION` in `torcontrol.cpp`, but it could cause subtle bugs if someone later adds an `IsNetworkExplicitlyLimited` check for I2P.

**Suggested fix:**
Limit the loop to only user-facing networks:
```cpp
for (auto net : {NET_IPV4, NET_IPV6, NET_ONION}) {
    if (!onlyNetNets.count(net)) {
        SetLimited(net);
        SetNetworkExplicitlyLimited(net);
    }
}
```

<details>
<summary>🤖 Prompt for AI agents</summary>

"In file `src/init.cpp`, in the `-onlynet` handling block around lines 1541-1547, the loop iterates `for (int n = 0; n < NET_MAX; n++)` which includes `NET_UNROUTABLE`, `NET_I2P`, `NET_CJDNS`, and `NET_INTERNAL` — none of which are user-selectable via `-onlynet`. Replace the loop with an explicit iteration over only the relevant networks: `for (auto net : {NET_IPV4, NET_IPV6, NET_ONION})`. Apply `SetLimited(net)` and `SetNetworkExplicitlyLimited(net)` only to those. Keep the existing error handling for unknown network names above this loop unchanged."

</details>

reubenyap:
@coderabbitai review

PR #1812, opened 2026-03-26
refactor: replace NULL with nullptr in SetCompact signature

operagxoksana:


reubenyap:
For consistency, could you also update the mirrored declaration in src/crypto/MerkleTreeProof/arith_uint256.h in this PR or a small follow-up?

operagxoksana:
Opened a small follow-up PR updating the mirrored declaration #1813

PR #1821, opened 2026-04-19
Spark cryptographic component benchmarks

navidR:
## PR intention

1. Add Spark cryptographic components benchmarks
2. Fix `firo_bench` compilation issue on macOS

### DIFF
--- a/cmake/script/GenerateHeaderFromRaw.cmake
+++ b/cmake/script/GenerateHeaderFromRaw.cmake
@@ -8,16 +8,17 @@ file(READ ${RAW_SOURCE_PATH} hex_content HEX)
 string(REGEX REPLACE "................" "\\0\n" formatted_bytes "${hex_content}")
 string(REGEX REPLACE "[^\n][^\n]" "std::byte{0x\\0}, " formatted_bytes "${formatted_bytes}")
 
+string(LENGTH "${hex_content}" content_length)
+math(EXPR array_size "${content_length} / 2")
+
 set(header_content
-"#include <cstddef>
-#include <span>
+"#include <array>
+#include <cstddef>
 
 namespace ${RAW_NAMESPACE} {
-inline constexpr std::byte detail_${raw_source_basename}_raw[] {
+inline constexpr std::array<std::byte, ${array_size}> ${raw_source_basename} {{
 ${formatted_bytes}
-};
-
-inline constexpr std::span ${raw_source_basename}{detail_${raw_source_basename}_raw};
+}};
 }
 ")
 file(WRITE ${HEADER_PATH} "${header_content}")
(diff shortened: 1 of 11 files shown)


reubenyap (commented):
Thanks for the PR — the framework shape looks reasonable and the `std::span` → `std::array` fix for the generated-header path looks sound. Jumping in as a third-party reader; a few things worth resolving before this lands:

### Blockers

1. **MSVC portability** — `src/bench/spark_bpplus.cpp` uses `__builtin_clzll`, which is GCC/Clang-only and will fail to build on MSVC. The surrounding `1 << (64 - __builtin_clzll(M))` is also UB for small `M`. Prefer `std::bit_width` / `std::countl_zero` (both C++20) or a small portable helper.
2. **Potential UB in `BenchRunner`** — `benchmark.h` calls `std::min_element` on the `measurements` vector and dereferences the result, but if a caller sets `min_iterations = 0` and `min_time_seconds <= 0` that vector is empty. Either guard the stats computation on `!measurements.empty()` or enforce non-zero via the setters.

### Fix before merge

3. **Header guard naming** — `benchmark.h` uses `FIRO_BENCH_BENCHMARK_H`; the rest of the tree still follows the upstream `BITCOIN_BENCH_*` convention (including the matching `#endif` comment).
4. **Silent `perf_event_open` failures** — the `PerfCounters` constructor swallows errors, so users on kernels with `perf_event_paranoid >= 2` will see zeroed counters with no indication why. Please emit a one-shot stderr warning with `errno` and a pointer to `/proc/sys/kernel/perf_event_paranoid`.
5. **Non-reproducible RNG** — `spark_bpplus.cpp`, `spark_schnorr.cpp`, and `spark_grootle.cpp` use unseeded `rand()` with modulo bias to generate test data. Benchmarks want determinism run-to-run; a seeded `std::mt19937_64` (or the project's existing RNG helpers) with an unbiased distribution would be better.
6. **Linux-specific CMake guard** — `src/bench/CMakeLists.txt` uses `if(UNIX AND NOT APPLE)` around the perf-counters block, which also covers the BSDs (no `perf_event_open`). `if(CMAKE_SYSTEM_NAME STREQUAL "Linux")` is more precise.

### Nice to have

7. **Namespace pollution** — file-scope `using namespace spark;` / `using namespace benchmark;` in three of the `spark_*.cpp` files; narrow per-symbol `using` declarations would avoid collisions.
8. **CMake duplication** — the four `add_executable` + `target_link_libraries` blocks for the Spark benches are identical modulo target name; a `foreach(bench IN ITEMS grootle bpplus chaum schnorr)` would shrink this considerably. The new targets also don't seem to link `core_interface` the way other targets do.
9. **Benchmark body duplication** — eight near-identical benchmark bodies in `spark_chaum.cpp` and parallel patterns in BPPlus/Grootle; a single parameterized helper per family would roughly halve the added line count.

### Question

- The `src/bench/mempool_eviction.cpp` edit (dropping `dPriority` from the `CTxMemPoolEntry` constructor call) reads as unrelated to the stated goals of the PR. If it's a drive-by build fix, worth a line in the description — or splitting it out.

PR #1823, opened 2026-04-23
Net: discover Tor SOCKS port via control port

reubenyap:
## Summary
- `auth_cb` now queries `GETINFO net/listeners/socks` to discover the actual Tor SOCKS port instead of hardcoding `127.0.0.1:9050`. Users running Tor on a non-default SocksPort (Tor Browser bundle at 9150, manually configured port) had outbound `.onion` traffic silently routed to a non-existent proxy.
- New `get_socks_cb` callback parses the reply, prefers `127.0.0.1` listeners when multiple are reported, and falls back to `127.0.0.1:9050` if Tor returns nothing usable (preserves pre-fix behaviour as a safety net).
- When clearnet is unreachable (`-onlynet=onion`) and no name proxy is already configured, `get_socks_cb` also sets the discovered proxy as the name proxy. Prevents hostname resolution (DNS seeds on the 11s retry path, `-addnode` via hostname, post-startup lookups) from leaking over clearnet. Guarded by `HaveNameProxy()` so `-proxy`, `-onion`, and `-torsetup` configurations are never overridden. Mirrors the explicit `-onion=<addr>` path at `init.cpp:1737`.

Follows Bitcoin Core's `torcontrol.cpp` pattern for SOCKS port discovery.

## Context
Follow-up to #1790. Testers reported fewer onion peer connections after that merge; root-cause analysis surfaced that the pre-existing hardcoded `127.0.0.1:9050` silently broke onion outbound for non-default Tor configurations. This PR does not fully fix the fresh-install onion-only bootstrap case (that needs onion entries in `pnSeed6_main`, tracked separately), but it does fix:
- Non-default Tor SOCKS port (was silently broken since the feature was added)
- DNS leak when onion-only auto-configuring via `-listenonion`
- DNS seed retrieval on the 11s retry path when `peers.dat` is non-empty and `auth_cb` completes in that window

## Test plan
- [ ] Build with `cmake`/`ninja` — confirms compiles
- [ ] Run with default Tor (SocksPort 9050): `-debug=tor` log shows `tor: Get SOCKS port yielded 127.0.0.1:9050` and `tor: Configuring onion proxy for 127.0.0.1:9050`
- [ ] Run with Tor on a custom port (set `SocksPort 9150` in torrc): log shows discovered port 9150; outbound `.onion` connections succeed
- [ ] Run with `-onlynet=onion -listenonion=1`: after auth completes, `getnetworkinfo` shows onion reachable and DNS-seed log does not show "Skipping DNS seed ... no clearnet network reachable" on subsequent retries
- [ ] Run with `-proxy=127.0.0.1:1080 -listenonion=1`: confirm `HaveNameProxy()` still points to the `-proxy` address (not overridden)
- [ ] Restart Tor daemon during operation: confirm `auth_cb` re-discovers port, peer connectivity resumes
- [ ] Run with Tor returning broken listener response (simulated): fallback to `127.0.0.1:9050` applies; no crash

___

## **CodeAnt-AI Description**
**Discover Tor's SOCKS port before setting up onion traffic**

### What Changed
- The app now asks Tor which SOCKS port it is using instead of assuming `127.0.0.1:9050`, so onion connections work with non-default Tor setups.
- If Tor reports a local SOCKS listener, that address is used; if Tor gives no usable answer, it falls back to `127.0.0.1:9050`.
- When clearnet access is blocked and onion is the only route, hostname lookups are now sent through Tor too, avoiding leaks.
- The chainlock test was updated to wait past the re-enable point before checking chainlock behavior, so the test matches the intended transition.

### Impact
`✅ Onion connections work with Tor Browser and custom SOCKS ports`
`✅ Fewer failed onion peer connections`
`✅ Lower risk of DNS leaks when only onion networking is allowed`

[🔄 Retrigger CodeAnt AI Review](https://api.codeant.ai/pr/retrigger_review?token=6fUB6uy6ebdSTh44hO_8SOovYdwgHJjolYEmhZQUV60&org=firoorg)<details>
<summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
This lets you have a chat with CodeAnt AI about your pull request, making it easier to understand and improve your code.

#### Example
<pre>
<code>@codeant-ai ask: Can you suggest a safer alternative to storing this secret?</code>
</pre>

### Preserve Org Learnings with CodeAnt
You can record team preferences so CodeAnt AI applies them in future reviews. Reply directly to the specific CodeAnt AI suggestion (in the same thread) and replace "Your feedback here" with your input:
<pre>
<code>@codeant-ai: Your feedback here</code>
</pre>
This helps CodeAnt AI learn and adapt to your team's coding style and standards.

#### Example
<pre>
<code>@codeant-ai: Do not flag unused imports.</code>
</pre>

### Retrigger review
Ask CodeAnt AI to review the PR again, by typing:
<pre>
<code>@codeant-ai: review</code>
</pre>

### Check Your Repository Health
To analyze the health of your code repository, visit our dashboard at [https://app.codeant.ai](https://app.codeant.ai). This tool helps you identify potential issues and areas for improvement in your codebase, ensuring your repository maintains high standards of code health.

</details>

### DIFF
--- a/qa/rpc-tests/llmq-cl-evospork.py
+++ b/qa/rpc-tests/llmq-cl-evospork.py
@@ -58,7 +58,8 @@ class LLMQChainLocksTest(EvoZnodeTestFramework):
         ##### Disable chainlocks for 10 blocks
 
         self.nodes[0].importprivkey(self.sporkprivkey)
-        self.disable_chainlocks(self.nodes[0].getblockcount() + 10)
+        reenable_height = self.nodes[0].getblockcount() + 10
+        self.disable_chainlocks(reenable_height)
         self.nodes[0].generate(1)
         assert(False == self.nodes[0].getblock(self.nodes[0].getbestblockhash())["chainlock"])
 
@@ -70,14 +71,19 @@ class LLMQChainLocksTest(EvoZnodeTestFramework):
 
         ##### Enable chainlocks
 
-        self.nodes[0].generate(10)
-        self.nodes[0].spork('list')
         connected_nodes = [n for n in self.nodes if n != self.nodes[5]]
+
+        # Mine up to the re-enable height, then use the next block for
+        # chainlock assertions to avoid the reactivation transition edge.
+        self.nodes[0].generate(reenable_height - self.nodes[0].getblockcount())
         sync_blocks(connected_nodes, timeout=120)
-        self.wait_for_chainlock_tip(connected_nodes, self.nodes[0].getbestblockhash(), timeout=90)
         sporks = self.nodes[0].spork("list")
         assert(not sporks["blockchain"])
         assert(not sporks["mempool"])
+
+        self.nodes[0].generate(1)
+        sync_blocks(connected_nodes, timeout=120)
+        self.wait_for_chainlock_tip(connected_nodes, self.nodes[0].getbestblockhash(), timeout=90)
         assert(True == self.nodes[0].getblock(self.nodes[0].getbestblockhash())["chainlock"])
         chainlocked_tip = self.nodes[0].getbestblockhash()
 
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -48,6 +48,8 @@ static const float RECONNECT_TIMEOUT_EXP = 1.5;
  * this is belt-and-suspenders sanity limit to prevent memory exhaustion.
  */
 static const int MAX_LINE_LENGTH = 100000;
+/** Default Tor SOCKS port */
+static const int DEFAULT_TOR_SOCKS_PORT = 9050;
 
 /****** Low-level TorControlConnection ********/
 
@@ -386,6 +388,8 @@ private:
 
     /** Callback for ADD_ONION result */
     void add_onion_cb(TorControlConnection& conn, const TorControlReply& reply);
+    /** Callback for GETINFO net/listeners/socks result */
+    void get_socks_cb(TorControlConnection& conn, const TorControlReply& reply);
     /** Callback for AUTHENTICATE result */
     void auth_cb(TorControlConnection& conn, const TorControlReply& reply);
     /** Callback for AUTHCHALLENGE result */
@@ -466,14 +470,12 @@ void TorController::auth_cb(TorControlConnection& _conn, const TorControlReply& 
     if (reply.code == 250) {
         LogPrint("tor", "tor: Authentication successful\n");
 
-        // Now that we know Tor is running setup the proxy for onion addresses
-        // if -onion isn't set to something else.
+        // Now that we know Tor is running, discover its actual SOCKS port and
+        // set up the onion proxy, unless -onion is already set to something else.
+        // Querying the control port avoids hardcoding 9050 (Bitcoin Core pattern).
         if (GetArg("-onion", "") == "") {
-            CService resolved(LookupNumeric("127.0.0.1", 9050));
-            proxyType addrOnion = proxyType(resolved, true);
-            SetProxy(NET_ONION, addrOnion);
-            if (!IsNetworkExplicitlyLimited(NET_ONION))
-                SetLimited(NET_ONION, false);
+            _conn.Command("GETINFO net/listeners/socks",
+                boost::bind(&TorController::get_socks_cb, this, _1, _2));
         }
 
         // Finally - now create the service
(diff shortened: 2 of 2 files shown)


reubenyap:
### Review summary

The PR is the right class of fix and broadly matches Bitcoin Core's pattern: it queries `GETINFO net/listeners/socks` instead of hardcoding `127.0.0.1:9050`, and validates each candidate via `LookupNumeric()` before accepting it (which makes the parsing slightly safer than upstream — non-TCP entries like `unix:/...` are dropped before they can shadow a real TCP listener). The `-onlynet=onion` name-proxy hardening is a reasonable Firo-specific improvement and is correctly guarded by `!HaveNameProxy()`. I don't see any consensus, wallet, or remote-peer attack surface introduced here — the change is local Tor-control/proxy configuration only.

### One robustness issue I'd fix before merge

`_conn.Command("GETINFO net/listeners/socks", ...)` is called without checking the return value. `TorControlConnection::Command()` returns `false` if `b_conn` is null or `bufferevent_get_output()` fails (`src/torcontrol.cpp:237-248`); in that case, `get_socks_cb` is never invoked, and therefore the fallback to `127.0.0.1:9050` inside `get_socks_cb` never runs either. That leaves `NET_ONION` unconfigured for the lifetime of the controller — a small regression vs. the pre-PR behaviour, which configured the fallback synchronously.

Bitcoin Core also ignores this return value, so this isn't a deviation from upstream — but the PR description explicitly states "falls back to `127.0.0.1:9050` if Tor returns nothing usable", and being defensive here costs almost nothing.

### Proposed fix

Pushed to branch [`claude/review-firo-pr-1823-CjIxo`](https://github.com/firoorg/firo/tree/claude/review-firo-pr-1823-CjIxo) (one commit on top of `0318ab8`):

- Extracts the onion-proxy configuration (`SetProxy` / `SetLimited` / conditional `SetNameProxy`) into a private helper `TorController::ConfigureOnionProxy(const CService&)` so the success and failure paths share identical behaviour.
- In `auth_cb`, when `Command()` returns `false`, logs the failure and calls `ConfigureOnionProxy(LookupNumeric("127.0.0.1", DEFAULT_TOR_SOCKS_PORT))` synchronously, restoring the pre-PR fallback guarantee.
- The helper also refuses to configure an invalid `CService`, replacing the previous silent `SetProxy(NET_ONION, /* invalid */)` corner case.

Diff is +21/-2, single file. Feel free to cherry-pick the commit (`f74df11`) onto `tor/discover-socks-port` — happy to open a separate PR if you prefer.

### Minor non-blockers (not addressed)

- `_conn` is unused in `get_socks_cb`. Other callbacks in this file (`add_onion_cb`, `protocolinfo_cb`, etc.) follow the same `_`-prefix-means-unused convention without an explicit `(void)_conn;`, so I left it consistent with the rest of the file.
- The candidate preference inherits Bitcoin's `127.0.0.1:`-prefix bias and doesn't specially prefer IPv6 loopback `[::1]:`. Acceptable as-is for parity with upstream.

### Recommendation

**Approve once the `Command()` failure fallback is in.** Without it the PR is still safe in normal operation, but it has a small robustness regression versus the old synchronous fallback.


---

reubenyap:
### Pushed `f74df11` to `tor/discover-socks-port`

I audited all four CodeRabbit threads and only one was a valid, unaddressed concern. Pushed the fix as a fast-forward on top of `0318ab8`.

| # | Thread | Severity | Status |
|---|--------|----------|--------|
| 1 | `kSocksPrefix` naming + listener validation ([r3132184737](https://github.com/firoorg/firo/pull/1823#discussion_r3132184737)) | Major | Already addressed in `0318ab8` ✅ |
| 2 | Unused-parameter warning on `get_socks_cb` declaration ([r3140775954](https://github.com/firoorg/firo/pull/1823#discussion_r3140775954)) | Minor | **Not valid** — see below |
| 3 | `_conn.Command()` return value ignored ([r3140775960](https://github.com/firoorg/firo/pull/1823#discussion_r3140775960)) | Major | **Fixed in `f74df11`** |
| 4 | `(void)_conn;` cast in `get_socks_cb` body ([r3140775967](https://github.com/firoorg/firo/pull/1823#discussion_r3140775967)) | Minor | **Not valid** — see below |

### Why threads 2 & 4 are not valid

`CMakeLists.txt:703-705` explicitly applies `-Wno-unused-parameter` project-wide:

```cmake
try_append_cxx_flags("-Wunused-parameter" TARGET warn_interface SKIP_LINK
  IF_CHECK_PASSED "-Wno-unused-parameter"
)
```

So an unused `_conn` parameter does not produce a warning, let alone a CI failure. The `_`-prefix convention is used throughout `torcontrol.cpp` precisely to signal "intentionally unused" — `add_onion_cb`, `connected_cb`, and `disconnected_cb` all take a `TorControlConnection& _conn` they never reference, and none of them carry a `(void)_conn;` cast. Adding one only to `get_socks_cb` would be inconsistent with the rest of the file.

### What `f74df11` does (thread 3)

`TorControlConnection::Command()` (`src/torcontrol.cpp:237-248`) returns `false` when `b_conn` is null or `bufferevent_get_output()` fails, and in that case the supplied callback is never invoked. The PR previously ignored that return value, so a queue failure would leave `NET_ONION` unconfigured for the lifetime of the controller — a small regression vs. the pre-PR synchronous fallback.

The commit:

- Extracts the onion-proxy configuration (`SetProxy` / conditional `SetLimited` / `SetNameProxy`) into a private helper `TorController::ConfigureOnionProxy(const CService&)` so success and failure paths share identical behaviour.
- In `auth_cb`, checks the `Command()` return value and on failure logs the error and calls `ConfigureOnionProxy(LookupNumeric("127.0.0.1", DEFAULT_TOR_SOCKS_PORT))` synchronously, restoring the pre-PR fallback guarantee.
- Helper guards against an invalid `CService` so `SetProxy(NET_ONION, /* invalid */)` is no longer silently called in pathological resolution failures.

Diff: +21/-2 in `src/torcontrol.cpp`, single file. CI should pick it up shortly.


---

PR #1865, opened 2026-06-15
Spark: Speed up reindex proof verification

navidR:
This patch speeds up reindex after Spark activation by removing duplicate Spark spend proof verification and making the remaining Grootle proof verification cheaper. During reindex/import, Spark spend proofs no longer need to be fully verified in the early CheckBlock path and then verified again during ConnectBlock; instead, the expensive proof check is deferred to the ConnectBlock path before Spark state is committed, while mempool, wallet, and VerifyDB proof checks remain enforced. The Grootle verifier is also optimized by avoiding a large intermediate commitment list and feeding the equivalent multi-scalar multiplication terms directly into secp256k1. In the 855k-860k Spark-heavy window, reindex time dropped from 1:26:13 to about 13:12, a 6.5x speedup and 84.7% wall-time reduction.

### DIFF
--- a/src/batchproof_container.cpp
+++ b/src/batchproof_container.cpp
@@ -17,9 +17,16 @@ void BatchProofContainer::init() {
     tempSparkTransactions.clear();
 }
 
+void BatchProofContainer::clear() {
+    tempSparkTransactions.clear();
+    sparkTransactions.clear();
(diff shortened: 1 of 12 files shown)


navidR:
@coderabbitai review

PR #1864, opened 2026-06-14
Evo: speed up early reindex merkle checks

navidR:
This patch speeds up early -reindex around the DIP3 deterministic masternode activation window by avoiding redundant cbTx masternode-list merkle-root rebuilds. During ConnectBlock, Firo already builds the deterministic masternode list and diff for the block; the patch reuses that freshly computed list for cbTx merkle validation, and if the diff only changes fields that are not serialized into CSimplifiedMNListEntry, it safely reuses the previous merkle root instead of rebuilding and rehashing the same effective simplified MN list. Consensus behavior is preserved because additions, removals, and every state field that affects the cbTx MN merkle root still force recomputation, while unrelated fields such as payout/last-paid metadata do not. It also gates several high-volume validation trace logs behind -debug=validation, reducing reindex log I/O without removing opt-in diagnostics.

PR #1866, opened 2026-06-17
Speed up DIP3 and Spark reindex validation safely

navidR:
Overall this PR should result in `-reindex` process 2x-3x faster. 

1. This PR speeds up early -reindex around the DIP3 deterministic masternode activation window by avoiding redundant cbTx masternode-list merkle-root work. During block connection, Firo already builds the deterministic masternode list and diff for the block; the patch reuses that freshly computed list for cbTx MN merkle validation, and skips rebuilding the simplified MN-list merkle root when a diff only changes fields that are not serialized into CSimplifiedMNListEntry. Consensus-relevant MN changes still force recomputation. 

2. For Spark-heavy reindex, this PR defers Spark spend proof verification during raw block import and batch-verifies proofs before validation state is persisted, while preserving fail-closed behavior. It also avoids rebuilding unnecessary full cover sets for Grootle batch checks, keeps proof-cache entries tied to exact proof inputs, verifies deferred batches before flush/shutdown/reindex completion, and gates high-volume Spark validation trace logs behind -debug=validation to reduce log I/O during reindex.

First part of this PR already evaluated/approved by @levonpetrosyan93  at https://github.com/firoorg/firo/pull/1864

### DIFF
--- a/src/libspark/grootle.cpp
+++ b/src/libspark/grootle.cpp
@@ -432,13 +432,6 @@ bool Grootle::verify(
         bind_weight.randomize();
     }
 
-    // Bind the commitment lists
-    std::vector<GroupElement> commits;
-    commits.reserve(S.size());
-    for (std::size_t i = 0; i < S.size(); i++) {
-        commits.emplace_back(S[i] + V[i]*bind_weight);
-    }
-
     // Final batch multiscalar multiplication
     Scalar H_scalar;
     std::vector<Scalar> Gi_scalars;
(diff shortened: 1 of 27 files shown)


navidR:
@coderabbitai review

navidR:
@coderabbitai review

justanwar:
Reindex on master 2fade3dfc: 27 hours 17 minutes
Reindex on this branch 84982a0: 9 hours 20 minutes

src/libspark/grootle.cpp
```diff
@@ -432,40 +432,25 @@ bool Grootle::verify(
         bind_weight.randomize();
     }
 
-    // Bind the commitment lists
```
levonpetrosyan93, line 435: Please make a benchmark comparison with the changes in this .cpp versus without those changes. The changes actually look reasonable and simplify the code, but our Multiexponent algorithm complexity is not linear, so we need to understand the impact of this. In case it's not much or even makes the performance worse, please revert the changes, as this is a very critical part.

navidR: Sure. Thanks for your comment. I am running the benchmarks. Will put the results here once finished. It will take few hours though.

PR #1867, opened 2026-07-05
Improve the performance of `firo-qt` on Mac when the UI is out of focus

psolstice:
## PR intention
When main window is out of focus on Mac the OS prioritises the performance of the application. This PR prevents that.

## Code changes brief
Additional code in `src/qt/macnapinhibitor.h/ .mm` prevents nap from happening


navidR:
`share/qt/Info.plist.in:63` already has `LSAppNapIsDisabled = True`, and `cmake/module/Maintenance.cmake:92` copies that plist into the app bundle, the PR duplicates an existing App Nap-disable mechanism without explaining why the plist is insufficient. Maybe our plist configuration does not work correctly?

reubenyap:
@codeant-ai review

navidR (approved):

PR #1872, opened 2026-07-10
Fix Firo Build Warnings

navidR:
## PR intention

This branch focuses on reducing Firo build-warning noise without changing runtime behavior or consensus/wallet semantics. It fixes straightforward project-owned warnings, scopes warning flags more carefully, keeps third- party/vendor warnings out of the main project warning surface, and moves immer into the depends system as a proper CMake-installed dependency instead of carrying it under src/. It also improves the CMake configure output by reporting expected global compile definitions/options as status notices while preserving stricter diagnostics for genuinely problematic global build-system usage.


depends/packages/packages.mk
```diff
@@ -1,4 +1,4 @@
-packages:=boost openssl libevent gmp zlib backtrace tor bls-dash
+packages:=boost openssl libevent gmp zlib backtrace tor bls-dash immer
```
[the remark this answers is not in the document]
navidR: immer is header only and installed to (custom) root defined by toolchain. So it is okay.

reubenyap:
@codeant-ai review

src/crypto/MerkleTreeProof/core.c
```diff
@@ -399,9 +399,11 @@ int validate_inputs(const argon2_context *context) {
         }
     }
 
+#if ARGON2_MIN_PWD_LENGTH > 0
     if (ARGON2_MIN_PWD_LENGTH > context->pwdlen) {
       return ARGON2_PWD_TOO_SHORT;
     }
+#endif
```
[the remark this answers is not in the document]
navidR: Not important. That change not related to PR.

reubenyap:
@codeant-ai review

depends/packages/packages.mk
reubenyap, line 1: **immer has no CMake wiring — it currently resolves by accident**

Re-raising the earlier codex thread, because I think the "header only, installed to the toolchain root, so it is okay" answer explains why it *happens* to work rather than why it is *wired*.

With `src/immer/` deleted, `src/evo/deterministicmns.h:16` still does `#include "immer/map.hpp"`, but there is no `find_package(immer)`, no imported target, and no `target_include_directories` for the depends prefix anywhere in the build. I traced how the header is actually found today:

- `depends/toolchain.cmake.in` sets `CMAKE_FIND_ROOT_PATH` to the depends prefix, but it does **not** put `${prefix}/include` on any compile line — `@CPPFLAGS@` only becomes `DEPENDS_COMPILE_DEFINITIONS` (definitions, not include dirs).
- `cmake/module/AddBoostIfNeeded.cmake:34` runs `find_package(Boost ...)`, which sets `Boost_INCLUDE_DIR` to `<depends>/include`, and `Boost::headers` propagates that as an interface include directory.
- `evo/deterministicmns.cpp` is in `bitcoin_common`, which links Boost — so `-I<depends>/include` arrives transitively, and immer installs into that same prefix.

So the include path is supplied by Boost, not by anything that knows about immer. Consequences:

- Any configuration where Boost does not come from the depends prefix (system Boost + depends for the rest, distro packaging, IDE/system-lib setups) fails with `fatal error: immer/map.hpp: No such file or directory`.
- Nothing declares the dependency, so a future change to Boost linkage silently breaks `evo/`.

immer 0.9.1 installs `immerConfig.cmake`, so this should be a small fix — roughly `find_package(immer CONFIG REQUIRED)` plus linking `immer` into `bitcoin_common`. That also documents the new dependency in the build files rather than leaving it implicit.

For the record, the rest of the packaging is correct: `immer.mk` uses the repo's `$(package)_cmake` macro (`depends/funcs.mk:259`), which already passes `CMAKE_INSTALL_PREFIX`, so the `DESTDIR` staging is right, and the sha256/URL are implicitly validated by the passing Linux/macOS/Windows and guix-darwin/guix-mingw jobs (depends verifies every download, and the cache key includes `hashFiles('depends/packages/*.mk')`).

---

CMakeLists.txt
```diff
@@ -824,7 +829,7 @@ if(WERROR)
   else()
     set(werror_flag "-Werror")
   endif()
-  try_append_cxx_flags(${werror_flag} TARGET core_interface SKIP_LINK RESULT_VAR compiler_supports_werror)
+  try_append_cxx_flags(${werror_flag} TARGET warn_interface SKIP_LINK RESULT_VAR compiler_supports_werror)
```
reubenyap, line 832: **`-Werror` moved here is the right call, but nothing in CI exercises it**

Moving `-Werror` off `core_interface` onto `warn_interface` is correct — it keeps vendored targets out of the error surface now that `core_interface` no longer pulls in `warn_interface`.

The gap is that no cmake invocation in `.github/workflows/ci-master.yml` passes `-DWERROR=ON` (the Linux/macOS/Windows jobs set only `-DBUILD_CLI`, `-DBUILD_TESTS`, `-DENABLE_CRASH_HOOKS`, `-DBUILD_GUI`). So the central claim of this PR — that the project-owned build is warning-clean — is not enforced anywhere, and any regression back to noisy warnings will land unnoticed.

Since the cleanup is done, this seems like the moment to lock it in: adding `-DWERROR=ON` to at least the `linux-cmake-Debug` job would make the invariant self-maintaining. Without it, this PR's value decays with the next few merges.

---

src/crypto/CMakeLists.txt
```diff
@@ -35,13 +35,14 @@ add_library(bitcoin_crypto STATIC EXCLUDE_FROM_ALL
 target_link_libraries(bitcoin_crypto
   PRIVATE
     core_interface
+    warn_interface
```
reubenyap, line 38: **This pulls vendored crypto *into* the project warning surface, against the PR's stated goal**

The PR description says it "keeps third-party/vendor warnings out of the main project warning surface", but `bitcoin_crypto` compiles a lot of vendored code alongside the first-party sources — `MerkleTreeProof/core.c`, `ref.c`, `thread.c`, `blake2/blake2b.c`, and all of `Lyra2Z/` — and linking `warn_interface` here subjects all of it to the project warning set.

That is precisely why this PR had to patch upstream argon2 in `src/crypto/MerkleTreeProof/core.c` with `#if ARGON2_MIN_PWD_LENGTH > 0` / `#if UINTPTR_MAX <= UINT32_MAX` guards. (Those guards are correct — `ARGON2_MIN_{PWD_LENGTH,SECRET,AD_LENGTH}` are all `0`, and `ARGON2_MAX_MEMORY` is `0xFFFFFFFF` on 64-bit against a `uint32_t m_cost` but `1<<21` on 32-bit where the check is load-bearing and correctly retained — so no behavior change. The issue is that we now carry local edits to vendored code purely to satisfy our warning flags.)

Combined with turning on `-DWERROR=ON` in CI, this also means future vendored-code warnings become hard build failures in a directory we do not own.

Suggest either splitting the first-party `.cpp`s from the vendored C sources into separate targets, or leaving `warn_interface` off and using `set_source_files_properties(<vendored sources> PROPERTIES COMPILE_OPTIONS -w)` — either way the argon2 edits could then be dropped.

---

src/secp256k1/cmake/TryAppendCFlags.cmake
```diff
@@ -19,6 +19,8 @@ endfunction()
 macro(try_append_c_flags)
   secp256k1_check_c_flags_internal("${ARGV}" result)
   if(result)
-    add_compile_options(${ARGV})
+    foreach(flag ${ARGV})
```
reubenyap, line 22: **This edits a subtree file — it will conflict with or be reverted by the next secp256k1 update**

The change itself is fine (wrapping each flag in `$<$<COMPILE_LANGUAGE:C>:...>` so C-only flags stop leaking onto C++ TUs), but `src/secp256k1/` is a vendored subtree. Anyone running `git subtree pull` for secp256k1 will either hit a conflict here or silently drop the fix, and the resulting warning regression will be invisible.

Worth a maintainer call on which of these is preferred:
- keep it as a documented patch (e.g. under `depends/patches/` or a note in the subtree docs) so it is reapplied deliberately, or
- achieve the same scoping from the parent scope rather than by editing the subtree.

Not blocking, but "silently reverted on the next dependency bump" is the kind of thing worth deciding now rather than rediscovering later.

---

cmake/module/WarnAboutGlobalProperties.cmake
```diff
@@ -8,14 +8,14 @@ include_guard(GLOBAL)
 # Instead, prefer the target-specific target_compile_definitions() one.
 get_directory_property(global_compile_definitions COMPILE_DEFINITIONS)
 if(global_compile_definitions)
-  message(AUTHOR_WARNING "The directory's COMPILE_DEFINITIONS property is not empty: ${global_compile_definitions}")
+  message(STATUS "Notice: directory-level COMPILE_DEFINITIONS are set: ${global_compile_definitions}")
```
reubenyap, line 11: **Downgrading to `STATUS` silences the module's entire purpose, not just the known offenders**

This module exists to catch directory-wide `add_compile_definitions()` / `add_compile_options()` creeping in where a `target_*` call belongs. Turning `AUTHOR_WARNING` into `STATUS` makes the current output clean, but it also means genuinely bad *new* global usage will never be flagged again — the check becomes decorative.

The actual offenders are few and identifiable: `CMakeLists.txt:933` (`include_directories(${CMAKE_SOURCE_DIR}/src)`) plus the `add_compile_definitions()` calls (e.g. `HAVE_WORKING_BOOST_SLEEP_FOR` in `cmake/module/AddBoostIfNeeded.cmake`, `CRASH_HOOKS_WRAPPED_CXX_ABI`).

Two alternatives that keep the guard useful:
- fix the handful of offenders so the warning is genuinely silent, or
- keep `AUTHOR_WARNING` but compare against an explicit allowlist of the known-accepted definitions, so only *unexpected* additions warn.

Either preserves the tripwire; the current change removes it.

---

src/rpc/server.cpp
```diff
@@ -573,8 +573,6 @@ UniValue CRPCTable::execute(const JSONRPCRequest &request) const
     {
         throw JSONRPCError(RPC_MISC_ERROR, e.what());
     }
-
-    g_rpcSignals.PostCommand(*pcmd);
```
reubenyap, line 577: **Removal is behavior-neutral — confirmed — but consider removing the now-dead public hook too**

I verified this is genuinely unreachable: every path inside the `try` block `return`s (both the `isObject()` and positional branches), and the `catch` re-throws, so control never reaches this line. Nothing in the tree calls `RPCServer::OnPostCommand`, so nothing observes the difference. No behavior change, and it correctly clears the `-Wunreachable-code` hit.

The thing left behind is that `RPCServer::OnPostCommand()` (`src/rpc/server.h:31`) and the `PostCommand` signal (`src/rpc/server.cpp:45`, `:63-65`) remain as public API that can now never fire under any circumstances. Any caller that registers a slot gets silence.

Two clean options — either is fine, but the current state is the worst of the three:
- delete `PostCommand` / `OnPostCommand` along with the call site, or
- make it actually fire (capture the result and emit before returning, or an RAII guard), if the hook is intended to be usable.

---

src/prevector.h
```diff
@@ -427,7 +427,20 @@ class prevector {
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#pragma GCC diagnostic ignored "-Wuninitialized"
+#endif
+        // GCC can report the inactive bytes of direct_or_indirect as maybe
+        // uninitialized when this swap is inlined. This is intentional: _size
+        // is the discriminator for the active representation, and the union is
+        // swapped as raw storage so direct prevector buffers stay uninitialized
+        // until elements are explicitly constructed or copied into them.
         std::swap(_union, other._union);
```
reubenyap, line 440: **Worth confirming this pragma actually suppresses the warning**

The reasoning in the comment is right — `_size` is the discriminator and the union is deliberately swapped as raw storage — so suppressing rather than "fixing" is the correct call here.

One caveat: GCC frequently attributes `-Wmaybe-uninitialized` arising from *inlining* to the caller's source location rather than the inlined callee's, and `#pragma GCC diagnostic` keys off the location the diagnostic is reported at. Since the whole point of this suppression is the inlined case ("when this swap is inlined"), the pragma may not actually take effect at the sites that warn.

Could you confirm the warning is gone in a build that previously produced it (rather than one where it did not fire)? If it survives, the usual workaround is suppressing at the caller or via `-Wno-...` scoped to the affected target.

Minor secondary point: `-Wuninitialized` in addition to `-Wmaybe-uninitialized` is broader than the described false positive needs. The push/pop keeps the blast radius to this one statement, so it is low risk, but dropping the unconditional one would keep real defects reportable here.

---

src/test/serialize_tests.cpp
```diff
@@ -351,18 +351,18 @@ BOOST_AUTO_TEST_CASE(class_methods)
     CSerializeMethodsTestSingle methodtest3;
     CSerializeMethodsTestMany methodtest4;
     CDataStream ss(SER_DISK, PROTOCOL_VERSION);
-    BOOST_CHECK(methodtest1 == methodtest2);
+    BOOST_CHECK(methodtest1.operator==(methodtest2));
```
reubenyap, line 354: **Nit: fixing the operator would be cleaner than calling it explicitly at five call sites**

`a.operator==(b)` works around the C++20 reversed-candidate ambiguity, but it reads oddly in a test and has to be repeated at every call site (5 here), so the next person adding a comparison will hit the same error and have to rediscover the workaround.

Making `CSerializeMethodsTestSingle::operator==` (and the `Many` variant) a `const` member taking `const&` should resolve the ambiguity at the source and let all of these go back to plain `BOOST_CHECK(a == b)`.

If there is a reason to keep the explicit form, a one-line comment explaining *why* would save the next reader the detour.

---

reubenyap (commented):
Reviewed the full diff (111 files, 26 commits) against the codebase, plus CI and the existing review threads. I did not build locally — a depends build is hours — but CI built this on Linux/macOS/Windows (Release + Debug) and guix darwin/mingw.

**I found no correctness bugs in the diff.** Every behavioral change I traced is either neutral or a genuine fix. The items below are about durability and build-system structure, not broken code.

## Worth resolving before merge

1. **immer has no CMake wiring** (`depends/packages/packages.mk`) — the vendored tree is gone but the include path is supplied incidentally by Boost's interface include directory. Details inline; this is the one structural item.
2. **CI never sets `-DWERROR=ON`** (`CMakeLists.txt:832`) — the PR's central claim is unenforced and will decay.
3. **Vendored crypto is now inside the project warning surface** (`src/crypto/CMakeLists.txt:38`) — this contradicts the stated goal and is why upstream argon2 had to be patched.
4. **CI is red, but it is infrastructure, not code** — `guix-x86_64-linux-gnu` and `guix-aarch64-linux-gnu` both failed at ~60s with `HTTP 502 Bad Gateway` fetching `guix-install.sh` from savannah.gnu.org. Needs a re-run for a green board.

## Verified correct — recording these so reviewers do not re-litigate them

- **`stacktraces.cpp` `#if __clang__` → `#if defined(__APPLE__)` is a real bug fix.** `CMakeLists.txt:194` wraps `__assert_fail` on all non-Windows hosts, so a Linux+clang build previously compiled `__wrap___assert_rtn` while the linker was told to wrap `__assert_fail` — the assertion crash hook was dead there. Now consistent.
- **Removing `RAND_screen()` / `RAND_event()` is not a security regression.** I checked this deliberately because `GetRandBytes` still routes through `RAND_bytes` (`src/random.cpp:127`). depends ships OpenSSL 1.1.1m, where both are `DEPRECATEDIN_1_1_0` wrappers around `RAND_poll()`; the 1.1.1 DRBG auto-seeds from the OS CSPRNG, and `RandAddSeed()` / `RandAddSeedPerfmon()` still feed `RAND_add`. Also a prerequisite for any OpenSSL 3.x move, where they are gone entirely.
- **The `.ui` renames are safe.** `addressbookpage.ui` had a duplicate `horizontalLayout`, `recover.ui` duplicated `horizontalLayout` and `horizontalSpacer`, `sendcoinsentry.ui` duplicated `horizontalLayoutWarning`. All duplicates are resolved on this head, and no `.cpp`/`.h` references any renamed name, so `ui->...` access is unaffected.
- **Both init-list reorders match declaration order exactly** — `walletmodel.h:286-301` and `validation.h:488-495`.
- **The `mtp_tests` swap is sound.** The removed `block1.nNonce != block3.nNonce` assertion was genuinely flaky (~1 in 500 at `pow_limit = 00ffff…`, since the two nonces are near-independent geometric draws). The replacement shares `block1.mtpHashData` via `shared_ptr`, but `mtp::verify` takes the header by const ref and only *reads* `hashRootMTP` / `nBlockMTP` / `nProofMTP`, so it cannot corrupt the later `verify(block1.nNonce, block1, …)` assertion.
- **Also confirmed:** `try_append_cxx_flags` does support `VAR` with `PARENT_SCOPE`, so the `-Woverloaded-virtual` / `-Wsuggest-override` CXX scoping works; all 21 first-party targets link `warn_interface` and only genuinely vendored targets are excluded; `OSX_MIN_VERSION` is set at `depends/hosts/darwin.mk:1` and included before `builders/darwin.mk` (`depends/Makefile:115` vs `:117`); `memcpy` + `strnlen` matches `strncpy` semantics given the preceding `memset`; `[=, this]` is the required C++20 fix for the deprecated implicit `this` capture; `CCriticalBlock` / `SparkModel::cs` types line up and `reset()` is strictly safer than the old `if (sparkModel)` guard (recursive mutex, so a repeat `setModel()` is still safe); the `MintAndStoreSpark` doc block now matches the real signature; the added `PRIVATE warn_interface` entries are all on calls that already used the keyword signature, so no plain/keyword mixing.

## On the existing bot findings

The three CodeAnt items marked Critical/Major are stale or describe pre-existing behavior:

- the `sparkname_tests` lambda captures **are** fixed on this head (`[overflowBlocks]` at lines 798 and 807) — those threads are outdated;
- the 32-bit Windows `skip_frames` path preserves master's behavior exactly (the `if (!pContext) skip++` was already only in the `__x86_64__` branch), so it is not a regression introduced here;
- the Qt 5.1.0 high-DPI guard (`> 0x050100`) is unchanged from master, and this project requires Qt 6.7.3+ regardless.

The `stacktraces.cpp` truncation point CodeAnt raises is real but pre-existing: `readlink` is called with `bufSize - 1`, so a return of `bufSize - 1` means truncation yet still satisfies `len < buf.size()` and never grows the buffer. This PR only added the `static_cast<size_t>`; worth a separate fix, not a blocker here.

## Minor

Commit prefixes do not follow the convention in `CLAUDE.md` (`Build:`, `Tests:`, `Qt:`, …) and the 26 commits use mixed styles — worth normalizing or squashing before merge.

Nice work on the crash-hook and argon2 findings in particular — those were real latent issues surfaced by taking the warnings seriously.

---

depends/packages/packages.mk
navidR: We don’t need this. Immer is header only, so no need to wire cmake.

src/rpc/server.cpp
navidR: Out of scope for this PR

src/prevector.h
navidR: Yes. Comment explains why.

src/test/serialize_tests.cpp
navidR: Out of scope for this PR.

cmake/module/WarnAboutGlobalProperties.cmake
navidR: This was correct call. Better to silence and change it to STATUS.

PR #1885, opened 2026-07-27
Wallet: Avoid O(N²) wallet scans for Spark outputs to fix slow Firo-QT startup

reubenyap:
## PR intention

Reduce Firo-QT startup time and fix the long freeze at "Starting network threads..." before the main window appears.

The splash message is misleading: "Starting network threads..." is the last message painted before init finishes and `initializeResult` constructs `ClientModel`/`WalletModel` synchronously on the GUI thread. `TransactionTablePriv::refreshWallet` then decomposes every wallet transaction under `LOCK2(cs_main, cs_wallet)`, and since the splash spinner is a self-rearming 50 ms timer on that same thread, the splash freezes until model construction completes.

The reason model construction is so slow: `CWallet::IsMine(const CTxOut&)` and `CWallet::GetCredit(const CTxOut&, filter)` recover a Spark output's serial context by linearly scanning **every output of every transaction in `mapWallet`** before doing the trial decryption. They are called per output per transaction during model construction, making it O(N²) in wallet size — minutes of GUI-thread work for wallets with thousands of Spark transactions. The same scans run during `AddToWallet` (rescan/sync), `LoadWallet` (`setWalletUTXO` rebuild at startup), balance computation, `AvailableCoins`, and the coin control dialog.

## Code changes brief

The Spark serial context depends only on the transaction containing the output (`spark::getSerialContext(tx)`), which every hot caller already has in hand — the wallet-wide scan exists only because the helpers receive a bare `CTxOut`.

- Add parent-aware overloads `CWallet::IsMine(const CTxOut&, const CTransaction&)` and `CWallet::GetCredit(const CTxOut&, const CTransaction&, const isminefilter&)` that derive the serial context directly from the containing transaction — the exact pattern `IsMine(const CTransaction&)` already used.
- The old bare-`CTxOut` signatures remain as a fallback: they now locate the containing transaction in the wallet and delegate, so results are unchanged for wallet-known transactions (the only case the scan could ever answer).
- Switch all callers that know the containing transaction: GUI transaction decomposition/description, tx-level `GetCredit`, `GetAmounts`, available/watch-only credit, `GetLegacyBalance`, `AvailableCoins`, address balances/groupings, `AddToWallet`, `LoadWallet`, `AutoLockMasternodeCollaterals`, `listCoins`, and the coin control dialog (`GetJMintCredit` now takes the parent transaction).
- Call sites without the parent transaction in hand (prevout lookups, coin-view debug output) intentionally keep the fallback.

Net effect: O(N²) → O(N) for wallet-wide iteration paths, which removes the frozen splash/late-appearing window and also speeds up balance polling, coin listing, and rescans. No consensus or RPC semantics change.

https://claude.ai/code/session_01PRhMvsTAUDw51jHJV5pTyZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRhMvsTAUDw51jHJV5pTyZ)_

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -510,7 +510,7 @@ void CoinControlDialog::updateLabels(WalletModel *model, QDialog* dialog, bool a
 
         // Amount
         if(out.tx->tx->vout[out.i].scriptPubKey.IsSparkSMint()) {
-            nAmount += model->GetJMintCredit(out.tx->tx->vout[out.i]);
+            nAmount += model->GetJMintCredit(out.tx->tx->vout[out.i], *out.tx->tx);
         } else {
             nAmount += out.tx->tx->vout[out.i].nValue;
         }
@@ -734,7 +734,7 @@ void CoinControlDialog::updateView()
         BOOST_FOREACH(const COutput& out, coins.second) {
             CAmount amount;
             if(out.tx->tx->vout[out.i].scriptPubKey.IsSparkSMint()) {
-                amount = model->GetJMintCredit(out.tx->tx->vout[out.i]);
+                amount = model->GetJMintCredit(out.tx->tx->vout[out.i], *out.tx->tx);
             } else {
                 amount = out.tx->tx->vout[out.i].nValue;
             }
--- a/src/qt/transactiondesc.cpp
+++ b/src/qt/transactiondesc.cpp
@@ -169,7 +169,7 @@ QString TransactionDesc::toHTML(CWallet *wallet, CWalletTx &wtx, TransactionReco
         //
         CAmount nUnmatured = 0;
         BOOST_FOREACH(const CTxOut& txout, wtx.tx->vout)
-            nUnmatured += wallet->GetCredit(txout, ISMINE_ALL);
+            nUnmatured += wallet->GetCredit(txout, *wtx.tx, ISMINE_ALL);
         strHTML += "<b>" + tr("Credit") + ":</b> ";
         if (wtx.IsInMainChain())
             strHTML += BitcoinUnits::formatHtmlWithUnit(unit, nUnmatured)+ " (" + tr("matures in %n more block(s)", "", wtx.GetBlocksToMaturity()) + ")";
@@ -196,7 +196,7 @@ QString TransactionDesc::toHTML(CWallet *wallet, CWalletTx &wtx, TransactionReco
         isminetype fAllToMe = ISMINE_SPENDABLE;
         BOOST_FOREACH(const CTxOut& txout, wtx.tx->vout)
         {
-            isminetype mine = wallet->IsMine(txout);
+            isminetype mine = wallet->IsMine(txout, *wtx.tx);
             if(fAllToMe > mine) fAllToMe = mine;
         }
 
@@ -211,7 +211,7 @@ QString TransactionDesc::toHTML(CWallet *wallet, CWalletTx &wtx, TransactionReco
             BOOST_FOREACH(const CTxOut& txout, wtx.tx->vout)
             {
                 // Ignore change
-                isminetype toSelf = wallet->IsMine(txout);
+                isminetype toSelf = wallet->IsMine(txout, *wtx.tx);
                 if ((toSelf == ISMINE_SPENDABLE) && (fAllFromMe == ISMINE_SPENDABLE))
                     continue;
                 CSparkOutputTx sparkOutput;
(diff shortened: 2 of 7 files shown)


reubenyap:
@codeant-ai review

reubenyap:
## Safety analysis — why this change cannot miss Spark coins or alter wallet state

For reviewers, here is the reasoning for why the parent-aware overloads are behavior-preserving, structured around what the old code actually computed.

### Equivalence by construction

`spark::getSerialContext(tx)` ([`spark/state.cpp` L1102](https://github.com/firoorg/firo/blob/master/src/spark/state.cpp#L1102)) is a **pure function of the transaction**: for Spark spends it serializes the spend's used lTags; otherwise it serializes the tx inputs with scriptSigs cleared. It reads no wallet state.

The removed `mapWallet` scans had exactly one purpose: find the transaction containing the `CTxOut` so this value could be computed from it. Every switched call site passes the literal containing transaction — the `txout` is drawn from that same transaction's `vout` in the same scope. Same input to the same pure function → same serial context → identical `sparkWallet->isMine()` / `getMyCoinV()` / `getMintMeta()` calls. This is also not a novel pattern: `CWallet::IsMine(const CTransaction&)` and the GUI's `RecvSpark` path already derived the context directly from the containing transaction.

The bare-`CTxOut` signatures still exist and now locate the containing transaction and delegate, so any caller not switched (prevout lookups, the coin-view debug output in `transactiondesc.cpp`) behaves as before.

### The only cases where results could differ

1. **Byte-identical output present in two different wallet transactions.** The old scan used whichever transaction map iteration found *first* — potentially the wrong parent, producing a wrong serial context and a **false `ISMINE_NO`** for the true coin. The new code always uses the actual parent. Spark coins embed per-coin randomness, so identical serializations across transactions are cryptographically negligible — but in the adversarial replay case, the new code is the correct one; the old scan was the risk.
2. **Transaction not in `mapWallet`.** Old: empty context → `ISMINE_NO` / 0. New: correct answer. This cannot trigger at the switched sites (all of them iterate `mapWallet` or operate on `CWalletTx` objects; `AddToWallet` inserts before its loop), and where it hypothetically could, "correct answer" is strictly safer than "always no".
3. **The `IsSpent` simplification.** The guard `if (tx && tx->tx->vout.size() > n)` guarantees `GetWalletTx(hash)` returned a wallet transaction, so the removed scan — which compared `tx->tx->vout[n]` against every wallet output *including its own* — always rediscovered that same transaction. Deriving the context from `*tx->tx` directly is equivalent by construction, not just in practice.

### What is untouched

- No consensus, validation, or Spark state code changes; no wallet DB write paths change. The values written via `setWalletUTXO` / `LockCoin` in `AddToWallet`/`LoadWallet` are computed from `IsMine`/`IsSpent` results that are identical per the above.
- Locking is unchanged: the parent-aware `IsMine` takes `cs_wallet` exactly like the bare version; the one added `LOCK(cs_wallet)` (around the pre-existing unguarded fallback scan in `GetCredit`) is recursive and respects the canonical `cs_main → cs_wallet` order.
- Filter semantics (`ISMINE_SPENDABLE` gate for smint credit, watch-only short-circuits) are byte-for-byte the same as before.

### Contract

The new overloads carry a documented precondition (`@pre txout is an output of tx`). All current callers satisfy it trivially since the output is taken from `tx.vout` at the call site; a mismatched pair would yield `ISMINE_NO`, never a false positive, and the bare fallback remains for callers without the parent in hand.

### Suggested independent verification

Cheapest real-world check: run an old and a new build against a copy of a large Spark wallet and diff `getwalletinfo`, `listunspentsparkmints`, and `listtransactions "*" 100000` — output should be identical while startup time drops substantially.

---

reubenyap:
## CI triage — both failures are unrelated to the change; re-run needed

Two jobs failed on `faafca3`; neither is caused by this PR, and I lack permission to re-run them, so a maintainer clicking **"Re-run failed jobs"** on [run 30261897926](https://github.com/firoorg/firo/actions/runs/30261897926) would be appreciated.

**1. `guix-x86_64-w64-mingw32` (push-event run) — infrastructure.** The job died in its first 3 minutes fetching the Guix installer: `cgit.git.savannah.gnu.org` answered `502 Bad Gateway`, before the source tree was touched. The identical failure broke master's run on the base commit `a539656` on 2026-07-23 ([run 30034952913](https://github.com/firoorg/firo/actions/runs/30034952913)).

**2. `linux-cmake-Release` (PR-event run) — flaky test, not a defect.** `spark_tests` segfaulted in `connect_and_disconnect_block`. Evidence this is nondeterministic rather than caused by the diff:

- The **same job on the same commit** in the parallel push-event run **passed** ([run 30261895059](https://github.com/firoorg/firo/actions/runs/30261895059)), as did `linux-cmake-Debug` in both runs.
- The full suite including `spark_tests` also **passed on `01ce079`**, the commit containing all semantically significant changes. The `faafca3` delta is only: the `IsSpent` scan removal (equivalent by construction — the guard already holds the fetched wallet transaction), one added *recursive* `LOCK(cs_wallet)` around a pre-existing unguarded traversal, `++i`, brace placement, and comments — none of which can introduce an invalid memory access.
- The test is explicitly timing-dependent: it hard-codes a 1-second sleep annotated "wait while another thread updates mint status in wallet" and mines 500 blocks per case with background wallet-sync threads. This change substantially shortens `cs_wallet` critical sections, which shifts thread interleavings and may make that latent race easier to hit — worth a separate look at the test's synchronization, independent of this PR.

Current status otherwise: Windows Release/Debug, macOS Release/Debug, and `linux-cmake-Debug` are green in both runs; remaining Guix jobs are still running.

---

levonpetrosyan93 (approved):
Looks okay,  changes don't include real optimizations, but are improving the coding  style, so good to have.

PR #1886, opened 2026-07-27
Wallet: Cache Spark coin lookups and show wallet load progress

reubenyap:
## PR intention

Follow-up to #1885, targeting the next startup bottleneck: the long, opaque stall at "Loading wallet...".

Two costs dominate that phase for wallets with many Spark coins:

1. `CSparkWallet::isMine`, `getMyCoinV`, `getMintAmount`, `getMyCoinIsChange`, `getMyCoinAddress`, and `getMintMeta(coin)` run a full EC trial decryption (`coin.identify`) on **every call**, even though every wallet-known coin is already recorded in the in-memory `coinMeta` map (loaded from the wallet DB at startup). These are called per output during wallet load, balance computation, and coin listing.
2. `getMintMeta(nonce)` **linearly scans** `coinMeta` comparing scalars. Since `IsSpent` goes through it per Spark output, wallet-wide loops were still O(outputs × coins).

On top of that, the splash message never updates during the whole phase, so users see a frozen "Loading wallet..." with no sign of life.

## Code changes brief

**Two lookup indexes into `coinMeta`** (values are its `lTagHash` keys), guarded by the existing `cs_spark_wallet` and maintained at every `coinMeta` mutation site — constructor bulk load, `addOrUpdateMint`, `eraseMint`, `clearAllMints`, `updateMintInMemory`; all other mutation paths funnel through these:

- **`coinLookup`: coin hash → lTagHash.** Consulted before trial decryption. A hit requires **full coin equality plus an equal serial context** (coin equality alone does not cover the serial context, which identification depends on). Since coins only enter `coinMeta` after a successful `identify()` at recording time, and `identify()` is deterministic in (coin, serial context, view key), a hit can only return exactly what the cryptography would — the EC path remains as fallback for unknown coins, so a cache miss degrades to current behavior, never to a wrong answer.
- **`nonceLookup`: nonce hash → lTagHash**, replacing the linear scan in `getMintMeta(nonce)`, with a nonce equality check on the hit to guard against hash collisions.

This is the same pattern the wallet already uses for spend-side ownership: `isMine(lTags)` has always been a pure `coinMeta` lookup, and all balances (`getAvailableBalance`, coin selection, `listunspentsparkmints`) already trust `coinMeta` as the source of truth.

**Progress reporting** while the wallet loads, using the splash screen's existing `InitMessage` support: a transaction count is emitted every 1000 tx records read from the wallet database ("Loading wallet... (N transactions)"), and a message marks the Spark wallet loading stage.

**Cache verification hardening** (added during review): answering these queries from recorded metadata shifts trust from the cryptography to wallet.dat integrity, so verification is restored without restoring its cost:

- A **background sweep** posted from the constructor re-runs `identify()` on every indexed record (unspent records first, early-out on shutdown) and evicts the fast-path entries of any record the cryptography does not confirm, with a log line per eviction. Queries on evicted records fall back to full identification; the record itself is untouched.
- **`-sparkcacheverify`** (default off) additionally cross-checks every index hit against identification at query time and rejects divergent hits.
- **`validateLookupIndexes()`** checks the structural invariant between `coinMeta` and the indexes.

**Unit tests** (`spark_tests.cpp`): `wallet_lookup_indexes` registers a mint meta built from foreign keys — trial decryption can never identify it, so every positive answer must come from the indexes, making maintenance bugs visible instead of masked by the EC fallback — and covers add/hit (by coin and nonce), the serial-context mismatch guard, in-memory update, erase, and clear, asserting the structural invariant after every mutation. `wallet_cache_verification` covers the sweep: the foreign-key record is evicted while a wallet-own record survives, and only the fast path is dropped.

No database format changes — the indexes are in-memory and rebuilt from `coinMeta` at construction. No consensus, RPC, or spend-building changes.

Combined with #1885 this removes all remaining per-coin EC work and quadratic scans from the startup path; "Loading wallet..." should collapse to roughly DB I/O time and visibly count up while it runs.

https://claude.ai/code/session_01PRhMvsTAUDw51jHJV5pTyZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRhMvsTAUDw51jHJV5pTyZ)_

### DIFF
--- a/src/spark/sparkwallet.cpp
+++ b/src/spark/sparkwallet.cpp
@@ -406,7 +406,7 @@ void CSparkWallet::eraseMint(const uint256& hash, CWalletDB& walletdb) {
     walletdb.EraseSparkMint(hash);
     auto it = coinMeta.find(hash);
     if (it != coinMeta.end()) {
-        removeFromLookups(it->second);
+        removeFromLookups(hash, it->second);
         coinMeta.erase(it);
     }
 }
@@ -420,7 +420,7 @@ void CSparkWallet::addOrUpdateMint(const CSparkMintMeta& mint, const uint256& lT
     }
     auto it = coinMeta.find(lTagHash);
     if (it != coinMeta.end())
-        removeFromLookups(it->second);
+        removeFromLookups(lTagHash, it->second);
     coinMeta[lTagHash] = mint;
     addToLookups(lTagHash, mint);
     walletdb.WriteSparkMint(lTagHash, mint);
@@ -451,7 +451,7 @@ void CSparkWallet::updateMintInMemory(const CSparkMintMeta& mint) {
     LOCK(cs_spark_wallet);
     for (auto& itr : coinMeta) {
         if (itr.second == mint) {
-            removeFromLookups(itr.second);
+            removeFromLookups(itr.first, itr.second);
             coinMeta[itr.first] = mint;
             addToLookups(itr.first, mint);
             break;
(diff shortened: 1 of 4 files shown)


reubenyap:
CI note: `guix-x86_64-w64-mingw32` failed during Guix bootstrap — `guix pull` hit a truncated substitute download from the GNU servers (`chunked-input-error-prematurely`, "corrupt input while restoring archive"), before the source tree was touched. Same GNU-infrastructure flakiness as on #1885 today (Savannah 502s there). Not caused by this diff; a maintainer clicking **"Re-run failed jobs"** on [run 30283122703](https://github.com/firoorg/firo/actions/runs/30283122703) once the rest of the matrix finishes will clear it — I don't have re-run permission.

Everything else on `01c9172` so far: macOS Release/Debug green, Linux/Windows and the other four Guix builds still running.

---

reubenyap:
@codeant-ai review

levonpetrosyan93 (approved):
Looks good to me, will be useful for wallet containing big amount of spark coins,

PR #1889, opened 2026-07-29
Wallet: Fix deadlock between walletpassphrase and the lockwallet relock timer

### LINKED ISSUE #1888: Node-wide deadlock between walletpassphrase and the lockwallet relock timer
## Summary

A headless `firod` with an encrypted wallet can deadlock completely and silently. Every thread stops, nothing is logged, the RPC port stops answering, and `SIGTERM` never completes — only the outer P2P threads exit, so the process has to be `SIGKILL`ed.

The cycle is between `walletpassphrase` and the `lockwallet` relock timer it schedules.

Reproduced on **v0.14.16.1**; the code is unchanged on **master** (`dde8fcb`).

## The deadlock

**Thread A — an RPC worker running `walletpassphrase`:**

```
walletpassphrase                 src/wallet/rpcwallet.cpp:3063
 -> RPCRunLater                  src/rpc/server.cpp:623      deadlineTimers.erase(name)
 -> ~HTTPRPCTimer                src/httprpc.cpp:34
 -> ~HTTPEvent                   src/httpserver.cpp:533
 -> event_free / event_del_nolock_
 -> evthread_posix_cond_wait                                 <-- blocked
```

`walletpassphrase` takes `LOCK2(cs_main, pwallet->cs_wallet)` at function scope and still holds both when it calls `RPCRunLater` at the end. `RPCRunLater` does `deadlineTimers.erase(name)`, which destroys the *previously* scheduled `lockwallet(...)` timer. That destructor reaches `event_del`, and libevent blocks there until the event's callback has finished executing.

**Thread B — the libevent HTTP loop, executing exactly that callback:**

```
ThreadHTTP                       src/httpserver.cpp:312
 -> event_base_loop / event_process_active_single_queue
 -> httpevent_callback_fn        src/httpserver.cpp:520
 -> LockWallet                   src/wallet/rpcwallet.cpp:3000   LOCK(pWallet->cs_wallet)
 -> __lll_lock_wait                                              <-- blocked
```

`LockWallet` needs `cs_wallet`, which Thread A holds. Thread A waits for Thread B's callback to finish. Neither can proceed, and neither has a timeout.

## Why the whole node stops

Thread A also holds **`cs_main`**, so everything that needs it piles up behind it. From the same dump:

| thread | blocked in |
|---|---|
| `firo-msghand` | `src/net_processing.cpp:1365` |
| `firo-scheduler` | `src/llmq/quorums_chainlocks.cpp:217` (`CChainLocksHandler::CheckActiveState`) |
| `firo-opencon` | `src/net.cpp` |

No thread is in `ConnectBlock`, `ActivateBestChain` or `AcceptToMemoryPool`, and the main thread idles normally in `WaitForShutdown` (`src/bitcoind.cpp:51`) — so the RPC worker is the only lock holder and the attribution is unambiguous.

This also explains the un-killable shutdown: the shutdown path needs the same locks.

## Trigger

A `walletpassphrase` call that arrives while the relock timer scheduled by the *previous* call is executing. It is a genuine race, so the intervals are irregular — in one environment that unlocks the wallet regularly over RPC, it fired anywhere from 40 minutes to 37 hours apart, several times per week.

## How Bitcoin Core handles this

`src/wallet/rpc/encrypt.cpp` upstream:

1. `cs_wallet` is taken in its own scope block and released *before* the relock is scheduled.
2. The relock runs on the scheduler rather than the libevent RPC timer. The comment there describes this exact hazard: *"Scheduled events don't get destructed until they are executed, and they are executed in series in a single scheduler thread so no cs_wallet lock is needed."*
3. A generation check (`if (nRelockTime != relock_time) return;`) replaces destroying the pending timer.
4. `m_unlock_mutex` serialises concurrent `walletpassphrase` calls.

## Proposed fix

A minimal version of (1) + (3) that keeps the existing `RPCRunLater` mechanism: release the lock before scheduling, and pass the expected `nRelockTime` into `LockWallet` so a stale callback does nothing. PR follows.

Happy to test a different approach if you would rather move the relock onto the scheduler as Core did.

### LINKED ISSUE #18811: wallet_descriptor intermittent error "Please enter the wallet passphrase with walletpassphrase first. (-13)"
https://travis-ci.org/github/MarcoFalke/btc_nightly/jobs/680655419#L412

Danswar:
Fixes #1888.

## PR intention

`walletpassphrase` holds `LOCK2(cs_main, pwallet->cs_wallet)` across its call to `RPCRunLater`. `RPCRunLater` erases the previously scheduled `lockwallet(...)` timer, and destroying a libevent timer blocks until that timer's callback has finished running. When `LockWallet` is the callback currently executing on the HTTP event loop, it is itself blocked acquiring `cs_wallet` — so the RPC worker and the event loop deadlock on each other.

Since the RPC worker also holds `cs_main`, message handling, the scheduler and connection handling all stall behind it and the node stops entirely. Nothing is logged, and `SIGTERM` does not complete because the shutdown path needs the same locks. Full thread-dump evidence is in #1888.

### This is a known hazard in Bitcoin Core

Core hit the same cycle and documented it in the code, directly above the relock scheduling ([permalink, `21b42f3`](https://github.com/bitcoin/bitcoin/blob/21b42f3c5569f9c884c41d828c087ea971697875/src/wallet/rpc/encrypt.cpp#L91-L95)):

```cpp
// rpcRunLater must be called without cs_wallet held otherwise a deadlock
// can occur. The deadlock would happen when RPCRunLater removes the
// previous timer (and waits for the callback to finish if already running)
// and the callback locks cs_wallet.
AssertLockNotHeld(wallet->cs_wallet);
```

Relevant Core changes:

- [bitcoin/bitcoin#18814](https://github.com/bitcoin/bitcoin/pull/18814) (merged 2020-05-13) — "Relock wallet only if most recent callback": adds the `nRelockTime != relock_time` generation check and `m_unlock_mutex`. Fixes [#18811](https://github.com/bitcoin/bitcoin/issues/18811).
- [bitcoin/bitcoin#32862](https://github.com/bitcoin/bitcoin/pull/32862) (merged 2025-07-08) — "use CScheduler for relocking wallet and remove RPCTimer": moves the relock off the libevent timer entirely and deletes `RPCRunLater`/`RPCTimerInterface`. Superseded [#32796](https://github.com/bitcoin/bitcoin/pull/32796), part of [#31194](https://github.com/bitcoin/bitcoin/issues/31194) (remove libevent). Its motivation was libevent removal rather than this deadlock, but it does eliminate the hazard structurally.

## Code changes brief

Three commits, mirroring Core's #18814.

**1. Break the deadlock** — `Fix deadlock between walletpassphrase and the wallet relock timer`

- `LOCK2(cs_main, pwallet->cs_wallet)` moves into a block that ends before `RPCRunLater` is called, so the relock is scheduled with no wallet lock held. This alone breaks the cycle — the pending callback can acquire `cs_wallet`, finish, and let `event_del` return.
- Releasing the lock earlier opens a small window in which the old timer's callback could lock a wallet that was just unlocked again. `LockWallet` now takes the `nRelockTime` it was scheduled with and returns without doing anything if `pwallet->nRelockTime` no longer matches. `walletlock` already resets `nRelockTime` to `0`, so a callback that fires after an explicit lock is also correctly ignored.

**2. Restore the serialization `cs_main` used to provide** — `Serialize walletpassphrase and guard the RPC timer map`

Scoping the lock also removed the mutual exclusion that holding `cs_main` incidentally gave the two steps after a successful unlock — publishing `nRelockTime` and installing the timer:

- A new per-wallet `cs_unlock` is held across both, so two overlapping `walletpassphrase` calls cannot publish `nRelockTime` and schedule the timer in opposite orders, which would leave the wallet unlocked past its timeout. It deliberately is not `cs_wallet`: it is held across `RPCRunLater`, whose callback takes `cs_wallet`.
- `deadlineTimers` in `rpc/server.cpp` gets its own `cs_deadlineTimers`. It is a plain `std::map` that `RPCRunLater` erases from and emplaces into and that `StopRPC` clears, previously unserialized. This also closes the pre-existing `StopRPC` / in-flight-`walletpassphrase` race, since `StopRPC` (`init.cpp:267`) runs before the HTTP workers are joined by `StopHTTPServer` (`init.cpp:268`).

The resulting lock order is acyclic — `cs_unlock → cs_main → cs_wallet` and `cs_unlock → cs_deadlineTimers` — and no timer callback acquires either new lock.

**3. Validate the timeout** — `Validate walletpassphrase timeout before unlock`

Written by @reubenyap and cherry-picked in at his request; ported from upstream Core's `walletpassphrase` validation.

- A negative `timeout` is rejected with `RPC_INVALID_PARAMETER` instead of scheduling a relock in the past.
- The timeout is clamped to `100000000` seconds. Core's stated reason for that bound is that larger values trigger a macos/libevent bug; it also keeps `GetTime() + nSleepTime` from overflowing `nRelockTime`.
- Both checks run before `Unlock()`, so a rejected call leaves wallet state untouched. Previously the timeout was only parsed after `Unlock()` and `TopUpKeyPool()`, so a non-integer value threw with the wallet already unlocked and no relock scheduled at all.
- `qa/rpc-tests/wallet-encryption.py` covers the rejection and that the wallet stays locked afterwards.

### Not included

I kept `RPCRunLater` rather than porting Core's scheduler rework (#32862), to keep the change small. Happy to do the larger version instead if you prefer — it removes the "destroy an event from inside a lock its callback needs" pattern rather than working around it.

Core's `weak_ptr` unload guard from #18814 does not apply here: this build has a single static `pwalletMain` and no dynamic wallet unload.

## Testing

The deadlock was diagnosed from a symbolized `thread apply all bt` taken on a live hung node, which pins both sides of the cycle to exact call sites (see #1888).

`qa/rpc-tests/wallet-encryption.py` covers unlock, timeout relock, timeout extension and now negative-timeout rejection. It is listed in `qa/pull-tester/rpc-tests.py`, which the Jenkins pipeline runs as `qa/pull-tester/rpc-tests.py -extended`. The deadlock itself is a timing race that is impractical to test deterministically — upstream added no test for #18814 either.

I still do not have a build environment for this repo, so I have not compiled or run the patched binary myself. @reubenyap reports having compile-checked both the branch head and the timeout-validation commit locally. The Jenkins run is the real gate, and it has not run on this branch yet.

### DIFF
--- a/qa/rpc-tests/wallet-encryption.py
+++ b/qa/rpc-tests/wallet-encryption.py
@@ -64,6 +64,11 @@ class WalletEncryptionTest(BitcoinTestFramework):
         assert_raises_jsonrpc(-1, "walletpassphrasechange <oldpassphrase> <newpassphrase>\nChanges the wallet passphrase from <oldpassphrase> to <newpassphrase>.",
             self.nodes[0].walletpassphrasechange, '', 'ff')
 
+        # Check that a negative timeout is rejected without unlocking the wallet
+        assert_raises_jsonrpc(-8, "Timeout cannot be negative.", self.nodes[0].walletpassphrase, passphrase, -10)
+        otp = self.get_otp(address)
+        assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address, otp)
+
         # Check that walletpassphrase works
         self.nodes[0].walletpassphrase(passphrase, 2)
         otp = self.get_otp(address)
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -34,7 +34,11 @@ static std::string rpcWarmupStatus("RPC server started");
 static CCriticalSection cs_rpcWarmup;
 /* Timer-creating functions */
 static RPCTimerInterface* timerInterface = NULL;
-/* Map of name to timer. */
+/* Map of name to timer, and the lock protecting it. Callers may run on any of
+ * the RPC worker threads, so every access has to be serialized. The lock is
+ * held while a timer is destroyed, which blocks until a running timer callback
+ * has returned, so no timer callback may acquire it. */
+static CCriticalSection cs_deadlineTimers;
 static std::map<std::string, std::unique_ptr<RPCTimerBase> > deadlineTimers;
 
 static struct CRPCSignals
@@ -408,7 +412,10 @@ void InterruptRPC()
 void StopRPC()
 {
     LogPrint("rpc", "Stopping RPC\n");
-    deadlineTimers.clear();
+    {
+        LOCK(cs_deadlineTimers);
+        deadlineTimers.clear();
+    }
     DeleteAuthCookie();
     g_rpcSignals.Stopped();
 }
@@ -618,6 +625,7 @@ void RPCRunLater(const std::string& name, boost::function<void(void)> func, int6
 {
     if (!timerInterface)
         throw JSONRPCError(RPC_INTERNAL_ERROR, "No timer handler registered for RPC");
+    LOCK(cs_deadlineTimers);
     deadlineTimers.erase(name);
     LogPrint("rpc", "queue run of timer %s in %i seconds (using %s)\n", name, nSeconds, timerInterface->Name());
     deadlineTimers.emplace(name, std::unique_ptr<RPCTimerBase>(timerInterface->NewTimer(func, nSeconds*1000)));
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3035,6 +3035,18 @@ UniValue walletpassphrase(const JSONRPCRequest& request)
         );
     }
 
+    // Prevent concurrent walletpassphrase calls for the same wallet. Without
+    // this, two overlapping calls can update nRelockTime and schedule the
+    // relock timer in opposite orders, so the slower call installs the last
+    // timer while the faster call owns nRelockTime. The stale callback then
+    // does nothing (see LockWallet) and no timer is left, i.e. the wallet stays
+    // unlocked past the requested timeout.
+    //
+    // This has to be a separate lock rather than cs_wallet, because it is held
+    // across RPCRunLater() below, which blocks on the relock callback, which in
+    // turn takes cs_wallet.
+    LOCK(pwallet->cs_unlock);
+
     int64_t nSleepTime;
     int64_t nRelockTime;
     {
@@ -3071,14 +3083,15 @@ UniValue walletpassphrase(const JSONRPCRequest& request)
         nRelockTime = pwallet->nRelockTime;
     }
 
-    // Schedule the relock *after* releasing cs_wallet. RPCRunLater() erases any
-    // previously scheduled lockwallet timer, and destroying a libevent timer
-    // blocks until that timer's callback has finished executing. When the
-    // callback (LockWallet) happens to be running at that moment it is itself
-    // waiting for cs_wallet, so holding the lock across this call deadlocks the
-    // RPC worker against the HTTP event loop thread. Because the RPC worker also
-    // holds cs_main, every other thread that needs cs_main then piles up behind
-    // it and the whole node stops making progress.
+    // Schedule the relock *after* releasing cs_wallet, but still under cs_unlock.
+    // RPCRunLater() erases any previously scheduled lockwallet timer, and
+    // destroying a libevent timer blocks until that timer's callback has finished
+    // executing. When the callback (LockWallet) happens to be running at that
+    // moment it is itself waiting for cs_wallet, so holding the lock across this
+    // call deadlocks the RPC worker against the HTTP event loop thread. Because
+    // the RPC worker also holds cs_main, every other thread that needs cs_main
+    // then piles up behind it and the whole node stops making progress.
+    AssertLockNotHeld(pwallet->cs_wallet);
     RPCRunLater(strprintf("lockwallet(%s)", pwallet->strWalletFile), boost::bind(LockWallet, pwallet, nRelockTime), nSleepTime);
 
     return NullUniValue;
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3065,6 +3065,18 @@ UniValue walletpassphrase(const JSONRPCRequest& request)
         // Alternately, find a way to make request.params[0] mlock()'d to begin with.
         strWalletPass = request.params[0].get_str().c_str();
 
+        nSleepTime = request.params[1].get_int64();
+        // Timeout cannot be negative, otherwise it will relock immediately
+        if (nSleepTime < 0) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Timeout cannot be negative.");
+        }
+        // Clamp the timeout so GetTime() + nSleepTime cannot overflow
+        // nRelockTime (same bound as upstream Bitcoin Core's MAX_SLEEP_TIME)
+        constexpr int64_t MAX_SLEEP_TIME = 100000000;
+        if (nSleepTime > MAX_SLEEP_TIME) {
+            nSleepTime = MAX_SLEEP_TIME;
+        }
+
         if (strWalletPass.length() > 0)
         {
             if (!pwallet->Unlock(strWalletPass)) {
@@ -3078,7 +3090,6 @@ UniValue walletpassphrase(const JSONRPCRequest& request)
 
         pwallet->TopUpKeyPool();
 
-        nSleepTime = request.params[1].get_int64();
         pwallet->nRelockTime = GetTime() + nSleepTime;
         nRelockTime = pwallet->nRelockTime;
     }
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2995,9 +2995,14 @@ UniValue keypoolrefill(const JSONRPCRequest& request)
 }
 
 
-static void LockWallet(CWallet* pWallet)
+static void LockWallet(CWallet* pWallet, int64_t nRelockTime)
 {
     LOCK(pWallet->cs_wallet);
+    // Skip if this is not the most recent relock callback. walletpassphrase()
+    // may have been called again in the meantime (extending the timeout), or
+    // walletlock() may have already locked the wallet and reset nRelockTime.
+    if (pWallet->nRelockTime != nRelockTime)
+        return;
     pWallet->nRelockTime = 0;
     pWallet->Lock();
 }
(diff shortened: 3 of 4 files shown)


Danswar:
Thanks — both findings are correct, and both are addressed in `67c1623`.

**1. Unsynchronized `deadlineTimers`.** Confirmed. `deadlineTimers` (`src/rpc/server.cpp:38`) is a plain `std::map` that `RPCRunLater()` erases from and emplaces into, and that `StopRPC()` clears, with no lock of its own. `walletpassphrase` is the only caller of `RPCRunLater()` in the tree, so before this PR those accesses were serialized only as a side effect of the RPC holding the global `cs_main`. With up to `-rpcthreads` (default 4) HTTP workers, releasing that lock earlier removes the serialization. `deadlineTimers` now has its own `cs_deadlineTimers` covering the `erase`/`emplace` in `RPCRunLater()` and the `clear()` in `StopRPC()` — matching upstream Bitcoin Core, which guards the same map with `g_deadline_timers_mutex`. That also closes a pre-existing race: `StopRPC()` runs before `StopHTTPServer()` joins the worker threads (`src/init.cpp:267-268`), so the `clear()` could already race with an in-flight `walletpassphrase`.

**2. An older request replacing the newer timer.** Confirmed, and it was a regression introduced by splitting the lock scope in the previous commit: publishing `nRelockTime` and installing the timer used to be atomic under `cs_main`, and no longer were. With the interleaving you describe, the slower call installs the surviving timer carrying its own generation, the generation check then correctly rejects it, and no timer is left — so the wallet stays unlocked past the requested timeout. Fixed with a per-wallet `cs_unlock` that `walletpassphrase` holds across both the unlock and `RPCRunLater()`, making the two steps atomic with respect to each other. Same approach as Bitcoin Core's `CWallet::m_unlock_mutex` ([bitcoin/bitcoin#18814](https://github.com/bitcoin/bitcoin/pull/18814), "Prevent concurrent calls to walletpassphrase with the same wallet").

On why `cs_unlock` is a separate lock and not `cs_wallet`: it is deliberately held across `RPCRunLater()`, which blocks until a running relock callback returns, and that callback takes `cs_wallet` — reusing `cs_wallet` here would reintroduce exactly the deadlock this PR fixes. Neither new lock is acquired by any timer callback, so the resulting order (`cs_unlock` → `cs_main` → `cs_wallet`, and `cs_unlock` → `cs_deadlineTimers`) is acyclic. I also added an `AssertLockNotHeld(pwallet->cs_wallet)` before the `RPCRunLater()` call so the invariant is machine-checked in `DEBUG_LOCKORDER` builds.

For context on scope: this release resolves every RPC to the single global `pwalletMain` (`GetWalletForJSONRPCRequest()` ignores the request and returns it directly), and `-wallet` takes a single filename, so the per-wallet lock alone would in practice serialize all `RPCRunLater()` calls today. I still gave the map its own lock rather than relying on that, since it depends on `walletpassphrase` remaining the only caller and does not cover the `StopRPC()` path.

Still unbuilt on my side, as noted in the PR description — the reasoning above is from reading the code, so the compile and the functional tests are worth treating as unverified until CI runs.

reubenyap:
@codeant-ai review

reubenyap (commented):
## Review summary

This is a correct, well-scoped fix that faithfully ports Bitcoin Core's solution. I verified the compile locally (details below) — recommend approval once CI has been run for this first-time contributor.

### Verification performed

- **Compile check** (the one open risk — the author noted they could not build): I built the bls-dash dependency and compiled both changed translation units at the PR head (`67c1623`) with GCC 13 in Release config — `rpc/server.cpp` and `wallet/rpcwallet.cpp` (which includes the modified `wallet.h`). **Both compile cleanly with no new warnings.** The constructs I specifically checked: `AssertLockNotHeld` exists in `src/sync.h:95` with a real `DEBUG_LOCKORDER` implementation in `sync.cpp:171` (no link failure in lock-debug builds), and the three-argument `boost::bind` is fine.
- **The deadlock mechanism is real:** `HTTPRPCTimer` owns an `HTTPEvent` whose destructor calls `event_free` (`httpserver.cpp:533`), and `evthread_use_pthreads` is enabled (`httpserver.cpp:405`), so destroying the timer genuinely blocks until a running callback returns. The thread dumps in #1888 match this cycle.
- **Faithful to upstream, as claimed.** Diffed against Bitcoin Core v28.0: the wallet side is a direct port of bitcoin/bitcoin#18814 (scoped `cs_wallet` block, captured relock generation, generation check in the callback, `AssertLockNotHeld`, and `cs_unlock` ↔ Core's `m_unlock_mutex`, correctly renamed to this codebase's `cs_` convention). The `server.cpp` hunk is line-for-line equivalent to Core's `g_deadline_timers_mutex`, guarding the same two sites (`RPCRunLater`, `StopRPC`). The parts of Core's fix that were *not* ported are correctly inapplicable here: the `weak_ptr` unload guard (single static `pwalletMain`, no dynamic unload) and the #32862 scheduler rework (deliberately skipped to stay minimal — the right call).
- **Supporting claims all check out:** `walletpassphrase` is the only `RPCRunLater` caller in the tree; `walletlock` resets `nRelockTime` under `cs_wallet` (`rpcwallet.cpp:3186`), so the generation check handles explicit locks; `StopRPC` runs before HTTP workers are joined (`init.cpp:267-268`), so `cs_deadlineTimers` does close a pre-existing shutdown race.

### Correctness

- Lock ordering is acyclic: `cs_unlock → cs_main → cs_wallet` and `cs_unlock → cs_deadlineTimers`; neither new lock is acquired by any timer callback, and `cs_unlock` has a single acquisition site.
- A stale callback firing between the `cs_wallet` release and `RPCRunLater` fails the generation check; `cs_unlock` prevents the overlapping-unlock ordering bug that the first commit alone would have introduced.
- No behavior change outside the fixed path: error paths, help text, `TopUpKeyPool` placement and the `RPCRunLater`-throws case are identical to master; the `SecureString` passphrase is now destroyed earlier, a marginal improvement.
- Functional coverage exists via `qa/rpc-tests/wallet-encryption.py` (unlock, timeout relock, timeout extension), which runs in CI. The deadlock itself is a timing race that is impractical to test deterministically — upstream added no test for #18814 either.

### Bot findings triage

- The two Codex P2 findings and the CodeRabbit finding were real and are fixed by `67c1623`; the author's replies are technically accurate. I am resolving those threads.
- The two codeant-ai "Critical" findings can be dismissed:
  - *"Use-after-free: `event_free()` doesn't synchronize with a running callback"* — incorrect. With `evthread_use_pthreads` enabled, `event_free`/`event_del` from another thread blocks until the running callback returns; that blocking is precisely the deadlock mechanism this PR fixes. The `StopRPC` timer-clearing placement is also unchanged from master.
  - *"`timerInterface` dangling pointer during shutdown"* — a real but **pre-existing** theoretical race (`StopHTTPRPC` at `init.cpp:265` deletes the interface while workers may still run until `StopHTTPServer` at `:268` joins them). This PR neither introduces nor meaningfully widens it; it belongs in a separate cleanup, ideally a Core-#32862-style scheduler rework.

### Minor notes (non-blocking)

- Roughly half the added lines are comments — more verbose than Core's four-line equivalent, but the lock-rationale content is genuinely useful; fine to keep.
- Pre-existing gaps Core later hardened (not regressions of this PR): no rejection of negative `timeout` and no clamp on huge values that can overflow `GetTime() + nSleepTime`. A follow-up adding Core's validation is being prepared separately.
- Process nits, no code change needed: the PR title lacks the repo's `Wallet:` area prefix, and the description doesn't follow the template headings (`PR intention` / `Code changes brief`).

---

reubenyap:
@Danswar — we'd like to fold the timeout-validation hardening mentioned in the review into this PR rather than land it separately. The commit is ready on a branch in this repo, stacked directly on your `67c1623`, so it cherry-picks cleanly:

```bash
git fetch https://github.com/firoorg/firo.git claude/review-pr-1889-firo-ygv719
git cherry-pick 8b4d7393a559f426a8df2691009bdf18c5d13728
git push
```

What it does (mirrors upstream Bitcoin Core's `walletpassphrase` validation):

- Rejects a negative `timeout` with `RPC_INVALID_PARAMETER` ("Timeout cannot be negative.") instead of scheduling a relock in the past.
- Clamps the timeout to 100000000 seconds so `GetTime() + nSleepTime` cannot overflow `nRelockTime`.
- Both checks run before `Unlock()`, so a rejected call leaves wallet state untouched.
- Adds coverage to `qa/rpc-tests/wallet-encryption.py` for the rejection and for the wallet staying locked afterwards.

The commit was compile-checked against your branch head (`rpcwallet.cpp` builds cleanly, no new warnings) and the test file passes `py_compile`. Since the fork branch can't be pushed to from our side, the cherry-pick needs to come from you — happy to answer anything about it.

---

Danswar:
@reubenyap thanks, I addressed your comments

reubenyap:
Confirmed — the cherry-pick at `a477c7e` matches the prepared commit exactly (tree-identical), and the description now follows the template. Everything from the review round is addressed; nothing further needed from your side.

Next step is on us: approving the CI workflow run for a first-time contributor, then review sign-off from the requested reviewers. Thanks for the quick turnaround and for a well-researched fix.

---

Danswar:
`guix-x86_64-linux-gnu` is red for an unrelated upstream reason. I don't have re-run rights on this repo — could a maintainer re-run that job?

The failure is in the Guix bootstrap, not in the build:

```
guix time-machine: error: Git error: cannot locate remote-tracking branch 'origin/keyring'
```

codeberg.org was degraded on 2026-07-31, so `guix time-machine` fell back to the Software Heritage archive. That replay only restores `refs/heads/master`, and without the `keyring` branch it cannot authenticate the pinned channel commit.

It is not specific to this PR:

- The same guix jobs failed on the master push in [run 30616408099](https://github.com/firoorg/firo/actions/runs/30616408099) (there `guix pull` hit an HTTP 503 from the same host), and on unrelated PRs the same day.
- On this PR's run the other four guix platforms passed on the identical commit (a477c7e).
- The diff touches only `src/rpc/server.cpp`, `src/wallet/rpcwallet.cpp`, `src/wallet/wallet.h` and one functional test.

Upstream looks recovered — `guix-x86_64-linux-gnu` is green again on the master push in [run 30659239531](https://github.com/firoorg/firo/actions/runs/30659239531), so a re-run should pass.

reubenyap:
Verified — the `guix-x86_64-linux-gnu` failure is the Guix channel bootstrap (`cannot locate remote-tracking branch 'origin/keyring'` after the Software Heritage fallback), hit before any Firo code was compiled. The other 11 checks on a477c7e are green, and the same job passed on master's latest run after codeberg recovered, so a re-run should clear it.

The re-run can't be triggered via the API from here, so it needs a manual "Re-run failed jobs" click — flagged for that. No action needed on your side.

---

Danswar:
sorry closed by mistake, reopened

Danswar:
any chances of moving forward with this fix? anything else I can do to help? 🙏

reubenyap:
Status update: CI was fully re-run on Aug 7 ([run 30912163059](https://github.com/firoorg/firo/actions/runs/30912163059)) and the earlier guix infra failure is cleared — `guix-x86_64-linux-gnu` is green now.

The only remaining red is `linux-cmake-Debug`, and I've verified it's a flaky test, not this PR: all 91 unit tests passed, and of the RPC suite only `dip3-deterministicmns.py` failed (`AssertionError: mnlists does not match provided mns` at the chain-revert step) — a masternode-list test nowhere near this diff, which **passed on this exact commit** in the July 31 run. `wallet-encryption.py`, the test this PR touches, passed.

So there's nothing for you to do — the branch has effectively demonstrated green across both runs combined. It needs one more "Re-run failed jobs" click from a maintainer, then reviewer sign-off. I've flagged it again on our side; sorry for the slow turnaround here.

---

Issue #1887, opened 2026-07-28
Wallet: reorg no longer removes Spark mints created by spend transactions (regression from #1873)


reubenyap:
## Summary

Since #1873, `DisconnectTip` no longer removes wallet Spark mint coins that were created as outputs of Spark **spend** transactions (e.g. change from a private spend) when their block is disconnected in a reorg. Mint coins from pure mint transactions are unaffected.

## Mechanism

`DisconnectTip` populates `block.sparkTxInfo` by re-running `CheckTransaction` with `fStatefulZerocoinCheck=false` (`src/validation.cpp:3234-3239`), then feeds the result to the wallet (`src/validation.cpp:3323-3330`):

```cpp
if (block.sparkTxInfo->spentLTags.size() > 0)
    pwalletMain->sparkWallet->RemoveSparkSpends(block.sparkTxInfo->spentLTags);
if (block.sparkTxInfo->mints.size() > 0)
    pwalletMain->sparkWallet->RemoveSparkMints(block.sparkTxInfo->mints);
```

#1873 added an early return to `CheckSparkSpendTransaction` (`src/spark/state.cpp:672`):

```cpp
if (!fStatefulSigmaCheck)
    return true;
```

Before #1873, a `fStatefulSigmaCheck=false` call still executed `CheckSparkSMintTransaction`, which pushed the spend transaction's smint outputs into `sparkTxInfo->mints`. Now the function returns before that call, so for spark spend transactions `block.sparkTxInfo->mints` stays empty in `DisconnectTip`, and `RemoveSparkMints` misses those coins. The wallet then carries mint coins whose creating block no longer exists on the active chain.

## Pre-existing gap in the same path (not a #1873 regression)

`block.sparkTxInfo->spentLTags` was **already** unpopulated on this path before #1873: the old code had `if (!fStatefulSigmaCheck) return passVerify;` *before* the `spentLTags` bookkeeping block, so `RemoveSparkSpends` at `src/validation.cpp:3325` has been receiving an empty set all along. If wallet spend-rollback on reorg is expected to happen here, that has been broken independently and for longer.

## Consensus is not affected

Verified all `fStatefulSigmaCheck=false` call sites:

- `ConnectBlock` passes `true` with `block.sparkTxInfo` (`src/validation.cpp:2760`) — chainstate bookkeeping (`spentLTags`, anonymity-set mints) runs in full.
[...]

PR #1890, opened 2026-07-31
Remove sparksmint from wallet after reorg

levonpetrosyan93:
Fix for #1887

### DIFF
--- a/src/spark/state.cpp
+++ b/src/spark/state.cpp
@@ -669,11 +669,6 @@ bool CheckSparkSpendTransaction(
 
     LogPrintf("CheckSparkSpendTransaction: tx metadata hash=%s\n", txHashForMetadata.ToString());
 
-    if (!fStatefulSigmaCheck)
-        return true;
-    bool isMempoolAcceptance = (!sparkTxInfo);
-    bool passVerify = false;
-
     uint64_t Vout = 0;
     std::size_t private_num = 0;
     for (const CTxOut &txout : tx.vout) {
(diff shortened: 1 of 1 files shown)


reubenyap:
I rechecked the current head (`d9483f26a0`) against both the disconnect and block-validation paths. Both findings remain valid.

1. **Wallet removal still receives contextless SMints.** The existing [serial-context thread](https://github.com/firoorg/firo/pull/1890#discussion_r3687590489) is correct. `CheckSparkSMintTransaction` deserializes each output and stores it in `sparkTxInfo->mints`, but `Coin::serial_context` is not serialized. Spend-created SMints bind `S` to `getSerialContext(tx)`, which serializes the LTags. `DisconnectTip` then passes the contextless copies to `RemoveSparkMints`; `identify()` fails its serial-commitment check, the exception is caught, and `eraseMint()` is never reached. There is no later reorg callback that repairs this. The existing `GetSparkMintCoins(tx)` helper restores the context.

2. **The moved early checks reduce the peer DoS score from 100 to 0.** The mixed-mint and private-output-limit branches return bare `false`. They now run during stateless `CheckBlock`, whose wrapper converts the failure to `Invalid()` with `DoS(0)`. Before this PR, they ran in `ConnectBlock`, whose wrapper applied `DoS(100)`. The block remains invalid, but a normal full-block source is no longer penalized. I found no compensating `Misbehaving()` path. This does not change consensus validity, and exploitation requires a valid-PoW invalid block, but it is an unintended peer-policy regression.

The smallest fix appears to be leaving these validation checks on the stateful side and gathering disconnect rollback coins with `GetSparkMintCoins(*tx)` at `DisconnectTip`. That reuses the existing context-aware parser and also covers pure Spark mints.

Please add one regression test that exercises a wallet-owned Spark spend through a reorg and asserts the mint metadata is erased. `Coin::operator==` ignores `serial_context`, so merely comparing `CSparkTxInfo::mints` will not catch this bug.

This recheck was static because the local checkout has no `test_firo` binary.

PR #1891, opened 2026-07-31
Qt: Make arbitrary funds private with Spark

reubenyap:
## PR intention

Make the Spark privacy action understandable without requiring users to know what Spark is, let them choose how much of their transparent balance to make private, and recover safely when the wallet cannot create the requested transaction.

## Code changes brief

- Replace the detached `Anonymize All` action with `Make Private...` beside the eligible transparent balance.
- Add a focused amount dialog with source and destination labels, an exact amount field, and a `Max` option.
- Reuse the existing Spark mint preparation and commit path, including fee subtraction for `Max` and an exact fee review before confirmation.
- If preparation fails, preserve the entered amount and offer `Change Amount` or `Cancel`. No automatic splitting, silent amount changes, or fund movement is added. Encrypted wallets relock before amount editing.
- Show the wallet's original transaction-creation error under `Details`, while keeping the main message concise.
- Preserve that error for the existing Send screen instead of emitting a duplicate wallet popup.
- Update related automatic privacy and settings copy to use the same user-facing language.
- Share the existing wallet worker helper so preparing and committing the Spark transaction does not freeze GUI repainting.

## Testing

- `git diff --check`
- Source-level checks for the recovery controls and propagated wallet error
- Added a Qt unit check that preserves transaction-creation error details
- Parsed all modified Qt `.ui` files as XML
- Local Qt build and test binary not run because this Windows checkout has no CMake/compiler toolchain and WSL is not installed

### DIFF
--- a/src/qt/automintdialog.cpp
+++ b/src/qt/automintdialog.cpp
@@ -23,7 +23,7 @@ AutoMintSparkDialog::AutoMintSparkDialog(AutoMintSparkMode mode, QWidget *parent
     ENTER_CRITICAL_SECTION(pwalletMain->cs_wallet);
 
     ui->setupUi(this);
-    ui->buttonBox->button(QDialogButtonBox::Ok)->setText(tr("Anonymize"));
+    ui->buttonBox->button(QDialogButtonBox::Ok)->setText(tr("Make Private"));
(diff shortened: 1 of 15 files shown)


reubenyap:
@codeant-ai review

PR #1893, opened 2026-07-31
Spark: Fix spend key buffer reuse

reubenyap:
## PR intention

Prevent the Windows access violation observed after clicking **Anonymize All**.

Spark deterministic spend-key construction cleared its hash input and output vectors, then wrote through storage outside the vectors' live element ranges. This undefined behavior could corrupt memory and surface later in unrelated code, which explains the misleading `CNodeStats` and `std::string` frames in the reported stack trace.

## Code changes

- Keep both key-derivation buffers at their required sizes while they are reused.
- Reset sensitive intermediate material with `memory_cleanse` instead of shrinking the vectors with `clear()`.
- Use `vector::data()` only while the buffers contain live elements.
- Add a regression test for deterministic `s1` and `s2` derivation from the same spend-key scalar.

## Testing

- `git diff --check`
- Added `spark_address_tests/spend_key_recovery`
- Full Boost test execution is pending CI because this checkout has no configured build directory.

### DIFF
--- a/src/libspark/keys.cpp
+++ b/src/libspark/keys.cpp
@@ -17,28 +17,29 @@ SpendKey::SpendKey(const Params* params) {
 SpendKey::SpendKey(const Params* params, const Scalar& r_) {
     this->params = params;
     this->r = r_;
-    std::vector<unsigned char> data;
-    data.resize(32);
+    std::vector<unsigned char> data(32);
     r.serialize(data.data());
-    std::vector<unsigned char> result(CSHA256().OUTPUT_SIZE);
+    std::vector<unsigned char> result(CSHA256::OUTPUT_SIZE);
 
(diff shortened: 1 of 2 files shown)


reubenyap:
Do not review this. Has serious issues.

reubenyap:
Closing in favor of #1894. Keeping `data` sized here changes the `s2` seed hash from `SHA256d("s2_generation")` to `SHA256d("s2_generation" || ser(s1))`, which changes every existing wallet's view keys and addresses — the same derivation break that #1767 introduced and #1784 reverted. #1894 fixes the out-of-range vector writes while keeping the hash inputs byte-for-byte identical, and adds a test that pins the deployed derivation so it can't silently drift again.

---

PR #1894, opened 2026-07-31
Spark: fix out-of-range vector writes in SpendKey derivation

reubenyap:
## PR intention

Replaces #1893 with a version that cleans up the undefined behavior in `SpendKey::SpendKey(const Params*, const Scalar&)` **without changing key derivation**.

The current code clears its hash input and output vectors and then keeps writing through `data()` / `&result[0]` on empty vectors. That is undefined behavior on paper (and asserts in MSVC debug builds), but in practice the writes stay inside the retained capacity of the same allocations, so it never corrupted memory in release builds — it does not explain the reported Windows access violations.

The dangerous part of #1893 (and of #1767 before it, reverted in #1784) is that keeping the buffers sized silently changes the `s2` seed hash from `SHA256d("s2_generation")` to `SHA256d("s2_generation" || ser(s1))`. Since `P2 = F*s2 + D` feeds the full/incoming view keys and every address's `Q2`, that would change every existing wallet's addresses on upgrade or seed restore, orphaning previously received funds. The constant-`s2` derivation is a known, accepted quirk (it makes incoming view keys effectively full view keys) and is consensus-frozen for wallet compatibility.

## Code changes brief

- `Tests:` add `spark_address_tests/spend_key_derivation`, which pins the deployed derivation before any code change: determinism of `(s1, s2)` in `r`, `s1` depends on `r` while `s2` does not, and both scalars match an explicit re-derivation via `CHash256` with the `s2` seed hashing **only** the `"s2_generation"` prefix. Any future derivation drift (as attempted by #1767 and #1893) now fails CI.
- `Spark:` keep both buffers at their required sizes for the whole derivation (no writes outside the vectors' live element range), drop the dead `s1.serialize()` into a never-hashed buffer, wipe the buffers with `memory_cleanse()` before scope exit instead of `clear()` (so key material is not left readable in freed heap memory), and document the frozen `s2` derivation in place.

The `s2` seed hash input remains byte-for-byte identical to the deployed derivation, so existing wallets keep their view keys and addresses.

https://claude.ai/code/session_01HVp6JH7LvzRpZXU4VEjdDG

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVp6JH7LvzRpZXU4VEjdDG)_

### DIFF
--- a/src/libspark/keys.cpp
+++ b/src/libspark/keys.cpp
@@ -17,28 +17,31 @@ SpendKey::SpendKey(const Params* params) {
 SpendKey::SpendKey(const Params* params, const Scalar& r_) {
     this->params = params;
     this->r = r_;
-    std::vector<unsigned char> data;
-    data.resize(32);
+    std::vector<unsigned char> data(32);
     r.serialize(data.data());
-    std::vector<unsigned char> result(CSHA256().OUTPUT_SIZE);
+    std::vector<unsigned char> result(CSHA256::OUTPUT_SIZE);
 
     CHash256 hash256;
     std::string prefix1 = "s1_generation";
     hash256.Write(reinterpret_cast<const unsigned char*>(prefix1.c_str()), prefix1.size());
     hash256.Write(data.data(), data.size());
-    hash256.Finalize(&result[0]);
-    this->s1.memberFromSeed(&result[0]);
+    hash256.Finalize(result.data());
+    this->s1.memberFromSeed(result.data());
 
-    data.clear();
-    result.clear();
     hash256.Reset();
-    s1.serialize(data.data());
 
+    // Consensus-critical: the s2 seed commits only to the prefix below. The
+    // historical code cleared `data` before this point, so its Write()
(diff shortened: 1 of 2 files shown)


reubenyap:
CI note: `guix-x86_64-w64-mingw32` failed during environment setup — `guix pull` got a 504 from codeberg.org before the source tree was ever built, so it's an infrastructure flake unrelated to this diff. I'll re-run the failed job once the workflow run completes.

---

reubenyap:
CI status: all six compile-and-test jobs are green (linux/windows/mac, Debug and Release), including the new `spark_address_tests/spend_key_derivation` test. The five red guix jobs all failed in environment setup during a codeberg.org outage (~08:20–08:34 UTC) before building any source.

I tried to re-run the failed jobs now that the run is complete, but the integration doesn't have permission (`403`). Could a maintainer hit **Re-run failed jobs** on [the workflow run](https://github.com/firoorg/firo/actions/runs/30615672455)? codeberg has had time to recover, so they should pass on retry.

---

PR #1897, opened 2026-08-11
Consensus: Bind FiroPoW header height

reubenyap:
### Motivation

- Prevent a header-level trust boundary where a serialized FiroPoW `nHeight` can differ from the chain context and allow a header to bypass full ProgPoW mix verification or create an index/header inconsistency. 

### Description

- Add a contextual header check in `ContextualCheckBlockHeader` to require that for ProgPoW headers `block.nHeight == pindexPrev->nHeight + 1`, and reject headers that do not match that invariant. 
- Add a unit test `progpow_header_height_must_match_chain_height` in `src/test/main_tests.cpp` that constructs a post-activation header with a mismatched `nHeight` and asserts it is rejected with `bad-blk-progpow`. 
- This change prevents malformed ProgPoW headers from entering the block index and avoids reconstructed-header/hash mismatches later in the load/serve path.

### Testing

- `git diff --check` and `git status --short` were run to validate the working tree state and reported no whitespace/format errors. 
- Building and running the unit test target with `cmake --build build --target test_firo -j2` could not be completed in this environment because the repository has no configured `build` directory, so the new test was not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7aa10cbf74832fa34e5fbdbb4af41f)

### DIFF
--- a/src/test/main_tests.cpp
+++ b/src/test/main_tests.cpp
@@ -36,6 +36,21 @@ BOOST_AUTO_TEST_CASE(block_subsidy_test)
     //TestBlockSubsidyHalvings(1000); // Just another interval
 }
 
+BOOST_AUTO_TEST_CASE(progpow_header_height_must_match_chain_height)
+{
+    const Consensus::Params& consensusParams = Params().GetConsensus();
+    CBlockIndex previous;
+    previous.nHeight = 100;
+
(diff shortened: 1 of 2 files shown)


reubenyap:
superseded by #1898

PR #1914, opened 2026-08-21
Qt: Redesign wallet UI with light/dark theme support

levoncrypto:
# Summary
- Redesigned the Qt wallet UI: rounded cards, new sidebar navigation, refreshed
  Send/Receive/Transactions/Masternodes/Overview pages and dialogs
- Added full light/dark theme support across the app, with a shared theming
  helper (`guitheme.cpp/h`) and live theme switching
- Fixed several sync-status and masternode-list display bugs surfaced during
  the redesign


reubenyap (changes requested):
Requesting changes for the following user-visible regressions at `6b80a0cc`:

1. **Synchronization details cannot reopen**

   When the latest block is less than 45 minutes old, `BitcoinGUI::setNumBlocks()` hides the modal overlay and calls `hideForever()` (`src/qt/bitcoingui.cpp:1524-1535`). `ModalOverlay::showHide()` then rejects every later attempt to show it, including an explicit user click (`src/qt/modaloverlay.cpp:307-326`). This is why clicking the Syncing card does nothing.

   Automatic presentation can stop once block sync is current, but a user-requested opening must still work. Either allow `userRequested` to override `foreverHidden`, or remove the permanent-hide state.

2. **“Syncing” can remain after synchronization completes**

   Startup calls `setClientModel()` before adding the wallet. The already-synced path hides the navigation card but does not clear `WalletFrame::bOutOfSync`. That member is also uninitialized, and `addWallet()` later passes it into the Overview page, where it selects “Syncing…” versus “Up to date” (`src/qt/bitcoin.cpp:547-556`, `src/qt/walletframe.cpp:41-50`, `src/qt/overviewpage.cpp:848-856`).

   The navigation card also starts visible, while `updateProgressBarLabel()` can show it based only on non-empty text without checking the authoritative sync state. Please initialize the flags, hide the card by default, and derive all sync text and visibility from one state-update path.

3. **The Tor control is effectively lost**

   The checkbox was moved to a footer below the Overview cards, outside a scroll area (`src/qt/forms/overviewpage.ui:550-568`). The form’s intended height grew from 407 to 560 pixels while the application restores users’ previous window geometry, so the footer can be clipped for existing installations. It is also styled as faint 9px text (`src/qt/overviewpage.cpp:484-486`), making an important privacy setting easy to miss.

   Please move Tor to **Settings → Network/Privacy**, show the restart requirement and `-torsetup` override clearly, and expose an obvious “Tor enabled” state after restart. The control should use normal Source Sans Pro body sizing, consistent with the [Firo brand guidelines](https://firo.org/branding/firo-brand-guidelines.pdf).

Before merging, please smoke-test Windows at 100%, 125%, and 150% scaling, including restored small-window geometry, already-synced startup, active synchronization, clicking the sync card before and after completion, light/dark Tor controls, and access to the Spark Names workflow.

These paths were verified statically against the current PR head. I did not run the Qt application locally because no built PR binary was available.

reubenyap (commented):
**Static review of the current head (9e84404).** No visual pass was done (no Qt 6 build was available on the reviewing machine), so everything below comes from reading the diff. The earlier bot findings are all addressed in the current head: the fiat placeholder, watch-only column labels, send-field mnemonics, double-click details, keyboard access on masternode cards, locale-aware numbers, fractional operator rewards, and the request date split.

The redesign is structurally sound: table widgets became proxy models with delegate-painted cards, the filter proxy gained a fast path, and the console has a scrollback cap. Three functional regressions, four GUI-thread cost regressions, and heavy stylesheet duplication remain.

### Bugs to fix before merge

1. **Menu bar text becomes unreadable after a live theme switch on Windows and Linux.** The item colour is baked once with the startup theme and never refreshed, while the bar background does follow the theme. Fix: delete this line and add `QMenuBar::item { color: $INK; }` to the dark override in `guiutil.cpp`.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/bitcoingui.cpp#L510

2. **No sync progress is shown when the wallet is behind but past initial block download.** The status-bar progress bar is now hidden unconditionally, and the sidebar card only shows when `isActivelySyncing()` is true, which depends on the IBD latch that never flips back. A wallet resumed from sleep a day behind shows only the spinner, where master showed "N behind". Fix: gate the card on `syncInProgress()`, which already includes the 45-minute tip-age check, keeping the regtest exclusion.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/bitcoingui.cpp#L1138-L1154

3. **Blocking `cs_main` lock on the GUI thread.** `ClientModel::inInitialBlockDownload()` calls `IsInitialBlockDownload()`, which takes `LOCK(cs_main)` until the latch flips. The PR calls it from a 2-second timer plus every block and peer-count update. During IBD or reindex, validation holds that lock for long stretches, so the GUI will stall; master never called it from the GUI thread. The tip callback (`BlockTipChanged` in `clientmodel.cpp`) already receives `initialSync`; store it in an atomic and return that instead.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/bitcoingui.cpp#L843-L850

### Efficiency and weight

4. **Every block now refreshes every transaction's status.** `updateConfirmations` was widened from 100 rows to all rows. The overview proxy has `showInactive` off, so its `filterAcceptsRow` calls `index()` per row, and `index()` does two try-locks and `updateStatus()` per row whenever the height changed, i.e. every 250 ms during IBD. Conflicted and abandoned transactions already arrive as `CT_UPDATED`, so the wide range is not needed for correctness. Restore the bound.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/transactiontablemodel.cpp#L300-L309

5. **Sidebar fully re-themed on every `showEvent`.** Each restore from tray rebuilds 8 icons with 6 tinted pixmaps apiece, resets 3 stylesheets and repolishes the toolbar. It already runs at construction and on theme change; drop it from `showEvent`.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/bitcoingui.cpp#L2022-L2023

6. **Drop shadows on large repainting cards.** About 24 `QGraphicsDropShadowEffect`s exist. The two on the overview balance and detail cards force an offscreen blur on every balance repaint, with 4x the pixels on Retina. The masternode page already dropped per-card shadows for this reason; the cards have borders, so drop these too.
   https://github.com/firoorg/firo/blob/9e844047851f2b2d565a0c897e79076e56a308d3/src/qt/overviewpage.cpp#L254-L255

7. **Styling belongs in the stylesheet, not per widget.** About 200 `setStyleSheet` calls restate the same field, caption, value and button rules; the primary gradient alone is declared inline in six places besides `primaryButtonStyle()`, in two different directions. A theme switch repolishes everything via `loadTheme()`, then ~30 `applyTheme` receivers repolish their subtrees again. Move the redesign rules into `firo.css` and `darkModeOverrideCss` once; the automint, notification, edit-address, open-URI, intro and manual-mint dialogs then need no theming code at all.

   | Duplicated pattern | Copies | Where |
   |---|---|---|
   | Empty-state widget | 5 | masternodelist, transactionview, sparknamespage, receivecoinsdialog, overviewpage |
   | Card-row delegate prologue | 4 | transactionview, addressbookpage, receivecoinsdialog, masternodelist |
   | Fit-dialog-to-screen | 6 | optionsdialog, recover, createsparknamepage, receiverequestdialog, transactiondescdialog, masternodelist |
   | Pixmap tinting | 2 | per-pixel loop in `guitheme.cpp` vs `CompositionMode_SourceIn` in `bitcoingui.cpp` |

8. **Dead code.** The status-bar progress bar and its hardcoded orange stylesheet (`bitcoingui.cpp:287`) are never visible now. `STYLE_INVALID` is unused. The generic `QPushButton` rule in `TransactionView::applyTheme` is overridden for the only button it matches.

9. **Nits.** `themed()` depends on replacement order to avoid prefix clashes (`$PANEL` vs `$PANEL_SOFT`); one regex pass over a map is safer. The primary button hover rule equals its normal rule, so primary buttons give no hover feedback. Several `applyTheme` bodies recreate the shadow effect on every theme change.

### UI consistency

| Aspect | What differs | Suggestion |
|---|---|---|
| Text units | Redesigned pages use px (12px most common); `firo.css` sets 12pt on QWidget, QLabel, QLineEdit and QPushButton. 12px is about 9pt, so new pages read smaller than untouched dialogs such as Options, Sign/Verify and coin control. | One unit, one scale |
| Corner radius | 13 distinct values from 4px to 22px | Three tokens: control 8, button 12, card 16 |
| Keyboard mnemonics | Send restored them; Receive dropped them on LABEL, AMOUNT and MESSAGE | Restore on Receive |
| Resize font scaling | Send and Receive still scale fonts with window width via `adjustTextSize`; Overview, Transactions and Masternodes no longer do | Delete the remaining three |
| Overview action row | Send and Receive use inline styles with a diagonal gradient; Make Private uses the helper with a vertical one | Use `primaryButtonStyle` / `secondaryButtonStyle` |
| Sort control | Bottom of Transactions, top of the Masternodes filter card | Same place on both |
| Same action, two labels | "Register Spark Name" on Receive, "Create Spark Name" on the Spark Names page | One label |
| Page margins | Overview 24/18, Transactions 24/20, Masternodes 24/20 (code overrides the form's 22/18), Spark Names 22/18 | One constant |
| Wording | "FIRO (Primary)" heads the balance card with no secondary asset; the legend says "Private (Spark):" with a colon but "Transparent" without | "Total balance"; drop the colon |

The CI mirror and translation-tooling changes are unrelated to the redesign but harmless, and the extract-script fix aligns the generated context with what master already ships.

PR #1918, opened 2026-08-25
P2P: Limit LLMQ signing sessions per peer

reubenyap:
## PR intention

Prevent unauthenticated peers from retaining an unbounded number of LLMQ signing-share sessions through QSIGSESANN announcements.

This adapts https://github.com/dashpay/dash/pull/7351 to Firo's older LLMQ implementation.

## Code changes brief

- Cap announcement-created sessions per peer and LLMQ type at `max(4 * quorum size, 100)`.
- Continue accepting re-announcements for known sessions and exclude locally created send-only sessions from the cap.
- Start session expiry on the first announcement without refreshing the timer on repeats, and clean recovered announcement-only sessions.
- Add focused unit coverage for cap enforcement, refreshes, send-only sessions, and per-type isolation.

## Validation

- `git diff --check origin/master...HEAD`
- Unit tests not run locally: no CMake/compiler or `test_firo` binary is available in this checkout.

### DIFF
--- a/src/llmq/quorums_signing_shares.cpp
+++ b/src/llmq/quorums_signing_shares.cpp
@@ -15,10 +15,23 @@
 
 #include "cxxtimer.hpp"
 
+#include <algorithm>
+
 namespace llmq
 {
 
 CSigSharesManager* quorumSigSharesManager = nullptr;
+
+namespace
+{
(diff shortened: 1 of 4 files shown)


reubenyap:
@codeant-ai review

psolstice (dismissed):

PR #1919, opened 2026-08-26
Logging: avoid empty rate-limit marker lines

navidR:
## PR intention
This change prevents repeated bare [*] lines while log rate limiting is active. UpdateTip() now writes the block update, optional warning, and newline as one atomic log record. Logging behavior is otherwise unchanged.


reubenyap:
@codeant-ai review

PR #1925, opened 2026-08-29
Net: Interrupt orphan transaction processing

navidR:
Process at most one expensive orphan transaction attempt per scheduling cycle. This prevents a newly accepted parent from causing a long uninterrupted validation stall while preserving orphan processing across later cycles.

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -879,6 +879,8 @@ public:
     CAmount lastSentFeeFilter;
     int64_t nextSendTimeFeeFilter;
 
+    std::set<uint256> orphan_work_set;
+
     // Challenge sent in VERSION to be answered with MNAUTH (only happens between MNs)
     mutable CCriticalSection cs_mnauth;
     uint256 sentMNAuthChallenge;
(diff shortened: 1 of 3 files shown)


reubenyap:
@codeant-ai review

reubenyap (commented):
I found two actionable lifecycle regressions in the deferred work:

- `src/net_processing.cpp:2293`: the remaining orphan IDs are stored on the peer that delivered the parent. If peer A supplied two children missing parent P and peer B supplies P, one child is attempted immediately and the other exists only in B’s `orphan_work_set`. If B disconnects before its next message-handler pass, `CNode` destruction drops that set; A’s orphan remains in the global orphan map, but P is already accepted and no path schedules it again. Please keep deferred orphan work in net-processing state independent of the parent peer’s lifetime (or explicitly transfer it during teardown).
- `src/net_processing.cpp:1343`: `setMisbehaving` is recreated on every helper call, while line 1367 deliberately stops after one live orphan. It therefore never suppresses a second penalty, unlike the old single-batch loop: successive deferred invalid orphans from one source can each add DoS score across passes. Please persist the punished-source state across the sliced batch, or otherwise retain the previous one-penalty-per-source behavior.

The existing test exercises two passes on a live node, so neither disconnect nor repeated-source case is covered.

PR #1926, opened 2026-08-29
bls: Guard null contribution verification vector

navidR:
Reject DKG contribution-share verification when its verification vector is missing. This prevents a null dereference and deterministic node crash during DKG processing.


reubenyap:
@codeant-ai review

reubenyap (commented):
I traced the null-vector case from DKG justification handling through the asynchronous verifier. Short-circuit evaluation now rejects the null pointer before dereference and returns the same resolved `false` future used for other invalid inputs. The full vector validation flagged elsewhere already existed before this PR, so it is not a regression here. The focused test and full CI matrix are green; I found no actionable defect.

PR #1929, opened 2026-08-30
LLMQ: Handle empty contribution batches

navidR:
Handle contribution batches that become empty after filtering, and complete empty BLS verification requests immediately. This prevents DKG phase processing from waiting indefinitely and adds focused regression coverage.

Upstream: [Dash PR #7486](https://github.com/dashpay/dash/pull/7486).


reubenyap (commented):
I reviewed both synchronous and asynchronous empty-input behavior and traced the DKG path where every queued contributor becomes invalid before verification. The member-index, verification-vector, and secret-share collections are built in lockstep; the new guard safely drains the moved queue and avoids constructing the verifier with zero inputs. The focused tests and full CI matrix are green; I found no actionable defect.

levonpetrosyan93 (approved):

PR #1931, opened 2026-08-30
Evo: handle validation failures gracefully

navidR:
Replace externally reachable Evo assertions with explicit validation failures. Add regression coverage for malformed coinbase payload and deterministic masternode state paths.

Upstream: [Dash PR #3504](https://github.com/dashpay/dash/pull/3504).

### DIFF
--- a/src/evo/cbtx.cpp
+++ b/src/evo/cbtx.cpp
@@ -227,7 +227,7 @@ bool CalcCbTxMerkleRootQuorums(const CBlock& block, const CBlockIndex* pindexPre
                 llmq::CFinalCommitment qc;
                 uint256 minedBlockHash;
                 bool found = llmq::quorumBlockProcessor->GetMinedCommitment(p.first, p2->GetBlockHash(), qc, minedBlockHash);
-                assert(found);
+                if (!found) return state.DoS(100, false, REJECT_INVALID, "commitment-not-found");
                 v.emplace_back(::SerializeHash(qc));
                 hashCount++;
             }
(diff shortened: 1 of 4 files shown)


navidR:
This not from upstream, but something we should address:

```
  diff --git a/src/evo/cbtx.cpp b/src/evo/cbtx.cpp
  @@ -227,7 +227,7 @@ bool CalcCbTxMerkleRootQuorums(
                   uint256 minedBlockHash;
                   bool found = llmq::quorumBlockProcessor->GetMinedCommitment(
                       p.first, p2->GetBlockHash(), qc, minedBlockHash);
  -                if (!found) return state.DoS(100, false, REJECT_INVALID, "commitment-not-found");
  +                if (!found) return state.Error("commitment-not-found");
                   v.emplace_back(::SerializeHash(qc));
                   hashCount++;
               }

  diff --git a/src/test/evo_validation_tests.cpp b/src/test/evo_validation_tests.cpp
  @@ -133,7 +133,7 @@ BOOST_FIXTURE_TEST_CASE(
       uint256 merkleRoot;
       CBlock block;
       BOOST_CHECK(!CalcCbTxMerkleRootQuorums(block, pindexPrev, merkleRoot, state));
  -    BOOST_CHECK(state.IsInvalid());
  +    BOOST_CHECK(state.IsError());
       BOOST_CHECK_EQUAL(state.GetRejectReason(), "commitment-not-found");
       dbTransaction->Rollback();
   }
```

reubenyap (commented):
Reviewed the current head and traced the new failure paths through special-transaction validation, block processing, CValidationState, and peer scoring. The malformed-payload branches now fail cleanly, and the full CI matrix is green. The one remaining issue is already accurately covered by the existing review thread: a missing stored commitment is a local database error and should use state.Error rather than classify the peer block as invalid. I found no additional actionable issues.

PR #1932, opened 2026-08-30
Evo: Preserve DMN history across reorgs

navidR:
Preserve persisted deterministic-masternode diffs and snapshots when disconnecting blocks, while evicting only the in-memory cache entry. Add a regression test proving orphan history remains reconstructible across reorgs.

Upstream: [Dash PR #3435](https://github.com/dashpay/dash/pull/3435).

### DIFF
--- a/src/evo/deterministicmns.cpp
+++ b/src/evo/deterministicmns.cpp
@@ -606,9 +606,6 @@ bool CDeterministicMNManager::UndoBlock(
             prevList = GetListForBlock(pindex->pprev);
         }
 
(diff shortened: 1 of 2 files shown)


reubenyap (commented):
Reviewed the current head and traced disconnects through the EvoDB transaction, cache eviction, list reconstruction, reconnect, and VerifyDB rollback paths. Keeping the persisted diff/snapshot while evicting the disconnected block from memory preserves orphan history without letting stale cached state drive the active tip. The regression and full CI matrix are green. I found no actionable issues.

levonpetrosyan93 (approved):

PR #1934, opened 2026-08-30
LLMQ: Lock InstantSend input request IDs

navidR:
Protect InstantSend input request-ID insertion with the manager mutex already used by readers and cleanup paths. This removes a cross-thread data race without changing protocol behavior.

Upstream: [Dash PR #3829](https://github.com/dashpay/dash/pull/3829).


reubenyap (commented):
Reviewed the full change and the synchronization around every `inputRequestIds` access. The insertion now uses the same manager mutex as reads and cleanup, including re-entrant call paths, without changing protocol behavior. CI is green, and I found no new actionable defects.

PR #1935, opened 2026-08-30
LLMQ: bound pending recovered signatures

navidR:
Bound pending recovered signatures per peer and globally, with consistent cleanup during draining, banning, and disconnects. Add boundary and concurrent-accounting regression coverage.

Upstream: [Dash PR #7402](https://github.com/dashpay/dash/pull/7402).


reubenyap (commented):
I traced every mutation of the pending recovered-signature queue and its running count, including cap admission, randomized draining, empty-node pruning, banning, disconnect cleanup, and lock ordering. The count stays under `cs`, and `RemoveNodesIf` erases each matching peer's complete list, so the existing “only one batch is removed” warning is not borne out by the code. The focused tests and full CI matrix are green; I found no actionable defect.

PR #1936, opened 2026-08-30
LLMQ: Bound signature share verification batches

navidR:
Limit verification batches by the actual number of signature shares rather than the number of sessions. Add coverage for bounded, complete draining of a large session.

Upstream: [Dash PR #7415](https://github.com/dashpay/dash/pull/7415).


reubenyap (commented):
I traced the new share-count bound through collection and verification, including already-known shares, banned peers, randomized peer iteration, quorum lookup, and subsequent queue draining. Only admitted shares increment the counter, and the continuation condition prevents a verification batch from exceeding 32 actual shares. The focused regression coverage and full CI matrix are green; I found no actionable defect.

PR #1937, opened 2026-08-30
Net: bound automatic peer discouragement

navidR:
Replace unbounded persisted automatic bans with bounded rolling discouragement while preserving manual bans. Add coverage for expiry, disconnect behavior, and inbound eviction preference.

Upstream: [Bitcoin Core PR #19219](https://github.com/bitcoin/bitcoin/pull/19219).


reubenyap (commented):
Reviewed the discouragement lifecycle end to end: automatic scoring, address-wide disconnects (including onion addresses), inbound/outbound admission, eviction preference, manual-ban persistence, and the new tests. I also checked the existing concerns against real connection paths and per-test fixture isolation; CI is green, and I found no additional actionable defects.

PR #1938, opened 2026-08-30
LLMQ: bound locked transaction requests

navidR:
Limit locked-transaction requests to four suitable peers and use the correct inventory mutex. Add focused coverage for peer selection, masternode preference, and cleanup.

Upstream: [Dash PR #4680](https://github.com/dashpay/dash/pull/4680) and [Dash PR #4203](https://github.com/dashpay/dash/pull/4203).


reubenyap (commented):
Reviewed peer selection end to end, including connection lifetime/ref-counting, the inventory lock, masternode-first ordering, request cleanup, and the focused boundary test. The four-peer cap is enforced without dropping the preferred eligible peers; CI is green, and I found no new actionable defects.

PR #1939, opened 2026-08-30
LLMQ: Bound InstantSend lock intake

navidR:
Bound InstantSend input vectors and the pending-lock queue, and process at most 32 locks per pass. Add boundary coverage for each admission and processing limit.

Upstream: [Dash PR #7397](https://github.com/dashpay/dash/pull/7397) and [Dash PR #3399](https://github.com/dashpay/dash/pull/3399).


reubenyap (commented):
Reviewed the intake limit, queue boundary and draining behavior, recovered-signature shortcut, callers, and boundary coverage. The 32-item passes continue draining safely and the exact-signature check preserves verification semantics. CI is green; I found no additional actionable defects beyond the logging-format mismatch already raised in the discussion.

PR #1940, opened 2026-08-30
QT freeze during batch verification fixed

levonpetrosyan93:


src/validation.cpp
```diff
@@ -3988,12 +3988,13 @@ bool ActivateBestChain(CValidationState &state, const CChainParams& chainparams,
...
             }
-            BatchProofContainer* batchProofContainer = BatchProofContainer::get_instance();
-            batchProofContainer->fCollectProofs = ShouldBatchSparkProofs(pindexNewTip);
-            if (!VerifyPendingSparkBatch(state, "connecting new tip"))
-                return false;
         }
 
+        BatchProofContainer* batchProofContainer = BatchProofContainer::get_instance();
+        batchProofContainer->fCollectProofs = ShouldBatchSparkProofs(pindexNewTip);
+        if (!VerifyPendingSparkBatch(state, "connecting new tip"))
```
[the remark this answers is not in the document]
psolstice: valid point

psolstice (changes requested):

reubenyap (changes requested):
Requesting changes on `8deca81`.

- `fCollectProofs` remains a public, unsynchronized state flag. At the old-to-recent batching boundary, concurrent `ActivateBestChain()` calls can make one thread skip a Spark proof, another clear collection mode, and `finalize()` discard that proof. The affected block can then be accepted without the proof being checked inline or retained for batch verification.
- `verify_pending()` treats equal vector sizes as batch identity. A disconnect followed by a same-size replacement can apply a stale verdict and clear an unverified replacement batch.
- Cover sets are loaded outside `cs_main`, but `GetCoinSet()` subsequently reads mutable `coinGroups`, chain, and block-index data unlocked. Concurrent connect/disconnect can produce a data race or a mixed-tip cover-set snapshot.
- Releasing `cs_main` after publishing chain, wallet, index, and Spark state lets other threads consume or extend state before the deferred proof verdict succeeds.
- The empty-success path returns without removing `sparkbatchfailed`, so a clean empty batch can force an unnecessary reindex on restart.

Please make collection, finalization, and verification one generation-checked state transition, take all cover sets from one chain-consistent snapshot, and prevent unverified state from becoming externally usable. An atomic flag alone is insufficient. Add barrier-based regressions for the old-to-recent race, same-size replacement, concurrent cover-set mutation, and empty-marker cleanup.

Current exact-head Linux Release and Debug built and passed all 94 CTests, but both failed extended RPC tests, including the same `spark_mintspend.py` assertion.

src/validation.cpp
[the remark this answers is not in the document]
reubenyap: Fixed in df973799f, 8e9150e38, and f8d59594b. Batch collection and mode transitions are now guarded by cs_batch; ConnectBlock snapshots and verifies each block-local batch while still holding cs_main; and GetCoinSet holds cs_main for its complete traversal. verify_pending releases cs_batch before invoking the cover-set callback, avoiding cs_batch-to-cs_main lock overlap. The concurrency regression passes in spark_batch_tests; all 37 spark_tests and spark_batching.py also pass.

src/validation.cpp
```diff
@@ -2307,10 +2307,17 @@ bool AbortNode(CValidationState& state, const std::string& strMessage, const std
...
 
-static bool ShouldBatchSparkProofs(const CBlockIndex* pindex)
+static bool ShouldDeferSparkBatchVerification(const CBlockIndex* pindex)
 {
-    // Defer Spark proof verification for blocks older than a day, which means we are syncing or reindexing
-    return ((GetSystemTimeInSeconds() - pindex->GetBlockTime()) > 86400) && GetBoolArg("-batching", true);
+    // Defer Spark proof verification for blocks older than a day (IBD/reindex).
+    // GetTime() is mockable so -mocktime RPC tests use the same recent vs deferred
+    // split as a live node; wall-clock time would treat 2014 mocktime chains as IBD.
+    return ((GetTime() - pindex->GetBlockTime()) > 86400) && GetBoolArg("-batching", true);
```
reubenyap, line 2315: **[P1] Drain deferred proofs even when the tip stays old**

Tip age is not proof that syncing will continue. A network/bootstrap sync or higher-work reorg can stop at a tip older than 24 hours; the enrolled Spark proofs have skipped direct cryptographic verification, yet `ConnectTip` commits and activates them, and this predicate prevents the normal pending drain from running. Clean shutdown and full-reindex completion are the only other drains. Add an explicit sync/activation completion barrier, or bound and verify before publishing, so a stale tip cannot remain active with unchecked proofs.

src/batchproof_container.cpp
```diff
@@ -72,34 +184,135 @@ void BatchProofContainer::finalize() {
...
+                return true;
+            if (fBatchFailed)
+                return false;
+            if (sparkTransactions.empty() && historicalSparkTransactions.empty()) {
+                if (!fReindex)
+                    RemoveRecoveryMarker();
+                return true;
+            }
+
+            snapshotTransactions.swap(sparkTransactions);
```
reubenyap, line 246: **[P1] Keep an explicit in-flight batch and restore it on exceptions**

Swapping every member vector into locals makes the container look empty while verification runs. A second `ActivateBestChain` verifier can enter after `cs_main` is released, take the empty-success path, and delete `sparkbatchfailed`; a crash then leaves committed unchecked state with no recovery marker. `VerifySparkBatch` also rethrows `std::bad_alloc`, so unwinding destroys the only proof copy while P2P/RPC catch the exception and continue. Serialize pending verifiers or track an in-flight generation, and restore the snapshot with a scope guard unless its verdict is committed.

src/batchproof_container.cpp
```diff
@@ -72,34 +184,135 @@ void BatchProofContainer::finalize() {
...
+    LOCK(cs_batch);
+    if (!fCollectProofs)
+        return false;
     tempHistoricalSparkTransactions.push_back(tx);
     tempHistoricalSparkTxIds.push_back(txHash);
+    return true;
 }
 
 void BatchProofContainer::remove(const spark::SpendTransaction& tx) {
+    LOCK(cs_batch);
```
reubenyap, line 315: **[P2] Let disconnect cancel proofs already being verified**

`remove()` searches only the member vectors, but `verify_pending()` has moved those vectors into locals. A concurrent reorg therefore cannot remove a disconnected spend. If that snapshot then fails against post-reorg state, it is restored, `fBatchFailed` is latched, and the node aborts/reindexes for a block that is no longer active. Record removals against the in-flight generation and retry against the current active batch.

src/validation.cpp
```diff
@@ -3023,6 +3042,27 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         }
     }
 
+    // After InstantSend filtering so a conflicting block still returns
+    // conflict-tx-lock instead of a Spark-batch reject, and collection stays
+    // open through ProcessSpecialTxsInBlock as it did on master.
+    if (fCollectSparkProofs && !fDeferBatchVerify) {
```
reubenyap, line 3048: **[P2] Verify Spark before side-effectful special processing**

Moving this check after `ProcessSpecialTxsInBlock` preserves `conflict-tx-lock` precedence, but normal special processing runs with notifications enabled. If this later rejects `bad-spark-batch-proof`, EvoDB rolls back while deterministic-MN/UI notifications and quorum cache changes do not. Run the non-mutating InstantSend conflict check first, then verify the Spark batch, then apply and notify special-transaction state.

src/batchproof_container.cpp
```diff
@@ -72,34 +184,135 @@ void BatchProofContainer::finalize() {
     tempHistoricalSparkTransactions.clear();
     tempHistoricalSparkTxIds.clear();
     fCollectProofs = false;
+    if (!sparkTransactions.empty() || !historicalSparkTransactions.empty())
```
reubenyap, line 187: **[P2] Fail the block when the recovery marker cannot be created**

`finalize()` now relies on `WriteRecoveryMarker()` before deferred proofs have a verdict, but that helper returns `void`, only logs `fopen` failure, and does not durably sync the file. `ConnectTip` still commits the unchecked state, so a crash can restart without forcing reindex. Return and check marker-creation failure, and make the marker durable before allowing the block commit.

reubenyap (changes requested):
Requesting changes on the current head. The restored cross-block Spark batch lifecycle still has fail-open and reorg races. Please keep deferred proof state represented until one verifier commits a verdict, drain it at a deterministic sync boundary rather than using age alone, and keep side-effectful block processing behind successful proof validation. Please add focused regressions for the interleavings described inline.

reubenyap:
Rechecked **`f491e6c` against `4f0c771`**. **Recommendation: request changes.** All five open findings have a valid underlying concern, but two already exist on the base branch.

| Finding | Verdict and required change |
| --- | --- |
| **P1 — [In-flight batch ownership](https://github.com/firoorg/firo/pull/1940#discussion_r3925009697)** | **Valid regression.** Detached proofs make the container appear empty before verification finishes; exceptions can also discard the only proof copy. Keep in-flight work represented until its verdict is committed, and restore it on exceptional exits. |
| **P2 — [Disconnect cancellation](https://github.com/firoorg/firo/pull/1940#discussion_r3925009704)** | **Valid regression.** `remove()` cannot reach the detached snapshot, so disconnected proofs can cause a stale failure and unnecessary recovery. Reconcile cancellations and chain changes before committing the verdict. |
| **P2 — [Special-processing order](https://github.com/firoorg/firo/pull/1940#discussion_r3925009715)** | **Valid regression.** Masternode/UI notifications and peer-disconnection flags can be applied before Spark rejects the block; database rollback does not undo them. Verify before applying those effects. The notification issue is confirmed; cache invalidation alone does not establish cache corruption. |
| **P1 — [Old-tip drain](https://github.com/firoorg/firo/pull/1940#discussion_r3925009691)** | **Valid, pre-existing.** An old tip can remain active with pending proofs indefinitely. Add a completion/drain boundary independent of tip age. |
| **P2 — [Recovery-marker persistence](https://github.com/firoorg/firo/pull/1940#discussion_r3925009722)** | **Valid, pre-existing helper.** Marker creation errors are not propagated and persistence is not made durable. Require successful durable recovery-state persistence before committing deferred state. |

The earlier raw vector/collection-flag races, lock inversion, equal-size replacement-clearing bug, and ordinary empty-marker cleanup are **fixed as originally reported**. Complete `GetCoinSet()` traversal is now locked, but loading several groups still does not provide one common chain snapshot. The `.find()` nitpick is obsolete, boolean naming is fixed, and remaining brace-style comments are cosmetic. CodeRabbit’s “minimal risk” summary refers to the older `59019` revision.

Validation is green, but coverage needs correction:

- [Current CI](https://github.com/firoorg/firo/actions/runs/33753404015): Linux RelWithDebInfo and Debug each passed **94 CTests and the extended RPC suite**. A fresh local `spark_mintspend.py` also passed using the `72a909d91` binary, whose production source matches this head.
- The [cleanup test](https://github.com/firoorg/firo/blob/f491e6cd7bec1e4d60d4b69962e36aa2cb4219c0/src/test/spark_tests.cpp#L3263) passes `fJustCheck=true`, which disables collection; it does not test cleanup of enrolled proofs.
- The [replacement test](https://github.com/firoorg/firo/blob/f491e6cd7bec1e4d60d4b69962e36aa2cb4219c0/src/test/spark_batch_test.cpp#L75) adds another batch synchronously. It does not cover a second verifier, cancellation, exception recovery, or failed marker persistence.

These findings are based on source/control-flow review; concurrency fault injection and GUI responsiveness benchmarks were not run.

levonpetrosyan93:
Closing this and keep more cleaner alternative #1951

PR #1941, opened 2026-08-30
Evo: ignore stale block tip notifications

navidR:
Ignore delayed deterministic-masternode tip notifications when their block is no longer the active chain tip. Add regressions covering stale callbacks and valid pure-disconnect updates after repeated invalidations.


reubenyap (commented):
Reviewed the current draft head. I traced queued tip notifications through the deterministic-masternode manager, checked every tipIndex reader, and verified the cs_main-to-manager lock order against block processing. The stale callback is rejected while current and null-tip updates still work, and the full CI matrix is green. I found no actionable issues.

PR #1943, opened 2026-08-31
Spark: clear all state on reset

navidR:
CSparkState::Reset() currently leaves the linking-tag transaction index and extended mint metadata populated. Clear both containers so reset restores constructor-equivalent empty state and cannot expose stale entries after a rebuild or between tests.

### DIFF
--- a/src/spark/state.cpp
+++ b/src/spark/state.cpp
@@ -1587,6 +1587,8 @@ void CSparkState::Reset() {
     }
     usedLTags.clear();
     mobileUsedLTags.clear();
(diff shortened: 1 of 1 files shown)


levonpetrosyan93 (approved):

reubenyap (commented):
Reviewed the current head and traced CSparkState::Reset against the state members, rebuild path, and mobile spend RPC. Clearing the linking-tag index and extended mint metadata completes the reset without changing the live update behavior. The full CI matrix is also green. I found no actionable issues.

PR #1946, opened 2026-09-04
Qt: avoid blocking on Spark activation height

navidR:
Use ClientModel's nonblocking cached-height accessor for Spark send controls, leaving transaction creation's locked chain-height checks unchanged. Initialize the cache and client-model wiring before use, exclude header-only heights, and add Qt regression coverage. This addresses the height-query stall observed during Debug replay of 1,024 Spark spends; separate startup and replay stalls remain.

### DIFF
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -38,7 +38,7 @@ ClientModel::ClientModel(OptionsModel *_optionsModel, QObject *parent) :
 {
     cachedBestHeaderHeight = -1;
     cachedBestHeaderTime = -1;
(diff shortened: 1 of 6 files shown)


reubenyap (commented):
Reviewed head `ce28c3fe98f951d0b8f0a1c80e80aef4f37c6736`. No additional actionable issues were found. I traced confirmed-height initialization and block/header notifications through `versionedSparkSpendsAllowed()`, verified wallet/client wiring precedes model use, and checked the focused Qt regression and build integration.

psolstice (approved):

reubenyap (approved):

PR #1947, opened 2026-09-04
Wallet: prefilter masternode collateral scans

navidR:
## PR intention

Check collateral membership before running Spark ownership and spent-state checks during the startup wallet scan, preserving the existing locks and collateral-locking rules.

## Code changes brief

Add coverage for internal and external collateral, watch-only and spent outputs, manual locks, and repeated scans.
Debug wallet and deterministic-masternode tests pass, and startup was verified against the existing 1,024-spend regtest fixture.

### DIFF
--- a/src/wallet/test/wallet_tests.cpp
+++ b/src/wallet/test/wallet_tests.cpp
@@ -4,6 +4,12 @@
 
 #include "wallet/wallet.h"
 
+#include "evo/deterministicmns.h"
+#include "evo/evodb.h"
+#include "evo/providertx.h"
+#include "evo/specialtx.h"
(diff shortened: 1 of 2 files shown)


reubenyap (commented):
Reviewed head `5ccb44acbae38bf9dc99024666c81a685fdbf7e4`. No additional actionable issues were found. I traced `AutoLockMasternodeCollaterals()` through its sole startup caller and the ownership, spent-state, and deterministic-MN helpers; the collateral-first short-circuit preserves lock semantics, and the regression covers internal/external, watch-only, foreign, spent, manual-lock, and repeat-scan cases.

PR #1952, opened 2026-09-08
Validation: fix in-memory coin-cache allocation

reubenyap:
## PR intention

Correct the in-memory coin-cache byte budget and include EvoDB within `-dbcache`. Dividing the remaining bytes by 300 left only about 1.5 MiB for UTXOs at the default setting. EvoDB also received an extra 16 MiB outside the configured budget.

## Code changes brief

Assign the remaining bytes directly to `nCoinCacheUsage`. Reserve one eighth of the budget remaining after the block-index allocation for EvoDB, capped at 16 MiB, before allocating the coin caches. Small settings therefore retain space for each cache.

At the default 450 MiB with `txindex=0`, the allocation is 2 MiB block index + 16 MiB EvoDB + 8 MiB chainstate + 424 MiB UTXO. The existing unused-mempool allowance and flush safety factors remain unchanged. This is a cache target, not a hard process-memory limit; small EvoDB caches may increase database I/O.

Register `dbcache.py` in the regular RPC runner and check all four cache allocations for default, 2048 MiB, minimum 4 MiB, and txindex-enabled configurations.

Validation at `a3f78d354`:

- `cmake --build build --target firod firo-cli test_firo -j 8`: passed, native Windows/MinGW Debug.
- `python qa/rpc-tests/dbcache.py --tmpdir=build/dbcache-evodb-fixed --portseed=955 --nocleanup`: all four configurations passed. The updated test fails against the preceding PR binary `17f654270`, which still assigns 440 MiB to UTXOs before the extra EvoDB allocation.
- `build/bin/test_firo.exe --run_test=coins_tests,dbwrapper_tests --catch_system_error=no --log_level=test_suite -- DEBUG_LOG_OUT`: all 14 cases passed.
- `git diff --check` passed; merge-tree integration with master `f1cba134a` was clean.

Local configuration required a temporary missing-test-helper workaround and Berkeley DB include path; neither is committed. Fresh CI is pending. Full sync benchmarks, peak-memory/flush-latency measurements, the full test suite, and Linux/macOS builds were not run locally.

PR #1953, opened 2026-09-10
Wallet: identify block mints outside cs_main

reubenyap:
## PR intention

Reduce lock contention during Spark block-wallet mint scanning. The worker currently holds `cs_main` and `cs_spark_wallet` while identifying every Spark output, including outputs belonging to other wallets.

## Code changes brief

Split mint identification from recording. The existing single worker identifies owned outputs without either lock, then takes both locks, checks that the block is still active, and reads current height/spent state before updating the wallet. The view keys are fixed for the wallet's lifetime.

When rediscovering a mint, check both regular and stem pools for its pending spend. Read the spent flag and transaction hash together under the selected pool's lock.

Regressions cover owned/foreign outputs, persisted metadata, queued disconnect/reconnect, and pending spends in each pool. The stem-only regression reproduced unused mint records and a missing spend entry before the fix. Complements #1952, already in the base.

Validation passed on native Windows/MinGW Debug:

- `cmake --build build --target test_firo firod firo-cli -j 8`
- `build/bin/test_firo.exe --run_test=spark_wallet_tests,spark_tests,spark_state_tests,spark_mintspend,coins_tests,dbwrapper_tests --catch_system_error=no --log_level=test_suite -- DEBUG_LOG_OUT`: 72 cases passed.
- `python qa/rpc-tests/spark_mint.py --tmpdir=build/rpc-stem-fix-mint --portseed=1281 --nocleanup`
- `python qa/rpc-tests/spark_mintspend.py --tmpdir=build/rpc-stem-fix-mintspend --cachedir=build/rpc-cache-fixed --portseed=1282 --nocleanup`
- `git diff --check`

Local build/test compatibility workarounds (missing CMake helper files, BDB include path, Python `asyncore`) are excluded from the commits. Startup/sync speedup and peak memory have not been measured. GUI and Linux/macOS validation were not run locally.

### DIFF
--- a/src/spark/sparkwallet.cpp
+++ b/src/spark/sparkwallet.cpp
@@ -859,49 +859,51 @@ CAmount CSparkWallet::getMySpendAmount(const std::vector<GroupElement>& lTags) c
     return result;
 }
 
+CSparkWallet::IdentifiedMint CSparkWallet::IdentifyMint(spark::Coin coin, const uint256& txHash) const {
+    // These keys are initialized before the wallet worker starts and never change.
+    const auto identified = coin.identify(viewKey);
+    const auto recovered = coin.recover(fullViewKey, identified);
+    IdentifiedMint mint{};
+    mint.meta.txid = txHash;
+    mint.meta.i = identified.i;
+    mint.meta.d = identified.d;
+    mint.meta.v = identified.v;
+    mint.meta.k = identified.k;
+    mint.meta.memo = identified.memo;
+    mint.meta.serial_context = coin.serial_context;
+    mint.meta.type = coin.type;
+    mint.meta.coin = std::move(coin);
+    mint.lTag = recovered.T;
+    return mint;
+}
+
+void CSparkWallet::RecordMint(IdentifiedMint mint, CWalletDB& walletdb) {
+    auto& mintMeta = mint.meta;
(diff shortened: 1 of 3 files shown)


reubenyap:
@codeant-ai review

reubenyap (commented):
Found one actionable issue:

- **P2 – Synchronize post-block wallet reads with mint recording.** [Head b04e5d5, line 932](https://github.com/firoorg/firo/blob/b04e5d5a5a0db7ea4d6c5b9e5cdaaab3a9cfe629/src/spark/sparkwallet.cpp#L932) starts the full identification pass before cs_spark_wallet is acquired at line 948. A spend or balance caller can therefore take the wallet lock after ConnectTip but before recording, still see the just-mined mints with nHeight == -1, and reject a valid spend as InsufficientFunds. Both Linux Debug and RelWithDebInfo runs reproduce this in the unchanged spark_mintspend/spark_limit_test on the first spend immediately after GenerateBlock; the exact base commit passes both Linux jobs, and GetAvailableSparkCoins filters every nHeight < 1 mint. Please make post-block spend/balance snapshots wait for the pending record phase without moving the cryptographic work back under cs_main, and retain an immediate-post-block spend regression.

I also traced the reorg membership check, key lifetime, pool locking and pending-spend metadata, and all production callers; no other actionable issues remained.

psolstice (approved):

PR #1954, opened 2026-09-10
LLMQ: Reject unmined InstantSend parents

reubenyap:
## PR intention
Prevent InstantSend processing from treating an unmined stem-pool parent as a mined transaction.

`GetTransaction()` searches both transaction pools and can return `true` for a stem-only transaction without setting `hashBlock`. `CheckCanLock()` then passed the null hash to `mapBlockIndex.at()`, which throws `std::out_of_range` from the InstantSend worker path.

Require a non-null block hash before consulting `mapBlockIndex`. An unlocked, unmined parent remains ineligible for InstantSend, matching the existing policy for an unlocked parent in the main mempool.

## Code changes brief

- Reject a parent lookup that does not identify a containing block.
- Clarify the diagnostic as a failure to find a mined parent.
- Add a focused unit test with the parent present only in the stem pool.

### Validation

- Built the Windows `test_firo` target successfully.
- Confirmed the regression test fails on the vulnerable condition with `std::out_of_range: Unable to find key in unordered_map`.
- Confirmed `test_firo --run_test=quorums_instantsend_tests --catch_system_error=no --log_level=test_suite -- DEBUG_LOG_OUT` passes with the fix.
- `git diff --check` passes.

The supplied multi-node functional PoC was not rerun locally.

### Related work

This overlaps #1939 in `quorums_instantsend.h`, `src/test/CMakeLists.txt`, and `quorums_instantsend_tests.cpp` because both changes add InstantSend unit coverage. The production behavior changes are distinct. Whichever PR lands second will need a small test-file reconciliation.

### DIFF
--- a/src/llmq/quorums_instantsend.cpp
+++ b/src/llmq/quorums_instantsend.cpp
@@ -540,9 +540,9 @@ bool CInstantSendManager::CheckCanLock(const COutPoint& outpoint, bool printDebu
     CTransactionRef tx;
     uint256 hashBlock;
     // this relies on enabled txindex and won't work if we ever try to remove the requirement for txindex for masternodes
-    if (!GetTransaction(outpoint.hash, tx, params, hashBlock, false)) {
+    if (!GetTransaction(outpoint.hash, tx, params, hashBlock, false) || hashBlock.IsNull()) {
         if (printDebug) {
-            LogPrint("instantsend", "CInstantSendManager::%s -- txid=%s: failed to find parent TX %s\n", __func__,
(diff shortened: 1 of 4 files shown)


reubenyap (commented):
No additional actionable issues found. I traced stem/main-pool lookup precedence and races, txindex and hashBlock semantics, mapBlockIndex use, and every CheckCanLock caller and retry path. The new quorums_instantsend_tests suite passed in both Linux configurations; the lone Linux RelWithDebInfo failure was in an unrelated unchanged Spark test, while the remaining builds passed. The existing inline review already covers the test cleanup concern, and git diff --check is clean.

PR #1956, opened 2026-09-13
Wallet: Require Spark spend provenance for SMint outputs

reubenyap:
## PR intention

Prevent `OP_SPARKSMINT` outputs from being treated as authenticated Spark value when the containing transaction has no validated Spark spend.

Consensus remains unchanged, preserving historical blocks and valid Spark V1/V2 spends. Standard relay rejects unauthenticated SMints, while Spark state and wallet accounting ignore them if they appear in a block.

## Code changes brief

- Require `IsSparkSpend()` provenance before relaying, extracting, locating, owning, or crediting an `OP_SPARKSMINT`.
- Reuse the filtered mint extractor for mempool removal and block-conflict cleanup so addition and removal track the same coins.
- Add regression coverage for standalone and mixed transactions, plus a valid Spark spend control.

Validation:

- `test_firo --run_test=spark_wallet_tests/standalone_smint_has_no_wallet_provenance` passed.
- `test_firo --run_test=spark_wallet_tests` passed all 9 cases.
- `test_firo --run_test=spark_tests/get_outpoint` passed.
- All changed C++ translation units compiled in a clean local Windows build.
- `git diff --check` passed.

Not run: full CI, a live RPC restart/rescan reproduction, or a direct mempool removal/block-conflict cycle test.

Previously persisted metadata from the mixed public-mint plus SMint variant is not migrated on normal startup. `-zapwalletmints` rebuilds Spark mint metadata through reindex. The reported standalone variant did not create Spark mint metadata, and its wallet transaction accounting is corrected by this change.


reubenyap:
@codeant-ai review

PR #1958, opened 2026-09-17
RPC: skip UTXO scans after txindex misses

navidR:
Skip the expensive UTXO fallback after a transaction-index miss to avoid holding the main chain lock for unnecessary database lookups.
Keep existing behavior when txindex is disabled and add regression tests for indexed misses, confirmed transactions, and both transaction pools.

Related follow-up: #1959 contains the broader transaction-lookup and proof hardening, split out to keep this PR focused. This PR retains only the original `7ad2e1241` commit; #1959 depends on this PR landing first.


navidR:
```
  - Missing lookup: 64.63 ms → 0.66 ms.
```

PR #1961, opened 2026-09-20
Enable scanning of BIP44 internal (change) entries

psolstice:
## PR intention
Both `firod` and `firo-qt` previously skipped internal entries as defined in BIP44. Therefore, if a user creates a transaction using alternative client (such as campfire) and then switches to firod/firo-qt using the same mnemonic the change from the transaction won't be seen. This PR fixes that problem.

## Code changes brief
- src/wallet/wallet.h — adds BIP44_INTERNAL_INDEX, the TopUpInternalChain / MarkInternalChainKeyUsed / *InternalChainRescanPending API, and a CKeyID → child index map for the chain.
- src/wallet/wallet.cpp — derives m/44'/<coin>'/0'/1/<n> keys at startup and on unlock, with a -keypool-sized lookahead that extends when a key is seen on chain. Keys stay out of the keypool, so addresses and change still come off the external chain, and they carry the wallet birthday so a rescan isn't cut short. A wallet that already had history sets a persistent internal_chain_rescan flag, which forces one rescan from genesis and is cleared only when that rescan completes (skipped with a warning under -prune).
- src/wallet/rpcdump.cpp — dumpwallet marks these keys internal=1; importwallet parses the marker.
- Tests — new wallet-internalchain.py; wallet-dump.py asserts the internal=1 counts.


src/wallet/wallet.cpp
```diff
@@ -5172,6 +5319,19 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+        if (fPruneMode) {
+            // There is nothing to scan behind the pruned data, and a node that cannot serve the
+            // rescan must not be kept from starting by it.
+            InitWarning(_("Coins received on the bip44 internal chain before now cannot be found on a pruned node. Reindex to look for them."));
+            walletInstance->SetInternalChainRescanPending(false);
+        } else {
+            // ScanForWalletTransactions walks this forward to the wallet birthday, which no key on
+            // the internal chain is older than.
+            LogPrintf("Rescanning for transactions on the bip44 internal chain\n");
+            pindexRescan = chainActive.Genesis();
```
reubenyap, line 5332: [P1] Do not use the local wallet birthday to limit the migration rescan

For an existing wallet restored from a mnemonic after that seed's transactions occurred, `nTimeFirstKey` describes the local key creation time, not the beginning of the seed's history. Setting `pindexRescan` to genesis here does not guarantee historical recovery: `ScanForWalletTransactions` receives `fRecoverMnemonic=false` for this existing wallet and, without explicit `-rescan`, skips blocks older than `nTimeFirstKey - 7200`.

I reproduced this with the PR's CI binary: create a wallet with the same seed after its confirmed internal payments, using `-keypool=0` to model the pre-migration state and mock time one day later, then restart with `-keypool=2`. The log requests the internal-chain rescan, but the balance stays at 0. Restarting the same wallet with explicit `-rescan` finds the historical payments (3 FIRO in this fixture). The genesis pointer alone does not bypass the birthday optimization.

Make the pending internal-chain migration invoke the appropriate recovery scan mode, preserving any explicitly supplied scan date. Add a regression for an existing restored seed whose payments predate the wallet file.

reubenyap (changes requested):
Validated against PR head `8b283a8d893660c8cca7b66b2ecc2445efbf0f5f`. I would request changes: the missing test explains the Linux CI failure, and wallet recovery still misses funds.

I reran isolated regtest checks using the macOS RelWithDebInfo artifact from [CI run 35504351817](https://github.com/firoorg/firo/actions/runs/35504351817). Its embedded commit is `a56849b68ef2`, the CI merge of this head into `6284ec2`; I checked that the merge does not change the recovery code under review.

The following findings are confirmed. Existing discussions are linked to avoid opening duplicate threads.

| Finding | Validation |
| --- | --- |
| [Missing internal-chain test](https://github.com/firoorg/firo/pull/1961#discussion_r4056708716) | The registered `qa/rpc-tests/wallet-internalchain.py` is absent from the commit. Both Linux jobs pass 97/97 unit tests and then raise `FileNotFoundError` at that entry in the RPC runner. |
| [Locked-wallet lookahead](https://github.com/firoorg/firo/pull/1961#discussion_r4056708720) | With `-keypool=2`, confirm 1 FIRO to internal index 1 while the wallet is locked, unlock, then confirm 1 FIRO to index 2. Index 2 still has `ismine=false`; the wallet shows 1 FIRO instead of 2. The failed growth request is not replayed on unlock. |
| Wallet birthday truncates migration recovery — new inline comment | Automatic internal-chain recovery shows 0 FIRO, while restarting with explicit `-rescan` discovers 3 FIRO from the same history. The local wallet creation time is later than the seed's payments. |
| [Historical scan needs another pass after deriving more keys](https://github.com/firoorg/firo/pull/1961#discussion_r4056712509) | With lookahead 2 and confirmed 1-FIRO payments to internal indices 1, 2, 5, 4 in that order, a fresh seed restore shows 3 FIRO. A second rescan shows all 4. Index 5 is derived only after its block has already been scanned. |
| [Non-mnemonic BIP44 migration](https://github.com/firoorg/firo/pull/1961#discussion_r4056708722) | An existing `-usemnemonic=0` HD wallet receives a historical payment to internal index 0 while that key is unknown. After internal keys are derived, `ismine=true` but balance remains 0. Explicit `-rescan` finds the 1 FIRO. |

For the two migration reproductions, I used `-keypool=0` to construct wallets with no internal keys, then restarted with `-keypool=2`. The birthday fixture also uses mock time one day after the payments to model a previously restored wallet. These are controlled migration-state fixtures, not tests run with a pre-PR binary.

Two additional findings remain valid from tracing the source and database calls; I did not inject real disk failures:
- [Rescan-marker persistence](https://github.com/firoorg/firo/pull/1961#discussion_r4056712504): derived keys and the HD counter are written before the caller writes the pending marker, with no transaction spanning them. A crash after the counter write can suppress the migration on restart; `SetInternalChainRescanPending` also ignores database failure returns.
- [Scan completion after a block-read failure](https://github.com/firoorg/firo/pull/1961#pullrequestreview-5260344363): `ReadBlockFromDisk` failure sets `ret=nullptr`, but a later successful block sets it non-null again. The new startup check at `wallet.cpp:5355` then clears the pending marker despite incomplete scanning.

The existing `wallet-dump.py` test passes against the same CI binary. Please include the missing test and cover these recovery cases before merging. Jenkins logs require authentication, so I have not attributed its separate errors to these findings.

PR #1965, opened 2026-09-22
Consensus: reject negative transaction versions

reubenyap:
## PR intention

Prevent negative transaction versions from producing inconsistent version/type interpretation after serialization round trips.

Activation heights intentionally remain unset pending coordinated deployment, including ChainLock quorum readiness.

## Code changes brief

- Add a height-gated consensus rejection for negative transaction versions.
- Preserve existing packed transaction bytes while using defined unsigned packing for transaction hashes and legacy sighashes.
- Add regression coverage for the activation boundary and round-trip behavior.

Validation:
- git diff --check
- Reviewed the complete diff against current origin/master
- Unit tests not run because this host does not have a C++20-capable Firo build toolchain.


reubenyap:
@codeant-ai review

psolstice (dismissed):
Need to set HF block numbers!

